PR #2, opened 2025-05-01
Critical versioning fix. Match Ivy-Framework with Ivy to enable ivy db add to work at all.

OliverMidbrink:
Critical versioning fix. Match Ivy-Framework with Ivy to enable ivy db add to work at all.

Problem:
ivy init creates csproj referencing the assembly version of the local ivy console version (a problem itself that i fixed in other pull request), this causes it to look for the same version of the ivy framework which is not found. The simplest fix would be to sync the ivy versions.


nielsbosma:
@OliverMidbrink I'm working on this - this is not the solution though.

PR #3, opened 2025-05-11
Number input fix

OliverMidbrink:
Before there seemed to be auto rounding of decimals like tax rates and prices to the nearest integer when typing manually. This is changed, so stepping with buttons works for integers, while manual typing results in the decimals being kept.


nielsbosma:
@OliverMidbrink Remove the hardcoded version in the project file. Not supposed to be there.

OliverMidbrink:
Should work now, the hardcoded verison was a left over commit from a previous branch (removed now from this branch).

PR #18, opened 2025-06-05
Finish Supabase Auth Provider

zachwolfe:
See also https://github.com/Ivy-Interactive/Ivy/pull/143, https://github.com/Ivy-Interactive/Ivy/issues/99

### DIFF
--- a/Ivy/Auth/AuthOption.cs
+++ b/Ivy/Auth/AuthOption.cs
@@ -2,8 +2,7 @@
 
 namespace Ivy.Auth;
 
-public record AuthOption(AuthFlow Flow, string? Name = null, string? Id = null, Icons? Icon = null);
-public record WorkOSAuthOption(AuthFlow Flow, string? Name = null, string? Id = null, Icons? Icon = null, string? ConnectionId = null) : AuthOption(Flow, Name, Id, Icon);
+public record AuthOption(AuthFlow Flow, string? Name = null, string? Id = null, Icons? Icon = null, object? Tag = null);
 
 // Comment here for PR auto check purposes - working variant
 public enum AuthFlow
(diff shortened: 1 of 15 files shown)


Ivy/Auth/AuthOption.cs
```diff
@@ -3,6 +3,8 @@
 namespace Ivy.Auth;
 
 public record AuthOption(AuthFlow Flow, string? Name = null, string? Id = null, Icons? Icon = null);
+public record WorkOSAuthOption(AuthFlow Flow, string? Name = null, string? Id = null, Icons? Icon = null, string? ConnectionId = null) : AuthOption(Flow, Name, Id, Icon);
```
nielsbosma: Can we solve this through a new generic solution so the core doesn't need to know about WorkOS. Maby just add ConnectionId to the AuthOption. Or add an even more generic `object Tag`?

Ivy/Auth/AuthToken.cs
```diff
@@ -0,0 +1,3 @@
+namespace Ivy.Auth;
+
+public record AuthToken(string Jwt, string? RefreshToken = null, DateTimeOffset? ExpiresAt = null);
```
nielsbosma, line 3: Why do we need to pass RefreshToken and ExpiresAt to the frontend?

Ivy/Auth/AuthOption.cs
zachwolfe: Sure.

Ivy/Auth/AuthToken.cs
zachwolfe: They need to be stored if we want to keep the user logged in for longer than the access token will allow (e.g., longer than ~1hr)

rorychatt:
@zachwolfe delete branch if it is no longer needed

PR #19, opened 2025-06-11
Matt updated card.md with full usage, variants, styling, and states

GitRectify:


nielsbosma:
Sorry, but there's a bunch of stuff in here that's not even part of the current framework,

PR #26, opened 2025-06-18
feat: bool input view state handling

rorychatt:
Modified the models so the invalid state for bool input component are working in a similar manner to existing components

### DIFF
--- a/Ivy.Samples/Apps/Widgets/Inputs/BoolInputApp.cs
+++ b/Ivy.Samples/Apps/Widgets/Inputs/BoolInputApp.cs
@@ -24,42 +24,42 @@ public class BoolInputApp : SampleBase
                | falseState.ToBoolInput().Label("Label").Description("Description")
                | trueState.ToBoolInput().Label("Label").Description("Description").Disabled()
                | trueState.ToBoolInput().Label("Label").Description("Description").Invalid("Invalid")
-               | nullState.ToBoolInput().Label("Label").Description("Description")
+               | new Box("Not Implemented")
                
                | null!
                | trueState.ToBoolInput().Label("Label")
                | falseState.ToBoolInput().Label("Label")
                | trueState.ToBoolInput().Label("Label").Disabled()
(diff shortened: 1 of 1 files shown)


rorychatt:
image

frontend/src/components/ui/toggle.tsx
```diff
@@ -30,13 +30,19 @@ const Toggle = React.forwardRef<
...
->(({ className, variant, size, ...props }, ref) => (
-  <TogglePrimitive.Root
-    ref={ref}
-    className={cn(toggleVariants({ variant, size, className }))}
-    {...props}
-  />
-))
+>(({ className, variant, size, ...props }, ref) => {
+  let toggleClass = toggleVariants({ variant, size, className });
+  if (className?.includes('bg-red-50')) {
```
rorychatt, line 35: A discussion to have is that maybe there should be a theme support, then the color would be defined as "bg-alert" of some sort, then it would be easier to set up priority in colors.

As if we attach the red it should always be prioritized over green (alert vs primary)

IDK at this moment if we have any themes / idea to support that. For now, this solution works, but maybe we need a more generic one with themes

Ivy.Samples/Apps/Widgets/Inputs/BoolInputApp.cs
```diff
@@ -23,42 +23,42 @@ public class BoolInputApp : SampleBase
                | trueState.ToBoolInput().Label("Label").Description("Description")
                | falseState.ToBoolInput().Label("Label").Description("Description")
                | trueState.ToBoolInput().Label("Label").Description("Description").Disabled()
+               | trueState.ToBoolInput().Label("Label").Description("Description").Invalid("Invalid")
                | new Box("Not Implemented")
-               | nullState.ToBoolInput().Label("Label").Description("Description")
```
rorychatt, line 28: removed this as pretty sure this was not actually implemented

frontend/src/components/ui/toggle.tsx
nielsbosma: We leave theme support for the future. But yes.

PR #27, opened 2025-06-19
feat: add support for arm64

rorychatt:
Add support to build Ivy.Docs and Ivy.Samples for arm64 runtimes


rorychatt:
ref 
https://learn.microsoft.com/en-us/dotnet/core/rid-catalog



I haven't actually tested if the `ivy docs` command works now, as I am not sure if I would first need to deploy some sort of a test package and then install it.

PR #28, opened 2025-06-20
Feat/boolinput nullable state

rorychatt:
Fixed failing test on master (please review [this commit](https://github.com/Ivy-Interactive/Ivy-Framework/pull/28/commits/e5b9cf1756abab8a81cd51c71c62650ffef42ecd))
Implemented support for nullable bool states, nullable numeric states etc


Ivy/Core/Utils.cs
```diff
@@ -11,16 +11,14 @@ public static int StableHash(string text)
     {
         unchecked
         {
-            int hash = 23;
```
nielsbosma: Revert to previous code.

nielsbosma (changes requested):

Ivy/Core/Utils.cs
rorychatt: image

PR #34, opened 2025-06-28
Feat/search improvements

rorychatt:
Fixing react hook violation issue when using search


rorychatt:
Trashing this. Will rework it later.

PR #37, opened 2025-06-28
feat: fix search crash bug

rorychatt:
Fixed issue where sometimes search crashed.

Added new feature where it is now possible to select items with keyboard (start pressing up or down arrow keys after inputting something, and enter to open selected component)

https://github.com/user-attachments/assets/a0a42fa3-7b23-466f-a5af-0439cbb3bd42
https://ivy-interactive.slack.com/archives/C093A6GK5D1/p1751133619171709

### DIFF
--- a/frontend/src/widgets/layouts/SidebarLayoutWidget.tsx
+++ b/frontend/src/widgets/layouts/SidebarLayoutWidget.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect, useCallback, useMemo } from 'react';
 
 import {
   Sidebar,
@@ -212,8 +212,9 @@ export const SidebarMenuWidget: React.FC<SidebarMenuWidgetProps> = ({
   // Register only the sidebar menu container with useFocusable
   const { ref: focusRef } = useFocusable('sidebar-navigation', 1);
 
-  // Flatten items for keyboard navigation in search mode
-  const flatItems: FlatMenuItem[] = searchActive ? flattenMenuItems(items).filter(i => !i.isGroup) : [];
+  const flatItems: FlatMenuItem[] = useMemo(() => {
+    return searchActive ? flattenMenuItems(items).filter(i => !i.isGroup) : [];
+  }, [searchActive, items]);
 
   useEffect(() => {
     setSelectedIndex(0);


frontend/src/widgets/layouts/SidebarLayoutWidget.tsx
```diff
@@ -156,10 +199,94 @@ const renderMenuItems = (items: (MenuItem)[], eventHandler:WidgetEventHandlerTyp
...
+  items,
+  searchActive = false
 }) => {
   const eventHandler = useEventHandler();
-  return renderMenuItems(items, eventHandler, id, 0)
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const menuRef = sidebarMenuRef;
+
+  // Flatten items for keyboard navigation in search mode
+  const flatItems: FlatMenuItem[] = searchActive ? flattenMenuItems(items).filter(i => !i.isGroup) : [];
```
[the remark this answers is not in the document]
rorychatt: A good suggestion... since elements don't change and the list is constant... But what if it's not...

Might add this later

frontend/src/widgets/inputs/TextInputWidget.tsx
```diff
@@ -285,15 +286,25 @@ const SearchVariant: React.FC<{
     onChange(e);
   };
 
-  const styles:React.CSSProperties = {
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
+      sidebarMenuRef.current?.focus();
```
nielsbosma: Can we solve this in a more general way? Seems dirty to reference sidebar here.

rorychatt: @nielsbosma 

I will have a look. I agree this looks dirty. 
I was solving the issue that to navigate out, an extra user interaction step was needed. With this, was able to bypass this.

rorychatt:
@nielsbosma If I were to make a more generic solution, then it would look somewhat like this.

It comes down the point that we need an event handler, which we have on the backend, and then we need a FE based focus management system.

My guess is that solution could be then creating a focus management hook that can then be used by other components with following functionality:

- Generic focus management hook that can be used by any component
- Registry-based system that tracks focusable elements
- Priority-based ordering for focus navigation
- No hardcoded dependencies on specific components

We could then later expand this and also use in other places.

But having looked through the previous solution and tried this one out, if we don't want this focusable hook, old code worked fine

But make sure to try out using the search. I think this really is cool to be able to navigate this way!

PR #38, opened 2025-06-28
Feat/color input component

rorychatt:
image


rorychatt:
I need windows machine for testing, will work on this from home.

rorychatt:
Okay so I should just scrap this PR.

With latest updates to packages, there is a lot more easier way to implement most of the features we need

PR #52, opened 2025-07-01
feat: switch ivy port to 5010

rorychatt:


Ivy/Server.cs
```diff
@@ -398,7 +398,7 @@ public static class IvyServerUtils
 {
     public static ServerArgs GetArgs()
     {
-        var portOption = new Option<int>("--port", () => 5000);
+        var portOption = new Option<int>("--port", () => 5010);
```
rorychatt: I actually see now that we are hardcoding the port value here

PR #57, opened 2025-07-01
opening already opened creates dublicate

rorychatt:
Now if I open any component, like BoolInput, by opening it again I do not create dublicates.

Meaning if I search for something, I can only have 1 tab of that in the list.

If I still try to open it, we will open existing tab, and not create another, identical one.

### DIFF
--- a/Ivy.Auth.Supabase.Test/Program.cs
+++ b/Ivy.Auth.Supabase.Test/Program.cs
@@ -1,6 +1,7 @@
 ﻿using System.Globalization;
 using Ivy;
 using Ivy.Auth.Supabase;
+using Ivy.Chrome;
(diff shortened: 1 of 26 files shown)


nielsbosma:
@rorychatt  There might be a cases in some admin where you want this - like in customer case where you have multiple customer views open. Make it configurable in ChromeSettings.

rorychatt:
> @rorychatt There might be a cases in some admin where you want this - like in customer case where you have multiple customer views open. Make it configurable in ChromeSettings.

Good point. I looked at the solution, let's try adding a AllowTabDuplicates = False.

But I also have some other thoughts about this, let's discuss at the office

PR #106, opened 2025-07-07
feat: number input display currency

ArtemKhvorostianyi:


frontend/src/widgets/inputs/NumberInputWidget.tsx
```diff
@@ -157,7 +157,7 @@ const NumberVariant = memo(
           max={max}
           step={step}
           format={formatConfig}
-              placeholder={placeholder}
+                placeholder={placeholder}
```
rorychatt: need to look into why precommit hook didn't work on your machine

ArtemKhvorostianyi:
#92 (number-input): show currency signs on slider values

PR #129, opened 2025-07-09
Select Input

sudipto80:
- Possible concern 
  MultiSelect doesn't work implicitly for Toggle variant always.


Ivy/Widgets/Inputs/SelectInput.cs
```diff
@@ -25,7 +25,7 @@ public abstract record SelectInputBase : WidgetBase<SelectInputBase>, IAnySelect
     [Prop] public string? Invalid { get; set; }
     [Prop] public string? Placeholder { get; set; }
     [Prop] public SelectInputs Variant { get; set; }
-    [Prop] protected bool SelectMany { get; set; } = false;
+    [Prop] public bool SelectMany { get; set; } = false;
```
rorychatt: ahaa. It's this one.

One of my recent PR-s was about this. If we need to select many, users should use array types.

ie UseState(string) would be for single select, and UseState(string[]) would be for plural.

rorychatt: We can discuss this during todays call

Ivy.Docs/Docs/02_Widgets/02_Inputs/Select.md
```diff
@@ -1,81 +1,394 @@
...
+        
+        return Layout.Vertical()
+            | Text.Label("Normal SelectInput:")
+            | normalSelect.ToSelectInput(options)
+                .Placeholder("Choose an option...")
+            
+            | Text.Label("Invalid SelectInput:")
+            | invalidSelect.ToSelectInput(options)
+                .Placeholder("This has an error...")
+                .Invalid("This field is required")
```
rorychatt, line 271: like in some other examples, we can make disabled and invalid field conditional. So when you open the component, fields are enabled / disabled, but it is possible to fix this.

So probably it should be a condition like it was done is previous PR-s

rorychatt (approved):
Good examples.

I like the coffee one. I will demo it to our team on Friday.

I will also open an issue and fix if there is bug with state that you had encountered

PR #134, opened 2025-07-10
overall fixes after library upgrades

rorychatt:
new:

Toggle input gets green background when it is selected
image

Invalid icon code input better placement

image

Fixed spacings in feedback, now using theme vars too

image

Number inputs take full width and have icons aligned

image

Similar changes to select/multiselect input

image

Text input default full width

image

Minor styling changes for grids in diffent apps, like use of inlinecodes instead of code blocks


rorychatt:
Multi Select needs more love. Or maybe I need to rework the way I test it in the samples, some cases look odd. This should not be done in this PR

PR #146, opened 2025-07-11
feat: add show copy button to code input

rorychatt:
image


rorychatt:
Will need to write a test for invalid + copy button

PR #161, opened 2025-07-12
(docs): installation rework

rorychatt:
Before:

image

Now:

image


rorychatt:
Note that I removed > from the md code for terminal sections.

Arguably, this should not be done here and should be configurable. Idea is that every terminal should have this.

rorychatt:
The `>` should probably be part of terminal component and not be selectable.

PR #164, opened 2025-07-12
better alignments and centerings

rorychatt:


rorychatt:
I failed to reproduce original issue. Maybe it got fixed over time with other PR-s.

PR #173, opened 2025-07-14
feat: file input size and format features

rorychatt:
image
image


rorychatt:
We should note that max file count and filetype is being checked right now in the frontend,

And we should move the logic to the backend.
So FileInput.cs should be responsible for tracking max allowed amount of files vs current amount of uploaded files, and the same for file format.

PR #231, opened 2025-07-18
Pie chart

sudipto80:
There are several things that can enhance the PieChart

1. An onchange event. If a certain pie is clicked it can return which part. Can really enhance the user experience in Drill Down situation (Please see example, Right now it feels a little awkward). The idea is people can click on the external donut representing countries and can see the inner pie. 
2. Possibly making a toggle to make a pie stand out - Nice to have. But will make it look awesome
3. The ChartStyles apart from Dashboard still shows line like markers in legends. Kindly look.

### DIFF
--- a/Ivy.Docs/Docs/02_Widgets/06_Charts/PieChart.md
+++ b/Ivy.Docs/Docs/02_Widgets/06_Charts/PieChart.md
@@ -2,17 +2,80 @@
 
 Pie charts represent parts of a whole. Each slice is drawn from the provided data.
 
-```csharp
-var data = new[]
-{
-    new { Month = "January", Desktop = 186, Mobile = 100 },
-    new { Month = "February", Desktop = 305, Mobile = 200 },
-    new { Month = "March", Desktop = 237, Mobile = 300 },
-};
+```csharp demo-below
 
-new PieChart(data)
-    .ColorScheme(ColorScheme.Rainbow)
-    .Pie("Mobile", "Month")
-    .Tooltip()
-    .Legend();
+public class PieChartDemo : ViewBase 
+{    
+    
+    public override object? Build()
+    {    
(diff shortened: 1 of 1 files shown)


Ivy.Docs/Docs/02_Widgets/06_Charts/PieChart.md
```diff
@@ -2,17 +2,247 @@
...
+            new { Country = "United States", Population = 333 },
+            new { Country = "Sweden", Population = 10 },
+            new { Country = "China", Population = 1412 },
+            new { Country = "Brazil", Population = 215 },
+            new { Country = "Canada", Population = 38 },
+            new { Country = "Germany", Population = 83 }
+        };
+
+        // Population data for drill-down chart
+        var populationData = new []
```
rorychatt, line 134: for readabilty, maybe it is better to declare a helper function that you move towards the end of file, like generate StatePopulationData(), and call it here.

So users don't have to scroll all the way down to see what actually brings value

rorychatt:
what you describe as OnChange should realistically be called OnClick -> OnChange is supposed to be called when something happens to a form field value, which can be input, for example.

But good idea with the user interact. This is a feature, however, and current plan is to fix bugs, and not add more features for the weekend.

I will create a comprehensive Todo for this OnClick behavior

rorychatt:
same goes for the second point - it could and should be handled via OnClick -> that needs to be implemented for this

rorychatt:
3 -> not sure if it is a bug, actually. You can override the styling for legend markers if you really want to. But I will look into this, if it is actually working as expected

rorychatt (changes requested):
See my comments

sudipto80:
Hello., added more example and showing more options

PR #263, opened 2025-07-23
do not merge: Cleanup e2e tests

rorychatt:
@nielsbosma I used cursor and set up a playwright MCP server. 
On second attempt where I made the right prompt, it just fixed the tests in the intended way.

total time spent: 15 minutes

FYI here is the prompt:

```text
Hey.

I have e2e tests that I have skipping now. Before our solution used iframes, now we got rid of it.
We have a lot of retries that we added when we had iframes.

I want you to clean that shit up and remove the retries.
Then I want you to make the tests pass, 1 by 1.

After you fix each test, I want you to commit it. Make sure to from start switch to a new feature branch

### DIFF
--- a/frontend/e2e/bool-input.spec.ts
+++ b/frontend/e2e/bool-input.spec.ts
@@ -1,13 +1,7 @@
-import {
-  test,
-  expect,
-  type Frame,
-  type Page,
-  type ElementHandle,
-} from '@playwright/test';
+import { test, expect, type Page } from '@playwright/test';
 
 // Shared setup function
-async function setupBoolInputPage(page: Page): Promise<Frame | null> {
+async function setupBoolInputPage(page: Page): Promise<void> {
   await page.goto('/');
   await page.waitForLoadState('networkidle');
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: frontend/e2e/bool-input.spec.ts [hunk 1]
```diff
@@ -28,75 +22,31 @@ async function setupBoolInputPage(page: Page): Promise<Frame | null> {
...
-  return contentFrame;
 }
 
 // Shared verification function for all elements
-async function verifyAllElements(appFrame: Frame) {
-  // Now look for the h1 in the app frame
-  const h1 = appFrame.locator('h1');
+async function verifyAllElements(page: Page) {
+  // Look for the h1 in the page
+  const h1 = page.locator('h1');
```
rorychatt: Spots like this perhaps should also use data test-ids, but for now this works.
H1 should be individual per page anyways.

H2-s are not unique per page, so this is "smelly", but it will work unless we completely change the structure.
Will address it if the problem comes

frontend/e2e/bool-input.spec.ts
```diff
@@ -202,164 +172,58 @@ test.describe('Bool Input Tests', () => {
...
-        test.skip('invalid state should toggle correctly', async () => {
-          await testTrueFalseToggle('toggle-true-state-width-invalid');
-        });
-      });
-    });
+  test('should verify disabled states', async ({ page }) => {
+    await testDisabledState(page, 'checkbox-true-state-width-disabled');
+    await testDisabledState(page, 'switch-true-state-width-disabled');
+    await testDisabledState(page, 'toggle-true-state-width-disabled');
   });
```
rorychatt, line 298: it might have gotten rid of some tests actually...

rorychatt:
Did a follow-up prompt

```text
the fields dont really have a default value. 

And we are not really testing interaction.
We still need to test for all the test ids like before.
Do not remove the test ids.

Group tests together by using describe. After each describe is done, create a commit message

nielsbosma:
Demo this tomorrow!

rorychatt:
commit id before the merge for demo purposes:
71a43fef505a29955dc17a26a0b18f1580d61892

PR #298, opened 2025-07-27
feat: add GlideDataGrid component 

dcrjodle:
- Implemented frontend/src/widgets/glide-data-grid/ folder with GlideDataGridWidget component
  - Using a non-monolithic approach (decomposed component)
  - Hook based decomposition
  - Children pattern
  - Separation of concern
- Added @glideapps/glide-data-grid dependency to package.json
- Added Ivy/DataTables/GlideDataGridBuilder.cs and GlideDataGridView.cs for backend grid functionality
- Created Ivy/Widgets/DataTables/GlideDataGrid.cs widget definition
- Updated widget registry and service integrations for new data grid functionality

Note: The GlideDataGrid package doesn't have a npm version compatible with React 19, so I've created the .npmrc file to bypass this

Screenshot 2025-07-27 at 13 23 47


frontend/src/widgets/glide-data-grid/styles/glideTheme.ts
rorychatt, line 1: we should maybe not use hard coded colors, we have a lot of colors already defined in root .css file

frontend/src/widgets/glide-data-grid/README.md
rorychatt, line 1: we didn't document any FE widgets in Ivy. Maybe we could do this more.

But the idea is that end users don't need to touch the FE code at all. Unless you are developing new components, of course

frontend/.npmrc
rorychatt, line 1: need to maybe investigate more the react 19 incompatibility for this package.

frontend/src/widgets/glide-data-grid/README.md
dcrjodle: Since this was simply generated, I removed it for now

PR #316, opened 2025-07-29
documentation: styling and bug fixes across several pages

sudipto80:
Many changes as per this 

Screenshot 2025-07-28 at 6 54 41 PM

* Select is still pending. Didn't understand the pattern 
* Text (Ctrl + N) shortcut is not clear.


rorychatt:
@sudipto80 Ctrl+N is a bad example because it is a common shortcut used in browsers, so we should show some other shortcut

Issue #352, opened 2025-08-04
[Bug]: updated terminal has no longer got perfectly centered icons


rorychatt:
### Description

After we updated the fonts, the `>` is no longer perfectly centered

### Code Samples

```csharp

```

### Screenshots

<img width="844" height="169" alt="Image" src="https://github.com/user-attachments/assets/0b34e146-c430-47aa-bafb-c1aeb18ee240" />

rorychatt:
@ArtemKhvorostianyi have a look

rorychatt:
resolved

PR #363, opened 2025-08-04
[Bug]: updated terminal has perfectly centered icons #352

ArtemKhvorostianyi:
Screenshot 2025-08-04 at 23 26 01


rorychatt:
@ArtemKhvorostianyi can you share an image of what it looks like all together? Kind of hard to tell from the reference you shared

rorychatt:
@ArtemKhvorostianyi I think the > needs to just be moved up a bit?
image
This is how it is now on main

PR #384, opened 2025-08-06
feat: increase delay

rorychatt:


rorychatt:
We need a proper solution.

This worked way better than before due to page loading optimizations we had, but I did notice that on a really slow laptop, in some cases the 500 ms didn't cut it.

PR #407, opened 2025-08-07
(docs): create tooltip docs

rorychatt:


rorychatt:
@nielsbosma this was in docs before, and I think we should remove this and show examples of .WithTooltip() as it is perhaps better than nested objects?

image

PR #456, opened 2025-08-09
(docs): implement chat documentation

ArtemKhvorostianyi:
Screenshot 2025-08-10 at 00 45 05


rorychatt:
In the image, I believe there is too much text for essentially the same two examples.
Demo this during today's docs writing workshop

PR #486, opened 2025-08-11
feat: support for mermaid in docs

rorychatt:


rorychatt:
@dcrjodle IDK, i think originally had the solution without it and it didn't work.
I'll keep this extra layer for now, if it bites in the future, should be easy refactor

PR #512, opened 2025-08-12
CLI Docs Improvements

zachwolfe:


Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/02_Init.md
```diff
@@ -121,8 +99,31 @@ Global using statements for common Ivy namespaces:
 
 ```csharp
 global using Ivy;
```
rorychatt, line 102: maybe we should create an issue about this. This is super easy to get outdated and should maybe be auto-generated or improted.

zachwolfe: Agreed!

zachwolfe: @rorychatt I opened the auto-generation issue here: #530

I agree that it would be nice to simplify this output as well.

PR #551, opened 2025-08-14
bug: better handle empty nodes

rorychatt:
this does not solve the issue.

It just makes so the page does not crash, and it provides more logging now. Real issue should be addressed at some point.


rorychatt:
@claude review this

rorychatt:
@claude review this with new changes

PR #646, opened 2025-08-25
(docs): implement Widgets doc

ArtemKhvorostianyi:
Screenshot 2025-08-26 at 00 19 15

Not finished yet
@rorychatt, need your opinion about this type of impelementation of this doc.


rorychatt:
Good starting point, merging this so we can polish it

PR #677, opened 2025-08-27
(docs): (contributors): display actual names not nicknames

rorychatt:


rorychatt:
git commit commit always has author, git commit author not always. I think this is correct git logic

PR #705, opened 2025-08-30
(Samples): (Products): fix gaps in product view(edit)

ArtemKhvorostianyi:
Screenshot 2025-08-30 at 14 51 49

The issues was in the FE part of FooterLayout. Edit Product window is a sheet, and it's content inits as a footer content. Footer sets extra padding for content above, so this change will fix all the sheets visibility, too (I saw this issue in sheet doc, which had footer usage)

Sheets with footer usage now:
Screenshot 2025-08-30 at 14 55 37
The card in it also had extra padding before

And because of it footer with content mentions changed:
Screenshot 2025-08-30 at 14 57 16
I think it's correct and more natural now, btw
@rorychatt , need your thought


rorychatt:
seems to be the case, after we introduced the scroll area and fixed external padding.

PR #737, opened 2025-09-03
(docs):(SidebarLayout): fix UI in SideBarLayout doc

ArtemKhvorostianyi:
Screenshot 2025-09-04 at 00 13 14


rorychatt:
hey good fix

We should not use separators for header/footer in sidebar as you get the visual line automatically based on the structure

rorychatt:
@ArtemKhvorostianyi I need you to double check this file that we don't do this in other examples

PR #776, opened 2025-09-07
(UI): Fix LineCharts visibility

ArtemKhvorostianyi:
Screenshot 2025-09-07 at 16 12 08

Because of no margins in FE of LineCharts, it cutted data from every side: top, bottom, right, left
Added margin 10 for every side

Screenshot 2025-09-07 at 16 08 19

Also added a few examples in Samples to test strange data visibility.

Screenshot 2025-09-07 at 16 15 03


rorychatt:
@ArtemKhvorostianyi the margin you added needs to be consistent between all chart types
ie `linechart`, `piechart`, `barchart` and the 4th one that I forgot the name of.

Also see if you can make it smaller and use tailwindcss class like `m-2` instead which is better in responsiveness (for this case abs px count is fine too, but `m-2` is better for consistency both in code style and the UI (we use m-2 where we have internal gaps between already gapped elements)

PR #786, opened 2025-09-09
(forms): document importing

rorychatt:
image


rorychatt:
@niels do we want to add statements like how to import the specific feature? Or do we say that everything is imported for you in the glbal usings? (I think we should add this info anyways)

rorychatt:
as discussed, ban this!

PR #788, opened 2025-09-09
(UI): fix text usage in the cards

ArtemKhvorostianyi:
Screenshot 2025-09-09 at 14 42 30

There is no need to use Text.H3 in content of cards, because this class of text has margin that brake horizontal layout


rorychatt:
@claude review this

PR #793, opened 2025-09-09
(charts): fix charts margins

ArtemKhvorostianyi:
Screenshot 2025-09-09 at 17 56 21

Removed mt-4 from divs
added margins to other charts to, so it looks well everywhere now and there is no data that been cutted
Screenshot 2025-09-09 at 17 58 06
Screenshot 2025-09-09 at 17 58 30
Screenshot 2025-09-09 at 17 58 45


rorychatt:
@ArtemKhvorostianyi can you resolve this space?

image

I know this happens due to y axis labels and they need space, but this looks weird that we have this space to the left always

rorychatt:
@ArtemKhvorostianyi Also in this case if it is a chart in a card we have a really big space at the bottom. We need to fix this. Could you look into it?

rorychatt (changes requested):
see my comments above

rorychatt:
@claude review this

rorychatt:
@ArtemKhvorostianyi see feedback by claude. We indeed need something similar

rorychatt:
@ArtemKhvorostianyi don't do any changes to charts for now, we will see if we replace the library we use competely first.

PR #801, opened 2025-09-09
(Footer): fix cutted border of footer

ArtemKhvorostianyi:
Screenshot 2025-09-10 at 00 34 47


rorychatt:
@claude review this

rorychatt:
@ArtemKhvorostianyi  this is a scary commit due to -mx-4.

Is there another way to solve this?

rorychatt (changes requested):
See if there is another way to solve this. Maybe this is good, but it feels sketchy that you have to do negative spacings. Can this be done as positive value to the parent instead?

ArtemKhvorostianyi:
@rorychatt , I implemented solution via removing parent padding for border

rorychatt:
@ArtemKhvorostianyi let's see if anything breaks!

PR #812, opened 2025-09-12
(docs): fix incorrect paths for docs regen instructions

svetlana-soboleva:
Updated the path from `cd Ivy.Docs` to `cd Ivy.Docs.Shared` in the documentation generation instructions.

This change ensures that users follow the correct steps when generating documentation files.


rorychatt:
@svetlana-soboleva congrats on first small, but valuable contribution!

PR #817, opened 2025-09-12
[Feat]: implement submitting of auth form with return/enter

ArtemKhvorostianyi:
Screenshot 2025-09-13 at 01 44 18

### DIFF
--- a/Ivy/Auth/DefaultAuthApp.cs
+++ b/Ivy/Auth/DefaultAuthApp.cs
@@ -101,9 +101,9 @@ public class PasswordEmailFlowView(IState<string?> errorMessage) : ViewBase
 
         return Layout.Vertical()
          | Text.Label("User:")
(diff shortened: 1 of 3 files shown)


ArtemKhvorostianyi:
Idea was that we want to submit auth with clicking enter/return key
Added event that suppose to realise it
Hadn't implement testing of it
@rorychatt , need your advice, how can we test if it works

rorychatt:
I think if we press enter we want to blur our the current active input.
If we press enter in forms and the form is valid, the form should submit.

we don't want any input to "submit" if we press enter. "Submit" behavior is responsibility of a form.

@ArtemKhvorostianyi Also see in the BE if we have an enum for the keyboard event key names.

rorychatt:
@ArtemKhvorostianyi on screenshot its a random layout not a form.

PR #824, opened 2025-09-14
(button): add new button variants (Success, Warning, Info) and update styles

svetlana-soboleva:
## Summary

- Added new button variants: Success, Warning, Info
- Updated ButtonApp to support and display all variants
- Improved button layout: switched to wrapping layout for better spacing when there are many buttons
- Updated styles to prevent buttons from being squeezed

## Motivation

With more button variants, the previous grid layout was too crowded. Wrapping and spacing improvements make the UI more readable and visually balanced.

### DIFF
--- a/Ivy/Widgets/Button.cs
+++ b/Ivy/Widgets/Button.cs
@@ -27,12 +27,18 @@ public enum ButtonVariant
     Outline,
     /// <summary>Secondary button style with reduced visual emphasis, typically used for supporting actions or less prominent interactions.</summary>
     Secondary,
+    /// <summary>Success button style with positive action colors, typically used for confirming actions or indicating success.</summary>
+    Success,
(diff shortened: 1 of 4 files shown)


Ivy/Widgets/Button.cs
rorychatt, line 1: better xml docs for colors

There has to be a better way of describing specific colors in the xml docs (shorter ones)

svetlana-soboleva: I have been following the pattern already provided. Shall i refactor this documentation to make it shorter, more concise comments to reduce verbosity and improve readability?

rorychatt: hey I will fix this myself but the idea is that xml docs should be minimal yet yield all neccessary info

PR #849, opened 2025-09-16
[Fix]: improve forms autofill style 

ArtemKhvorostianyi:
We had an issue that on autofilling forms, which presented in a card, form field had black background color. Fixed

How it was:
Screenshot 2025-09-16 at 15 08 21
How it now:
Screenshot 2025-09-16 at 15 25 59

Implemented solution with additional css rule in the index.css, where other rules for autofill were implemented

### DIFF
--- a/frontend/src/hooks/use-autofill.ts
+++ b/frontend/src/hooks/use-autofill.ts
@@ -9,25 +9,8 @@ const observedInputs = new WeakMap<
   Set<HTMLInputElement | HTMLTextAreaElement>
 >();
 
-// Performance monitoring (can be removed in production)
-let observerCreationCount = 0;
-let inputRegistrationCount = 0;
-
-// Simple logger that can be easily disabled
-const logger = {
-  info: (message: string) => {
-    // Logging disabled by default - uncomment next line to enable during development
-    // console.log(message);
-    void message; // Suppress unused parameter warning
-  },
-};
-
 function getGlobalResizeObserver(): ResizeObserver {
   if (!globalResizeObserver) {
-    observerCreationCount++;
-    logger.info(
-      `🔍 Created global ResizeObserver (${observerCreationCount} total observers)`
-    );
     globalResizeObserver = new ResizeObserver(entries => {
       for (const entry of entries) {
         const container = entry.target as HTMLElement;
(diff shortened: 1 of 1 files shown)


ArtemKhvorostianyi:
@claude review this

rorychatt:
image
@ArtemKhvorostianyi we need to clean up the css file.

See this feedback by claude, it correctly points out that we essentially have 160+ lines of code that are overridden by this.
We need to remove as much as possible

rorychatt (changes requested):
See comments above

ArtemKhvorostianyi:
Screenshot 2025-09-17 at 12 45 43

Screenshot 2025-09-17 at 12 45 56

I’ve written a small utility to handle autofill colors instead of overloading index.css.

– It now handles all autofills, not only those for forms inside cards.

– It reduces the amount of code required in index.css (kept to a minimum).

– Tested in both dark and light themes — works fine.

@rorychatt, I’d like your feedback on this approach. I added some small helper functions and if statements in textArea.ts and input.ts, and also created a new file, auto-fill.ts. Please let me know if you approve of these changes.

rorychatt:
@ArtemKhvorostianyi How well does this work if you have hundreds of inputs? Won't this tank performance as you create so many observers?

ArtemKhvorostianyi:
Optimised solution with using different approach of creating observers

rorychatt:
@dcrjodle how would you tackle this?

frontend/src/hooks/use-autofill.ts
dcrjodle, line 1: This is a lot of code for such a simple problem. Maybe I'm missing something, but if we're trying to make sure the autofilled text have correct styling, maybe this is a bot of over-engineering
What about something more simple like, depending on what the styling problem is

```
input:-webkit-autofill {
  -webkit-box-shadow: 0 0 0 1000px var(--input-bg) inset !important;
  -webkit-text-fill-color: var(--input-text) !important;
}
```

dcrjodle: Or if you want to use a hook for solving the text styling problem, you can 
```
function useAutofillStyling() {
  const inputRef = useRef(null);
  
  useEffect(() => {
    const input = inputRef.current;
    if (!input) return;
    
    // Simple one-time styling fix
    input.addEventListener('input', () => {
      if (input.matches(':-webkit-autofill')) {
        // Apply fix
      }
    });
  }, []);
  
  return inputRef;
}
```

dcrjodle: If possible, I would prefer to keep the code more simple. Let's not create a solution for something unless we absolutely need it. You PR doesn't really provide much information to what problem this hook is fixing. maybe you could provide a video to explain more the styling issue?

PR #859, opened 2025-09-17
(select): smarter scrollbars

rorychatt:
This pull request makes a small change to the rendering logic for the checkbox variant in the `SelectInputWidget` component. The update ensures that the scrollable styling is only applied when there are more than six options, improving the user experience for shorter lists.


rorychatt:
@claude  review this

Issue #827, opened 2025-09-15
(epic) Responsive Design


nielsbosma:
## Problem

Ivy Framework lacks a comprehensive responsive design system. While the framework has flexible sizing (Full, Fit, Fraction, Min/Max), AspectRatio support, and layout improvements, there is no way to:

1. **Define breakpoints** - Cannot specify different property values at different screen sizes (like Tailwind's `md:`, `lg:` prefixes)
2. **Control visibility responsively** - Cannot hide/show widgets based on viewport size
3. **Adjust layout orientation** - Cannot switch `StackLayout` from horizontal (desktop) to vertical (mobile)
4. **Change sizing responsively** - Cannot adjust `Width`, `Height`, `Gap`, `Padding` at different breakpoints
5. **Modify density** - Cannot change `Density` based on screen size

Currently, the frontend has a `useIsMobile` hook (768px breakpoint) but no C# API exposes responsive capabilities to developers.

Many closed issues (#2415, #2388, #2343, #2303) have improved responsive foundations, but a unified breakpoint system is still missing.

## Solution

Create a comprehensive responsive design system with breakpoints, responsive properties, and conditional visibility.

### 1. Define Breakpoint System

Create `Breakpoint` enum matching common device categories:

```csharp
// src/Ivy/Shared/Breakpoint.cs
public enum Breakpoint
{
    Mobile = 640,      // sm in Tailwind
    Tablet = 768,      // md in Tailwind
    Desktop = 1024,    // lg in Tailwind
    Wide = 1280        // xl in Tailwind
}
```

### 2. Create Responsive<T> Wrapper Type

Add a generic wrapper that stores values per breakpoint:

```csharp
// src/Ivy/Shared/Responsive.cs
public record Responsive<T>
{
    public T? Mobile { get; init; }
    public T? Tablet { get; init; }
    public T? Desktop { get; init; }
    public T? Wide { get; init; }

    // Implicit conversion from single value (applies to all breakpoints)
    public static implicit operator Responsive<T>(T value)
        => new() { Mobile = value, Tablet = value, Desktop = value, Wide = value };
}

public static class ResponsiveExtensions
{
    public static Responsive<T> At<T>(this T value, Breakpoint breakpoint)
    {
        return breakpoint switch
        {
            Breakpoint.Mobile => new Responsive<T> { Mobile = value },
            Breakpoint.Tablet => new Responsive<T> { Tablet = value },
[...]

PR #877, opened 2025-09-19
dashboard grid responsive done (#827)

Jeelislive:
Screenshot from 2025-09-20 00-53-53
Screenshot from 2025-09-20 00-54-06

### DIFF
--- a/frontend/src/widgets/layouts/GridLayoutWidget.tsx
+++ b/frontend/src/widgets/layouts/GridLayoutWidget.tsx
@@ -15,6 +15,8 @@ interface GridLayoutWidgetProps {
   childRow?: (number | undefined)[];
   childRowSpan?: (number | undefined)[];
   className?: string;
+  /** Optional minimum column width in pixels used to auto-resize columns responsively. Defaults to 280. */
+  minColumnWidthPx?: number;
 }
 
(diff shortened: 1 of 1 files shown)


rorychatt:
Hey @Jeelislive 

i will need to look at what you did in the morning,

but keep in mind that defining grids in a FE-based solution and using responsive widths like this when Width is set from the backend.

I think that in general we should not optimize views for Mobile yet, especially use charts as reference, as in the end we might opt in for another charts package, and then stuff might break.

Jeelislive:
@rorychatt okay understood that we don't have need responsiveness as of now.

Jeelislive:
@rorychatt 
@nielsbosma  is good to go? or i can close if not needed as of now

rorychatt:
@nielsbosma I am not convinced that we should merge this.

We have WrapLayout that can have similar behavior is user properly sets the Min/Max widths for cards.
Otherwise solution code-wise is good.

Merge this is if you think its okay.
Here is demo of what the branch does:

https://github.com/user-attachments/assets/5d8ef185-1a5e-4767-963e-27b6f0c7f877

rorychatt:
After some careful consideration, we figured that we can not merge this now as we need to come up with a BE defined responsiveness system

PR #878, opened 2025-09-19
[DropDownMenu]: fix subcontent overlay

ArtemKhvorostianyi:
Here is how it was: 
Screenshot 2025-09-19 at 22 25 15

Fixed one:

Screenshot 2025-09-19 at 22 54 22

Was trying to use minimum of changes to handle it right. 
In the dropdown-menu we already use sideoffset, so I think that's ok to add it as a part of a solution for subcontent.

To check it, go to Ivy.Docs -> DropDown Menu -> Nested Menu Items section


rorychatt:
can you drag mouse from popup A to popup B? 
Have you tried a design where there is a gap between the popups? (right now they share a side)

@ArtemKhvorostianyi

PR #901, opened 2025-09-21
Allow custom submit title in toForm() method

ShogunFire:
It allows the user to pass the text that is in the submit button directly in the toForm() method without having to do more work, I think it is usefull because other wise he would have to create the button himself and handle the click

Code example
DOC 2

Result:
DOC 1

### DIFF
--- a/Ivy.Docs.Shared/Docs/01_Onboarding/02_Concepts/Forms.md
+++ b/Ivy.Docs.Shared/Docs/01_Onboarding/02_Concepts/Forms.md
@@ -265,6 +265,25 @@ public class FieldManagementExample : ViewBase
 
 ## Validation
 
+### Custom Submit Text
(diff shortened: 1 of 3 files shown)


Ivy.Docs.Shared/Docs/01_Onboarding/02_Concepts/Forms.md
```diff
@@ -265,6 +265,25 @@ public class FieldManagementExample : ViewBase
 
 ## Validation
 
+### Custom Submit Text
+
+Change the text of the submit button by passing it as a parameter of the `.toForm()` method
```
nielsbosma, line 270: Should be .ToForm()

nielsbosma (changes requested):

Issue #879, opened 2025-09-19
[Bug]: excessive xml docs in DropDownMenu


ArtemKhvorostianyi:
### Description

<img width="703" height="841" alt="Image" src="https://github.com/user-attachments/assets/29dd0d86-f51c-4b0e-bd7c-b512dc1e0108" />

### Code Samples

```csharp

```

PR #906, opened 2025-09-23
[DropDownMenu]: fix sub content (nested menu) overlaying of main content

ArtemKhvorostianyi:
Fixed:
Screenshot 2025-09-23 at 10 34 26

Solution:
Added parametr SubSideOffset to DropDownMenu.cs
Screenshot 2025-09-23 at 10 45 13

GIve this parametr to frontend, dropdown-menu.tsx and DropDownMenuWidget.tsx
Screenshot 2025-09-23 at 10 47 59

Avoided hardcoding in frontend part,  but added parametr to backend

Screenshot 2025-09-23 at 10 48 38

Also reduced size of xml docs for DropDownMenu.cs

This PR should also close issue https://github.com/Ivy-Interactive/Ivy-Framework/issues/879 , it's about xml docs in the DropDownMenu.cs - done

### DIFF
--- a/frontend/src/components/ui/dropdown-menu.tsx
+++ b/frontend/src/components/ui/dropdown-menu.tsx
@@ -41,14 +41,13 @@ DropdownMenuSubTrigger.displayName =
 const DropdownMenuSubContent = React.forwardRef<
   React.ElementRef<typeof DropdownMenuPrimitive.SubContent>,
   React.ComponentPropsWithoutRef<typeof DropdownMenuPrimitive.SubContent>
->(({ className, sideOffset, ...props }, ref) => (
+>(({ className, ...props }, ref) => (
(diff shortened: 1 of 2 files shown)


nielsbosma:
@rorychatt Not sure I like the naming, that it's in pixels, and why do these need to be separate? 
Smells bad that this is needed. Ivy users should just get a nice UI with defaults - don't see a reason for why this is a setting at all. We should have settings for stuff that don't add any logic value.

ArtemKhvorostianyi:
@rorychatt , avoided using sideoffset in BE part, I decided to handle it with frontend, with margins. Please, review changes

Screenshot 2025-09-25 at 15 21 06

It looks exactly the same, I think this is a better solution of this problem

PR #915, opened 2025-09-23
(html): removing the Rendered result part and fix security formating

ShogunFire:
In the html doc there are a lot of parts called rendered result that are redondant with the demo, it's the only page in the doc where we have those rendered result parts, so I removed them, I also fixed the part on security that was badly formated


rorychatt:
@ShogunFire good catch!

PR #916, opened 2025-09-23
(feat): implement Size for File Input type

ArtemLaza:
image


rorychatt:
@ArtemLaza resolve the (i), greyed out text in last option and make sure there is some small padding inside so text does not directly touch corners

rorychatt (changes requested):
see comment above.

Also in this case maybe you should not set minimum heights in the FE because that should be BE defined 

Image

nielsbosma (approved):

rorychatt:
@dcrjodle how does this look for you? (ui change in the fe)

Issue #763, opened 2025-09-05
[Feature]: ivy hello app we need an :ivystar: custom emoji


rorychatt:
### Description

<img width="720" height="581" alt="Image" src="https://github.com/user-attachments/assets/8e1bc401-81ec-49ed-bd29-e9c545b3aced" />

We need a custom emoji for a star that would be the correct Ivy green color

### Code Examples

```csharp

```

ShogunFire:
@rorychatt Not sure if this is what you want because it's a hack but it is possible to apply a filter with css to change the emoji's color, I did it only with chrome dev tools because styles don't seem to work yet in the Text.Html() method.

<img width="427" height="207" alt="Image" src="https://github.com/user-attachments/assets/868b9152-3ea0-4d70-a876-976177703e09" />

rorychatt:
@ShogunFire can you propose a solution? 

Would be cool if we could do like you had shown in the image, but instead of a star we put something like :ivy-branded-star: and it would get rendered as you had circled 

Maybe we can keep this solution just in the FE

ShogunFire:
Ok, I will try to create a way to have custom emojis in the MarkdownRenderer in the front end

PR #918, opened 2025-09-24
(emoji): add a custom emojis functionality in the markup renderer. 

ShogunFire:
Feature [issue 763](https://github.com/Ivy-Interactive/Ivy-Framework/issues/763)

Add custom emoji fonctionality in the markup renderer and add an example with a ivy branded star emoji in the hello sample

This is the result:
image


ShogunFire:
Do we also need to change the hello code generated by ivy init --hello command ? I didn't find where the code for ivy console is

rorychatt:
@ShogunFire this is amazing!

You're right, for security issues, CLI is not public. After we push this in the framework, I will make sure everywhere else we also update the code.

frontend/public/ivy-branded-star.svg
nielsbosma, line 1: Can you remove the black border.

nielsbosma (changes requested):

frontend/src/components/custom-emojis/emojiFunctions.tsx
```diff
@@ -0,0 +1,17 @@
...
+      Object.keys(emojiMap).map(k => k.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')) +
+      ')',
+    'g'
+  );
+
+  return text.replace(pattern, match => {
+    const emoji = emojiMap[match];
+    if (!emoji) return match;
+
+    return `<img src="${emoji.src}" width="${size}" height="${size}" style="display:inline-block;vertical-align:middle;" />`;
```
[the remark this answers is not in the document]
ShogunFire: He is right, my solution wasn't working with multiple custom emojis

frontend/public/ivy-branded-star.svg
ShogunFire: Done

rorychatt (changes requested):
The security warnings are valid.

You indeed can do funny stuff if you add it inside the : :

ShogunFire:
> The security warnings are valid. See if you can adress them yourself, else I can look into it.
> 
> You indeed can do funny stuff if you add it inside the : :

I thought it was ok because we only use the strings present in emojiMap.tsx (emoji.src) to create the html so no user input. 

If we need to sanitize can I add DOMPurify nuget ?

rorychatt:
@ShogunFire valid point. I will verify this with Niels tomorrow, and get back to you!

nielsbosma:
@ShogunFire @rorychatt given our foucs on security i don't want anything that can be flaged as a XSS in any scan - so for that reason make sure this is not an issue.

ShogunFire:
Got it, I did sanitization

rorychatt:
@ShogunFire We need to sanitize this on the BE side (FE side sanitization is good too, but not without BE one) -> could you implement this?

rorychatt (changes requested):
We also need the BE to sanitize the stuff in case someone bypassed the FE code.

ShogunFire:
@rorychatt Ah I didn't get that markup could accept html tags, so you want sanitization for the whole markup not just emojis image ? My frontend solution is not good because it deletes all the html tags that are not images. I checked the Html widget and there is no sanitization in the BE either, do you want to do it there too ?

rorychatt:
@ShogunFire My guess that the ultimate solution is to add sanitation layer around the whole Html widget in the BE now.
If you do that, it should also cover this `: XSS inject :` vulnerability

ShogunFire:
Finally I changed things a bit by doing things the right way: creating a remark plugin, now we don't return raw html anymore so there should be no xss problem flagged.

ShogunFire:
@rorychatt I changed the code to stop returning raw html and created a remark plugin instead so there should be no xss problem now

rorychatt:
Need to update all examples to use this new `:ivy-branded-star:`

PR #937, opened 2025-09-25
[Docs]: implement Volume docs

ArtemKhvorostianyi:
We didn't have documentation about File System and Storage
Screenshot 2025-09-25 at 14 09 31
This section was missed from Release v1.0.112 update, https://github.com/Ivy-Interactive/Ivy-Framework/releases/tag/v1.0.112
I added Volume doc page
Screenshot 2025-09-25 at 14 12 51

We actually need to test added features from release and test out code provided in this documentation on real Ivy Project to be sure it works correctly


rorychatt:
@claude review this

rorychatt:
@nielsbosma Do we want to reference this for File Uploads/Downloads? 
Is it good as a separate concept, or should we add examples with this to other docs?

nielsbosma:
> @nielsbosma Do we want to reference this for File Uploads/Downloads? Is it good as a separate concept, or should we add examples with this to other docs?

Let's wait a bit with that. I think we will use another interface for storage. This is for mounted volumes where you put db files.

PR #938, opened 2025-09-25
(git) Fix: In the precommit script, remove git add .

ShogunFire:
Because of the husky precommit script it was impossible to keep unstaged files and I was often committing files that I didn't want to commit.

I removed the line `git add .` This line was usefull to stage the changes on the .cs files made by `dotnet format` but it also staged all the other unstaged changes

So I replace it by doing `git add` only on the .cs files that were already staged. 

I also do dotnet format only on those staged files because it's more efficient.


rorychatt:
@ShogunFire awesome PR. I will test it out today-tomorrow

PR #948, opened 2025-09-26
Reverse order of TerminalWidget stoplight buttons

zachwolfe:
Nitpick: I think the red "button" should be in the corner of the widget

Before:
Screenshot 2025-09-26 at 3 47 24 PM

After:
Screenshot 2025-09-26 at 4 21 57 PM


rorychatt:
@zachwolfe good catch!

PR #959, opened 2025-09-27
[Embed]: implement embed handling for most of social media

ArtemKhvorostianyi:
Some socials like facebook and pinterest aren't availible link their embed
Read few arcticles about this, people say that this is because of its politics, but provided active links
Git hub embed handling is too large, the most of FE file size. I can reduce it, but we won't see proper demo of it.

Updated Samples for Embed
Screenshot 2025-10-02 at 13 22 53
Screenshot 2025-10-02 at 13 23 24

I initially set the frontend size the same way as YouTube — full width.
But now I think a better approach would be to add size options for embeds, so the user can decide how large they want the embed to be.


rorychatt:
@claude review this

rorychatt:
@ArtemKhvorostianyi fix feedback by copilot

rorychatt (changes requested):
fix feedback by copilot

ArtemKhvorostianyi:
@claude review changes

rorychatt:
@ArtemKhvorostianyi Can you update the description with all the latest screenshots?

Update the `Ivy-Samples` to include an example of each state of the `Embed` widget

ArtemKhvorostianyi:
@rorychatt Updated Samples for Embed
Screenshot 2025-10-02 at 13 22 53
Screenshot 2025-10-02 at 13 23 24

Presented it via two horizontal cards
Screenshot 2025-10-02 at 13 23 47

I initially set the frontend size the same way as YouTube — full width.
But now I think a better approach would be to add size options for embeds, so the user can decide how large they want the embed to be.

rorychatt (approved):

frontend/src/widgets/primitives/EmbedWidget.tsx
```diff
@@ -1,4 +1,274 @@
...
+    twttr?: {
+      widgets: {
+        load: () => void;
+      };
+    };
+  }
+}
+
+// Lazy load embed components for better performance
+const TwitterEmbed = lazy(() =>
```
dcrjodle: All of these are not lazy loaded since they are in the same file as the component. I would split the components out in to separate files, then lazy loading would be correct and we can have less code in this file

frontend/src/widgets/primitives/EmbedWidget.tsx
```diff
@@ -1,4 +1,274 @@
...
+  useState,
+  lazy,
+  Suspense,
+  Component,
+  ErrorInfo,
+  ReactNode,
+} from 'react';
+import Icon from '@/components/Icon';
+
+declare global {
```
dcrjodle: I don't think this is used anywhere

frontend/src/widgets/primitives/EmbedWidget.tsx
rorychatt: I am not completely sure on the lazy loading issue, as when I tried it out with a slow network I think it actually was being lazy loaded.

I do agree that splitting the responsibility is a good idea for future code maintentace too. (this also will fix the mentioned issue if that is the case)

@ArtemKhvorostianyi could you look into splitting the file?

frontend/src/widgets/primitives/EmbedWidget.tsx
```diff
@@ -1,4 +1,274 @@
...
+
+const sanitizeUrl = (url: string): string | null => {
+  if (!isValidUrl(url)) {
+    return null;
+  }
+
+  try {
+    const urlObj = new URL(url);
+    // Remove potentially dangerous URL components
+    urlObj.search = '';
```
dcrjodle: do we know that know that no urls need query params? 
We could also validate protocol/host and let provider-specific extractors decide what to keep

frontend/src/widgets/primitives/EmbedWidget.tsx
dcrjodle: All of these are not lazy loaded since they are in the same file as the component. I would split the components out in to separate files, then lazy loading would be correct and we can have less code in this file

ArtemKhvorostianyi: @dcrjodle Hi!
I splitted components to different files
Screenshot 2025-10-02 at 16 04 29
Folder with embeds contains all social media embeds implementation, error handlings cards and load logic in shared.ts
In shared.ts implemented better url handling, sanitazing, allowed hosts list
`export const sanitizeUrl = (url: string): string | null => {
  try {
    const urlObj = new URL(url);

    // Only allow http and https protocols
    if (urlObj.protocol !== 'http:' && urlObj.protocol !== 'https:') {
      return null;
    }

    // Validate hostname for known embed providers
    const allowedHosts = [
      'youtube.com',
      'youtu.be',
      'twitter.com',
      'x.com',
      'facebook.com',
      'instagram.com',
      'linkedin.com',
      'pinterest.com',
      'pin.it',
      'github.com',
      'gist.github.com',
      'reddit.com',
      'tiktok.com',
    ];

    const hostname = urlObj.hostname.toLowerCase();
    const isAllowed = allowedHosts.some(
      host => hostname === host || hostname.endsWith('.' + host)
    );

    if (!isAllowed) {
      return null;
    }

    return urlObj.toString();
  } catch {
    return null;
  }
};`

In main EmbedWidget file there are lazy loads and rendering of different embeds, each of them are defined in their own component
Screenshot 2025-10-02 at 16 08 45

Tested out lazy loads with throttling-> it works fine.

Thanks for your feedback, code looks much cleaner now

PR #963, opened 2025-09-28
(feat): URL state preservation

zachwolfe:


zachwolfe:
@claude review please

zachwolfe:
## Response to Claude (for the benefit of other humans)

### High Priority Issues

1. Code is correct here, comment is wrong. We don't use `appId` from the closure, we just get a fresh value each time.
2. Claude is confused by an outdated review comment again. (maybe I should start deleting old review comments before requesting new ones)
3. Possibly valid feedback, but that line was not added by this PR.

### Medium Priority Issues

4. Fixed in d84fca54525844ebc1d62c4e05534b901681c4f3
5. Not sure if this is worth fixing, but I did so anyway in 1b7eb1413f1abe2140ca543e06424b188c288615
6. Might be valid feedback, but unrelated to this PR. @rorychatt

### Low Priority Issues
7. Fixed in 8ffd907c8d9480997390c1f132e0ac2203385267
8. Fixed in cf4757739498fed5c8a9262e62beeb2fc564667f

frontend/src/hooks/use-backend.tsx
```diff
@@ -276,7 +305,7 @@ export const useBackend = (
         currentConnectionRef.current = null;
       }
     };
-  }, [appArgs, appId, machineId, parentId]);
+  }, [appArgs, stableAppId, machineId, parentId]);
```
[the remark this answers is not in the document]
zachwolfe: We don't always want a change in the chrome parameter to recreate the connection, I think. Either way, this did not come up as a problem in any of my testing.

rorychatt:
6 is valid. We can create an issue about this after we merge this one

Ivy/Chrome/Shared.cs
```diff
@@ -47,26 +48,54 @@ public static ChromeSettings DefaultApp<T>(this ChromeSettings settings)
...
+        if (this.AppId == null)
+        {
+            throw new InvalidOperationException("Cannot create AppHost: AppId is null.");
+        }
+
         return new AppHost(this.AppId, this.AppArgs != null ? JsonSerializer.Serialize(this.AppArgs) : null, parentId);
     }
 
     public string GetUrl(string? parentId = null)
     {
```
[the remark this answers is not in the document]
zachwolfe: That is precisely the intended behaviour

Ivy/AppHub.cs
```diff
@@ -182,6 +210,12 @@ public override async Task OnConnectedAsync()
                 LastInteraction = DateTime.UtcNow,
             };
 
+            if (appId != AppIds.Chrome && sessionStore.FindChrome(appState) == null)
+            {
+                var navigateArgs = new NavigateArgs(appId, Chrome: GetChromeParam(httpContext));
```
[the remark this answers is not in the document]
zachwolfe: In this case `Redirect` will only set the current browser history state, not redirect to a different page. So looping is not possible

rorychatt:
@zachwolfe I will try this out, looks super cool so far. I think this could be part of the next week's release.

rorychatt:
if frontend dist is missing, you get this

image

(not sure if this needs fixing, as you only end up here if you did not build frontend before)

rorychatt (approved):
Tested locally from the same project. Everything works as expected.

zachwolfe:
@rorychatt Great!

Also I’m not familiar with the current behaviour on `main` when `dist` is missing, but I guess that error message is probably fine. Maybe a future area of improvement would be to handle that case and say the frontend needs to be built.

I will wait for @nielsbosma’s review since he said he would have opinions on the server changes

frontend/src/hooks/use-backend.tsx
nielsbosma, line 308: Do we need to fix this?

Ivy/Chrome/Shared.cs
```diff
@@ -47,26 +48,59 @@ public static ChromeSettings DefaultApp<T>(this ChromeSettings settings)
 [Signal(BroadcastType.Chrome)]
 public class NavigateSignal : AbstractSignal<NavigateArgs, Unit> { }
 
-public record NavigateArgs(string AppId, object? AppArgs = null)
+public enum NavigationPurpose
+{
+    NewDestination,
```
nielsbosma, line 53: What is the definition of these? Can we make better names that are self describing.

frontend/src/hooks/use-backend.tsx
zachwolfe: If there is a problem here, I don't think adding `chrome` like Copilot suggested is the solution, because it will recreate the connection too often. I think. But I will look into it.

Ivy/Chrome/Shared.cs
zachwolfe: Sure. I will change this in a followup PR.

PR #981, opened 2025-09-29
(feat): Implement Size for Audio Recorder type

ArtemLazarchuk:
image


rorychatt (changes requested):
icons should scale too

Issue #988, opened 2025-09-30
(auth): Auth0 hits rate limit


ViktorWb:
### Description

When using Auth0, after reloading the page just a few times I get signed out because the ValidateJwtAsync function fails with rate limit error. The Auth0 authentication API has very strict rate limits.

ValidateJwtAsync should not need to call Auth0 at all, just verify the JWT against the JWKS.

### Code Samples

```csharp

```

zachwolfe:
Fixed by #997

PR #997, opened 2025-10-01
(auth0): support direct JWT verification

ViktorWb:
This PR changes how the Auth0 provider handles JWTs, and how it gets the user info. Instead of using the Auth0 Authentication API, the JWT is verified using the Auth0 JWKS, and user info (id, email, name, etc.) is read straight from the JWT claims.

Issue #988

### DIFF
--- a/Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
+++ b/Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
@@ -164,8 +164,8 @@ To allow the Ivy application to properly display the users email, name and avata
 4. **Add the following code:**
 ```javascript
 exports.onExecutePostLogin = async (event, api) => {
-    api.accessToken.setCustomClaim('http://schemas.xmlsoap.org/ws/2005/05/identity/claims/email', event.user.email);
-    api.accessToken.setCustomClaim('http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name', event.user.name);
+    api.accessToken.setCustomClaim('email', event.user.email);
+    api.accessToken.setCustomClaim('name', event.user.name);
     api.accessToken.setCustomClaim('avatar', event.user.picture);
 };
 ```
--- a/Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
+++ b/Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
@@ -154,6 +154,28 @@ For more information on setting up Apple authentication, see Auth0's [Apple docu
 
 For more information on setting up Twitter/X authentication, see Auth0's [Twitter documentation](https://marketplace.auth0.com/integrations/twitter-social-connection).
 
+### Step 6: Add claims to JWT
+
+To allow the Ivy application to properly display the users email, name and avatar, these need to be added to the JWT. In Auth0, this is done using an action.
+
+1. **Go to Actions > Library** in the Auth0 Dashboard
+2. **Click "Create Action" > "Create Custom Action"**
+3. **Give it a name like "Add user claims to JWT" and choose "Login / Post Login" as the trigger.**
+4. **Add the following code:**
+```javascript
+exports.onExecutePostLogin = async (event, api) => {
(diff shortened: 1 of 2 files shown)


zachwolfe:
Thanks for reporting and fixing this. I will test this out today.

Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
```diff
@@ -154,6 +154,28 @@ For more information on setting up Apple authentication, see Auth0's [Apple docu
 
 For more information on setting up Twitter/X authentication, see Auth0's [Twitter documentation](https://marketplace.auth0.com/integrations/twitter-social-connection).
 
+### Step 6: Add claims to JWT
```
zachwolfe, line 157: It is unfortunate to have to add this configuration step, but I agree that it is probably worth it to avoid rate limits. In the long run hopefully we can start caching user info, which would enable us to call the API to get user info again.

ViktorWb:
@zachwolfe Thank you for your feedback, I think the JWT claims are now simplified and also the Google provider is working for me. Could you try it again?

Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Auth0.md
ViktorWb: I agree!

If the authentication API is rate limiting based on IP-address then caching might not help if there are many unique users (each unique user will miss the cache and require an authentication API request). An alternative worth considering is the management API, but that would also require additional configuration to setup.

zachwolfe: Good point.

zachwolfe:
@ViktorWb Hmm, now I'm not seeing any user info at all (with your latest commit and updated action code). I will do some digging.

zachwolfe:
@ViktorWb Here is a decoded JWT payload I get when I use this branch, before your latest commit (slightly redacted):
```
{
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/email": "myemailaddress@gmail.com",
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name": "Zach Wolfe",
  "iss": "https://<tenant-id>.us.auth0.com/",
  "sub": "google-oauth2|my-numeric-user-id",
  "aud": [
    "https://<tenant-id>.us.auth0.com/api/v2/",
    "https://<tenanit-id>.us.auth0.com/userinfo"
  ],
  "iat": 1759358127,
  "exp": 1759444527,
  "scope": "openid profile email",
  "azp": "alphanumeric-client-id"
}
```

And, here it is after (looks the same on the `main` branch as well):
```
{
  "iss": "https://<tenant-id>.us.auth0.com/",
  "sub": "google-oauth2|my-numeric-user-id",
  "aud": [
    "https://<tenant-id>.us.auth0.com/api/v2/",
    "https://<tenant-id>.us.auth0.com/userinfo"
  ],
  "iat": 1759359302,
  "exp": 1759445702,
  "scope": "openid profile email",
  "azp": "alphanumeric-client-id"
}
```

So it seems like the custom claim keys from your latest commit are not being included in returned JWTs.

ViktorWb:
@zachwolfe I don't know why it differs for us - I can see the claims just fine. And according to [this](https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims) I think the email, name and avatar fields should be possible to write?

However, in general Auth0 recommends namespacing claims. Maybe this will help, because the namespaced claims (http://schemas.xmlsoap.org/ws/2005/05/identity/claims/) did show up for you? My latest commits allows the user to choose any namespace and use the  `Auth0:Namespace` user secret to let Ivy know from which namespace to read from.

If we choose this approach we will need to implement a step in the Ivy CLI where the user can add their namespace, and the CLI adds the `Auth0:Namespace` user secret.

zachwolfe:
@ViktorWb I also don't know why we're seeing different things, but I can confirm that your latest commit now works for me, using the https://your-domain.com/ placeholder as my namespace. With that said, I think having the namespace be configurable is great to have as an advanced option, but I would like to see a default that most users can use without having to think about it. Could we use something like https://ivy.app/claims/, for instance?

ViktorWb:
@zachwolfe My latest commit makes the namespace default to https://ivy.app/. Please have a look!

zachwolfe (approved):
LGTM! I will pass this to @nielsbosma for final review.

Issue #1015, opened 2025-10-03
Improve search of DefaultSideBarChrome


ShogunFire:
### Description

Improve the search functionality of DefaultSideBarChrome by enabling user to search in app tags or description instead of just app name

### Use Case

Right now the search present in the default side bar only search app names that starts with what is in the search. input. It makes it hard to find content for example in the doc or in the samples. This feature would make the search functionality more usefull.

### Proposed Solution

It would be nice to be able to search in tags or description. For example above each app you would have:

[App(title: "TextInput", tags:["password","textarea",  "search", "email"]

and the user could type "mail" and the TextInputApp would show. 

### Alternatives Considered

Not sure this is possible but maybe by looking in each app content directly

### Impact

Easier to find content in docs, samples for developpers. They can also use that feature in their own app

### Code Examples

```csharp

```

ShogunFire:
Putting all the tags will take a lot of time though, I wonder if there is a better way

PR #1016, opened 2025-10-03
(search) Improve search in the sidebar

ShogunFire:
Feature [Issue 1015](https://github.com/Ivy-Interactive/Ivy-Framework/issues/1015)

 Add the possibility to add tags to apps to search them in the sidebar more easily, and add an example for textInput.

I also made a change so that we can search app name in the middle instead of only in the start.

### DIFF
--- a/Ivy/Apps/AppAttribute.cs
+++ b/Ivy/Apps/AppAttribute.cs
@@ -13,7 +13,7 @@ public class AppAttribute(
     int order = 0,
     bool groupExpanded = false,
     string? documentSource = null,
-    string[]? searchTags = null
+    string[]? searchHints = null
 )
     : System.Attribute
 {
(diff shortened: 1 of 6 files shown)


rorychatt (approved):
tested, works.

I think this is a good concept!

@nielsbosma have a look at the API changes, I will after we merge this make sure all other pages have more tags too

Ivy/Apps/AppAttribute.cs
```diff
@@ -12,7 +12,8 @@ public class AppAttribute(
     bool isVisible = true,
     int order = 0,
     bool groupExpanded = false,
-    string? documentSource = null
+    string? documentSource = null,
+    string[]? searchTags = null
```
nielsbosma: I'd prefer to use the term searchHints. "Tags" is something we use for something else (attaching contextual data to Buttons etc.).

nielsbosma (changes requested):
Don't forget to add to docs.

ShogunFire:
Sorry for the delay, I made the changes

rorychatt:
We need to implement the new SearchHints for all pages in docs and samples!

PR #1018, opened 2025-10-03
[PieChart]: Redefine fill color for label list

ArtemKhvorostianyi:
Old one:
Screenshot 2025-10-04 at 01 08 55

New one:
Screenshot 2025-10-04 at 01 10 22

In light theme:
Screenshot 2025-10-04 at 01 10 37


ArtemKhvorostianyi:
@claude review this

rorychatt:
@ArtemKhvorostianyi 

> Loss of Customization: The previous implementation allowed custom fill colors via the fill prop (fill && 'var(--${fill.toLowerCase()})'). Now all label lists will use the same muted color regardless of the fill prop value. Is this intentional?

-> verify this. Maybe we just set the color to an ugly one from the builder method .toChart

PR #1019, opened 2025-10-03
Feature: Prevent tabs from overflowing on mobile

ViktorWb:
This PR improves tabs (variant "Content") on mobile. The tabs widget will now respect the `width` prop, which by default is `Size.Full()`. There are two modes when width is restricted: Scroll in x-direction or wrapping.

By default scroll in x-direction is applied. Use `Layout.Tabs().Wrap(true)` to instead wrap.

To prevent both scroll and wrap and revert to the old behaviour use `Layout.Tabs().Width(Size.MaxContent())`.

Wrap visuals:

Screenshot from 2025-10-03 23-26-38
Screenshot from 2025-10-03 23-26-49
Screenshot from 2025-10-03 23-26-54

An alternative way to solve overflow on mobile would be hamburger menus. Since one can opt out of scroll/wrap with Size.MaxContent() I don't think they exclude one another. Have a look and see what you think of this solution!

### DIFF
--- a/Ivy/Views/TabsView.cs
+++ b/Ivy/Views/TabsView.cs
@@ -15,6 +15,7 @@ public class TabView : ViewBase
     private TabsVariant _variant = TabsVariant.Content;
     private bool _removeParentPadding = false;
     private Thickness? _padding = new Thickness(4);
+    private bool _wrap = false;
 
     /// <summary>
     /// Internal constructor that initializes a TabView with predefined tabs.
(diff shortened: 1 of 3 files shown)


rorychatt (approved):

rorychatt:
@ViktorWb resolve error in pipeline

rorychatt:
@ViktorWb looks good to me. Could you also update the docs?

nielsbosma (changes requested):
We should solve this using overflow menus - like we have with the other tabs type.

rorychatt:
> We should solve this using overflow menus - like we have with the other tabs type.

I did not first understand the reference, but the idea is that with regular tabs currently you get a dropdown menu with the list of tabs that are now "hidden" -> tested and confirmed that at least on Desktop for regular tabs it works fine

ViktorWb:
I have now edited this PR to instead add the dropdown-menu for the Content variant as well:

image

This simplifies and is a nicer solution in my opinion. I have removed the scroll and wrap behaviors, instead all tab variants show a dropdown.

PR #1023, opened 2025-10-04
[Dashboard]: avoid using Text.H4 to avoid leaking headings into TOC

ArtemKhvorostianyi:
I tried to make it look as familiar as possible, the way it was before
How it was:

Screenshot 2025-10-04 at 07 21 21

How it now:
Screenshot 2025-10-04 at 07 21 30

Changed, because: 
Screenshot 2025-10-04 at 07 22 07

### DIFF
--- a/Ivy/Views/Dashboards/MetricView.cs
+++ b/Ivy/Views/Dashboards/MetricView.cs
@@ -69,13 +69,13 @@ public class MetricView(
         var x = data.Value;
 
         return new Card(
-                Layout.Vertical().Gap(2)
+                Layout.Vertical().Gap(6)
                 | (Layout.Horizontal().Gap(2)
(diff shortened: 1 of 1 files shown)


rorychatt:
don't avoid, just modify selectors. 

Make it if headers are from an element that has aria role tab (demo-tabs) => they are not included

Ivy/Views/Dashboards/MetricView.cs
```diff
@@ -69,13 +69,13 @@ Func<Task<MetricRecord>> metricData
         var x = data.Value;
 
         return new Card(
-                Layout.Vertical().Gap(2)
+                Layout.Vertical().Gap(6)
```
rorychatt, line 72: -

Issue #831, opened 2025-09-15
(epic) Navigation Beacons


nielsbosma:
We want the ability for Ivy apps to navigate to another Ivy app that can handle a certain type without it being aware that that application exists. This is useful for when creating Links in apps where we don't know if that app already exists. If the app exists then the link can be shown. 

Applications can define "NavigationBeacons"

```
[App]
[NavigationBeacon(typeof(Entity), nameof(GetNavigationBeacon))]
public class FooApp : ViewBase
{
   public static NavigationBeacon<Entity> GetNavigationBeacon()
    {
        //Defines:
        //Entity
        //Primary key
        //Args builder PK -> Args
    } 
}
```

The Navigation API need to support:
- Is there a beacon for entity type X?
- Navigate to beacon

All type safe.

ShogunFire:
@nielsbosma I am interested in trying to do this. One thing I don't understand is how the caller will call Navigate method, is it by passing just the primaryKey or the entire entity ? 
For example does he do:

`NavigateBeacon<User>(user)`

or 

`NavigateBeacon<User>(idUser)`


If he only pass the primaryKey idUser, how does  he know what the primaryKey is ?


Also, what happens if we have multiple app declaring NavigationBeacon for the same entity ?

PR #1026, opened 2025-10-04
(navigation) Add navigation Beacon feature to navigate to an app that handle a certain type without knowing if this app exists 

ShogunFire:
Feature [issue 831](https://github.com/Ivy-Interactive/Ivy-Framework/issues/831)

Add the possibility to have navigation beacon to navigate to an app that handle a certain type without knowing if this app exists.

Add a very simple example (maybe too simple ?) to show how to use the navigation beacon attribute and how to call the new methods of INavigator:
`public void NavigateToBeacon<T>(T Entity)`

and 
`public bool HasNavigationBeaconFor<T>()`


rorychatt:
@ShogunFire this will be on hold for a bit, just so you know

ShogunFire:
@rorychatt No problem, this give me time to fix the bacon :smile:

rorychatt:
this is getting outdated, and we are not sure if this is what we needed.
So I am closing this for now so it does not appear in the list of open PR-s

PR #1072, opened 2025-10-08
[Docs]: fix leaking headings from examples to the TOC

ArtemKhvorostianyi:
First of all, issue was opened about leaked number in card documentation, metric view example, because it renders as Text.H4
Screenshot 2025-10-09 at 00 46 00

Fixed by adding filtering Text.H* which presented in examples (csharp demo-tabs, demo-below etc)

Screenshot 2025-10-09 at 00 47 52

We have a lot of these problems in docs. For example, in HTML docs from release:

Screenshot 2025-10-09 at 00 50 00

And this page with fix:
Screenshot 2025-10-09 at 00 51 29

### DIFF
--- a/frontend/src/widgets/internal/DemoBoxWidget.tsx
+++ b/frontend/src/widgets/internal/DemoBoxWidget.tsx
@@ -56,7 +56,7 @@ const DemoBoxWidget: React.FC<DemoBoxWidgetProps> = ({
   };
 
   return (
-    <div className={cn('border')} style={styles}>
+    <div className={cn('border')} style={styles} data-demo-box>
       {children}
     </div>
   );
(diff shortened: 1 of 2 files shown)


ArtemKhvorostianyi:
@claude review this

ArtemKhvorostianyi:
"Specificity: div.border might be too broad if other bordered elements exist that aren't example boxes. Consider if a more specific selector like div.border[class*="example"] or div.border[class*="demo"] would be safer." -> Cloude

Added a tag for demo boxes to identify it

Screenshot 2025-10-09 at 02 23 35

Improved how it’s found in TableOfContentsWidget.tsx

Screenshot 2025-10-09 at 02 24 24

frontend/src/widgets/internal/DemoBoxWidget.tsx
```diff
@@ -56,7 +56,7 @@ const DemoBoxWidget: React.FC<DemoBoxWidgetProps> = ({
   };
 
   return (
-    <div className={cn('border')} style={styles}>
+    <div className={cn('border')} style={styles} data-demo-box>
```
rorychatt, line 59: aria-role works too... but ok

PR #1073, opened 2025-10-08
[Code Input]: fix selection of all same symbols

ArtemKhvorostianyi:
We had an issue with the code input widgets — selecting one symbol caused all identical symbols in the widget to be selected.

Screenshot 2025-10-09 at 01 58 24

This was fixed by improving the CodeInputWidget.tsx component.
I removed the overly complicated code related to selection and replaced it with a much simpler solution. The root cause of the issue was the strange selection logic in CodeMirror (the library we use for code highlighting).

Screenshot 2025-10-09 at 02 01 21

Here’s how the selection looks now:

Screenshot 2025-10-09 at 02 01 54

And for the light theme:

Screenshot 2025-10-09 at 02 04 41


ArtemKhvorostianyi:
@claude review this

PR #1098, opened 2025-10-14
(docs): Improve auth provider docs, make Auth0:Audience parameter required

zachwolfe:
### Overview
- Made it more clear that `BasicAuth:HashSecret` and `BasicAuth:JwtSecret` are both **required** at runtime, even though they are optional during setup with `ivy auth add`.
- Added explanation of the difference between `BasicAuth:Users` when passed to `ivy auth add`  vs. at runtime (passwords are plaintext in the former case, but hashed in the latter case)
- Made `Auth0:Audience` required, because it is documented as such.
- Added `BasicAuth:` prefix to basic auth parameters in the docs, to be consistent with the docs for other auth providers
- Fixed leftover old-style auth parameters like `AUTH0_DOMAIN` in the auth overview page


zachwolfe:
@claude review please

PR #1123, opened 2025-10-16
[CopyTextButton]: Improve appearance of the copy text button

ArtemKhvorostianyi:
Had an issue:
Screenshot 2025-10-16 at 04 48 32

Solved.
Code blocks now display a copy button when hovered — the button appears above the code block.
Not hovered:
Screenshot 2025-10-16 at 04 52 17

On hover:
Screenshot 2025-10-16 at 04 51 22

### DIFF
--- a/frontend/src/widgets/primitives/CodeWidget.tsx
+++ b/frontend/src/widgets/primitives/CodeWidget.tsx
@@ -73,7 +73,7 @@ const CodeWidget: React.FC<CodeWidgetProps> = memo(
       }
 
       return baseStyles;
-    }, [width, height, showBorder]);
+    }, [width, height, showBorder, showCopyButton]);
 
     const highlighterKey = useMemo(
       () =>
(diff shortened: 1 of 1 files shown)


rorychatt:
now it can block the text above it :D

rorychatt:
image

@ArtemKhvorostianyi take this for inspiration.
@dcrjodle what do you think?

ArtemKhvorostianyi:
@rorychatt , done
Screenshot 2025-10-18 at 01 48 26

rorychatt:
@nielsbosma my guess is that if we just add a bg, then we just "hide" the issue, while it may be more visually pleasing, there will be issues with selecting the text that's right behind the icon.

Maybe with the @ArtemKhvorostianyi attempt, if we made the heading narrower...

-> I think it's totally fine if we also do the hidden bg. Not so common to need to select just that part that is getting hidden.
We can also make the copy button disappear unless the user is hovered on the code widget?

ArtemKhvorostianyi:
Screenshot 2025-10-21 at 01 31 02

1) Button is always visible
2) Button overlays the text behind it
3) It doesn't has opacity
4) Added some padding from the right, so now all the text doesn't hide behind the copy button

Screenshot 2025-10-21 at 01 35 30

It's very similar to what we had before these changes, minimum of changes
I'd like to hear @rorychatt opinion about adding padding from the right, but this change is logical as for me

rorychatt:
Let's merge this and come back to it later if someone complains.

PR #1130, opened 2025-10-17
(kanban): create Kanban widget

rorychatt:
https://github.com/user-attachments/assets/4c667738-99b4-4c53-a1a4-b5a4e86c4ead

### DIFF
--- a/Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
+++ b/Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
@@ -22,15 +22,21 @@ public class KanbanApp : SampleBase
         {
             new Task { Id = "1", Title = "Design Homepage", Status = "Todo", Priority = 2, Description = "Create wireframes and mockups", Assignee = "Alice" },
             new Task { Id = "2", Title = "Setup Database", Status = "Todo", Priority = 1, Description = "Configure PostgreSQL instance", Assignee = "Bob" },
-            new Task { Id = "3", Title = "Implement Auth", Status = "In Progress", Priority = 1, Description = "Add OAuth2 authentication", Assignee = "Charlie" },
-            new Task { Id = "4", Title = "Build API", Status = "In Progress", Priority = 2, Description = "Create REST endpoints", Assignee = "Alice" },
-            new Task { Id = "5", Title = "Unit Tests", Status = "Done", Priority = 2, Description = "Write comprehensive test suite", Assignee = "Bob" },
-            new Task { Id = "6", Title = "Deploy to Production", Status = "Done", Priority = 1, Description = "Configure CI/CD pipeline", Assignee = "Charlie" },
+            new Task { Id = "3", Title = "Implement Auth", Status = "Todo", Priority = 3, Description = "Add OAuth2 authentication", Assignee = "Charlie" },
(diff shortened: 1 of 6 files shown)


frontend/src/widgets/widgetMap.ts
```diff
@@ -8,6 +8,7 @@ import * as Blades from '@/widgets/blades';
 import * as Lists from '@/widgets/lists';
 import * as Details from '@/widgets/details';
 import * as Primitives from '@/widgets/primitives';
+import * as Kanban from '@/widgets/kanban';
```
dcrjodle: Let's decide for uniformity what's the best import approach. `{item1, item2, item3} from 'module'` or `* as module from 'module'`

rorychatt: Closing

As discussed, we will refactor all into the first style

frontend/src/widgets/index.ts
```diff
@@ -16,3 +16,4 @@ export { SlotWidget } from './SlotWidget';
 export { TooltipWidget } from './TooltipWidget';
 
 export { PaginationWidget } from './PaginationWidget';
+export { KanbanWidget, KanbanColumnWidget, KanbanCardWidget } from './kanban';
```
nielsbosma, line 19: Maybe we should make Kanban a Suspense widget?

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        return Layout.Vertical(
+            Text.H3("Task Board Demo"),
+            Text.P("Showcasing kanban features: field selectors, column/card ordering, custom column titles, and event handlers."),
+
+            // Kanban with common features
+            tasks
+                .ToKanban(
+                    groupBySelector: e => e.Status,
+                    cardIdSelector: e => e.Id,
```
nielsbosma: rename to idSelector, titleSelector, descriptionSelector,

Ivy/Core/AbstractWidget.cs
```diff
@@ -164,44 +164,67 @@ public bool InvokeEvent(string eventName, JsonArray args)
         if (eventDelegate == null)
             return false;
 
-        if (IsFunc(eventDelegate, out Type? eventType, out Type? returnType))
+        if (IsFunc(eventDelegate, out Type? eventType, out Type? returnType) && returnType == typeof(ValueTask))
```
nielsbosma, line 167: Looks good, but this is starting to become quite a complicated mess. Think we need to refactor this and move most of this into Utils functions and create unit tests to makes sure we handle all corner cases.

nielsbosma (changes requested):
How are these colors defined?

Image

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -0,0 +1,62 @@
...
+                    cardTitleSelector: e => e.Title,
+                    cardDescriptionSelector: e => e.Description)
+                .ColumnOrder(e => e.StatusOrder)
+                .CardOrder(e => e.Priority)
```
rorychatt: use order selector

frontend/src/widgets/index.ts
rorychatt: Lazy load it in widget map -> will try out a heavy example and see if suspense is needed

PR #1135, opened 2025-10-17
(Way of Working) Enable Commit Rules

dcrjodle:


frontend/commitlint.config.js
rorychatt, line 1: there is another file with super similar contents to this

dcrjodle:
@rorychatt Fixed it and made the (scope) tag optional so messages like `(datatables) fix: add solution` and `fix: add solution` are valid

rorychatt:
@dcrjodle I really want to merge this, but I also really don't want to enforce strict rules...

I think so long as we commit in a similar way, it is okay. It is just that in IDE-s it is not always straight forward for the end user why the commit did not pass. I think we should make it easy to commit... and we have the guidelines.

Do you still think that we need this change?

rorychatt:
Closing this as discussed.

We might shoot ourselves in the leg if we do this

PR #1137, opened 2025-10-17
[Forms]:  Implement Size for Forms

ArtemKhvorostianyi:
Screenshot 2025-10-18 at 02 19 39


ArtemKhvorostianyi:
@claude  review this

rorychatt (changes requested):
Something is off with the checkbox.
ALso, we should increase the gap between elements. When the form is bigger, the spacing should also increase so the forms are more spatious.

See reference screenshot in original issue

ArtemKhvorostianyi:
Added different gaps, fixed checkboxes

Screenshot 2025-10-22 at 11 31 54

Text at the top has the same size, because it's not actually a form part, it's a card's title, which contains form. Same thing for buttons, it isn't accessable to change its size (?).
Screenshot 2025-10-22 at 11 33 55

rorychatt:
@ArtemKhvorostianyi aha, I see. But then the Cards need support for Size.

But then I have this feedback:

image

In forms space below a description needs to be smaller than above the description. So the description should be closer to the field it relates to, not closer to the one above it.

ArtemKhvorostianyi:
Screenshot 2025-10-22 at 13 40 31

Sure. It's done, gaps are fixed

frontend/src/widgets/forms/FormWidget.tsx
```diff
@@ -1,16 +1,34 @@
...
+      case Sizes.Large:
+        return 'gap-4 text-base [&_input:not([type="checkbox"])]:text-base [&_input:not([type="checkbox"])]:h-12 [&_input:not([type="checkbox"])]:px-4 [&_input:not([type="checkbox"])]:py-3 [&_textarea]:text-base [&_textarea]:min-h-[120px] [&_select]:text-base [&_select]:h-12 [&_label]:text-base [&_p]:text-base [&_button:not([role="checkbox"])]:text-base [&_button:not([role="checkbox"])]:h-12 [&_button:not([role="checkbox"])]:px-6 [&_.field]:gap-3 [&_.field]:mt-3 [&_.field]:mb-1.5 [&_button[role="checkbox"]]:!h-5 [&_button[role="checkbox"]]:!w-5';
+      default:
+        return 'gap-3 text-sm [&_input:not([type="checkbox"])]:text-sm [&_input:not([type="checkbox"])]:h-10 [&_input:not([type="checkbox"])]:px-3 [&_input:not([type="checkbox"])]:py-2 [&_textarea]:text-sm [&_textarea]:min-h-[80px] [&_select]:text-sm [&_select]:h-10 [&_label]:text-sm [&_p]:text-sm [&_button:not([role="checkbox"])]:text-sm [&_button:not([role="checkbox"])]:h-10 [&_button:not([role="checkbox"])]:px-4 [&_.field]:gap-2 [&_.field]:mt-1 [&_.field]:mb-2 [&_button[role="checkbox"]]:!h-4 [&_button[role="checkbox"]]:!w-4';
```
rorychatt, line 28: what is this css for?

rorychatt: the paddings/gaps are BE defined... Maybe solution is to in the FormBuilder, when setting the Size, to override the paddings

PR #1139, opened 2025-10-18
[Tabs]: implement "Add Tab" button

ArtemKhvorostianyi:
The issue was that in TabsApp we had weird placement of "add tab" button, which was presented with floating panel.
Screenshot 2025-10-18 at 04 29 01

And task was to replace it in somewhere better place, but I thoughn that implementation of adding tab button via helper method would be better solving of this problem
Screenshot 2025-10-18 at 04 30 56

TabsVariant.Tabs have a bit more difficult design with on hover support.
On hover demo:
Screenshot 2025-10-18 at 04 32 21


ArtemKhvorostianyi:
@claude review this

rorychatt (approved):

PR #1161, opened 2025-10-21
(charts): resolve jank

svetlana-soboleva:
Prevents layout shifts when switching between tabs or loading new data.


frontend/src/index.css
rorychatt, line 1: good catch! I forgot to cleanup recharts css.... will create an issue about this

rorychatt (approved):

PR #1172, opened 2025-10-22
[Text Ellipsis]: improve ellipsis logic for text

ArtemKhvorostianyi:
We had some problems with text ellipsis
Screenshot 2025-10-22 at 12 06 24
But the issue was that fixing of this problem was causing other problem -> not all the text that should be cropped was actually cropped. Avoided these problems.
Screenshot 2025-10-22 at 12 09 25
Screenshot 2025-10-22 at 12 23 02


rorychatt:
resolve conflicts.
And as discussed during the call, let's use the FieldWidget more for examples like these!

PR #1200, opened 2025-10-23
[Text]: Improve ellipsis for long text in elements

ArtemKhvorostianyi:
Inputs: 
Screenshot 2025-10-24 at 02 00 48
Screenshot 2025-10-24 at 02 01 12
Screenshot 2025-10-24 at 02 02 58

Work still in progress
We have pretty many elements that don't have proper handling for extremely long values


rorychatt:
ASync select in the example above should have gotten a overflow ellipsis on the text

PR #1206, opened 2025-10-26
[Docs]: implement Kanban docs

ArtemKhvorostianyi:
Docs -> Widgets -> Advanced -> Kanban

We also have handleAdd, handleRemove methods in the BackEnd, but I found out that there is no anything about it in the Frontend. So, when this functionality will be done, we should update this doc with some more examples

Screenshot 2025-10-26 at 15 34 42
Screenshot 2025-10-26 at 15 35 22
Screenshot 2025-10-26 at 15 35 36
Screenshot 2025-10-26 at 15 35 53

we also have handleAdd, handleRemove methods in the BackEnd, but I found out that there is no anything about it in the Frontend. So, when this functionality will be done, we should update this doc with some more examples


ArtemKhvorostianyi:
@claude review this

PR #1208, opened 2025-10-27
Feat/charts polishing2

rorychatt:
Fixes charts getting bad heights in some examples

image


rorychatt:
Probably could be smarted, so there is less padding. I am hotfixing this right now and will create a separate issue for this!

PR #1210, opened 2025-10-27
Feat/configurable toolbox for charts

svetlana-soboleva:
Line & Area Charts: Configure toolbox with Data View, Magic Type, Save image.
Line & Area Charts Padding: Fixed X-axis padding so charts start exactly at the first tick, removing unnecessary space before the first data point.
Vertical Bar Charts: Disabled the toolbox temporarily because the layout appears broken with vertical orientation.
Pie & Area Charts: Toolbox now only allows “Save as Image” and "Data View" to avoid layout issues; other features (Restore, Magic Type) are disabled.


Ivy/Widgets/Charts/Shared/Toolbox.cs
```diff
@@ -0,0 +1,267 @@
+// ReSharper disable once CheckNamespace
+namespace Ivy.Charts;
+
+/// <summary>
+/// Represents a toolbox configuration for charts.
+/// </summary>
+public record Toolbox
+{
+    /// <summary>
```
rorychatt, line 9: you can have /// summary and closing tag on the same line = less vertical space usage

Ivy/Widgets/Charts/BarChart.cs
```diff
@@ -148,13 +153,17 @@ public static BarChart Horizontal(this BarChart chart)
     }
 
     /// <summary>
-    /// Sets the layout to vertical.
+    /// Sets the layout to vertical and disables the toolbox to avoid visual conflicts.
```
rorychatt, line 156: if we don't want a toolbox, just don't set it?

i don't think we need Enabled(false) -> if we don't set the object, it is not sent to the frontend and thus it should not be added

Ivy/Widgets/Charts/BarChart.cs
```diff
@@ -58,6 +58,11 @@ public BarChart(object data, params Bar[] bars)
     /// </summary>
     [Prop] public Legend? Legend { get; init; } = null;
 
+    /// <summary>
+    /// Gets or sets the toolbox configuration.
+    /// </summary>
+    [Prop] public Toolbox? Toolbox { get; init; } = new Toolbox();
```
rorychatt: might be weird to set a default value to an object instance when it is nullable

svetlana-soboleva:
I have discovered a render glitch in charts when switching tabs or showing previously hidden containers.
The issue appeared after the introduction of chartStyles in the chart wrapper. We had this issue before. These changes cause ECharts to miscalculate container dimensions when the chart is initially hidden.  As a result, the chart renders in a minified or squashed state for a brief moment when the container becomes visible.

rorychatt:
@svetlana-soboleva we have several ongoing PR-s around charts. We are doing a css refactor and fixing this mentioned issue in another PR

rorychatt:
@svetlana-soboleva left some feedback for you on LinkedIn

rorychatt (dismissed):

rorychatt:
@svetlana-soboleva we just merged a big refactor with styles for charts

nielsbosma (dismissed):

PR #1258, opened 2025-10-30
[Field]: In the Field docs use WithField() as recommended

ArtemKhvorostianyi:
Screenshot 2025-10-30 at 02 02 42

Same for other examples, except basic usage
Also refactored "When to use" section, which contained tip into callout

Was: 
Screenshot 2025-10-30 at 02 04 01

Now: 
Screenshot 2025-10-30 at 02 04 27


ArtemKhvorostianyi:
@claude review this

PR #1260, opened 2025-10-30
[Text]: Add text muted, bold, italic as properties

ArtemKhvorostianyi:
All text methods that return TextBuilder now support three modifier methods:

.Bold() 
.Italic()
.Muted()

Updated Text app page in Samples
Screenshot 2025-10-30 at 03 12 05

How it looked before (I think it was a good idea to update text app, because it wasn't representative):
Screenshot 2025-10-30 at 03 12 52

Updated Docs, just added a few examples in existing sections

Screenshot 2025-10-30 at 03 10 12

Screenshot 2025-10-30 at 03 10 59

_Specialized methods like Text.Code(), Text.Markdown(), Text.Json(), Text.Xml(), and Text.Html() create specialized widgets and don't support modifiers._

### DIFF
--- a/Ivy/Views/Text.cs
+++ b/Ivy/Views/Text.cs
@@ -590,10 +590,8 @@ public class TextBuilder(string content, TextVariant variant, Languages codeLang
         return this;
     }
 
-    /// <summary>
-    /// Sets whether the text should be displayed in bold.
-    /// </summary>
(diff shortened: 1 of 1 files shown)


ArtemKhvorostianyi:
@claude review this

PR #1275, opened 2025-10-30
fix(datatables): load batches automatically when container height exceeds initial batch size

### LINKED ISSUE #1235: (datatables): not loading new rows if height too big
### Description

If height is greater than the height of batch size received on init render, new batches don't load

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

dcrjodle:
## Summary
Fixes #1235 - DataTables now automatically load new batches when the container height is greater than the initial batch size.

## Problem
Previously, when a DataTable's container height was larger than the number of rows in the initial batch, new batches would not load automatically. This occurred because the `onVisibleRegionChanged` callback from glide-data-grid wouldn't fire immediately on initial render, leaving visible gaps in tall containers.

## Solution
Added a `useEffect` hook that monitors the visible rows and container height. The effect:
- Calculates how many rows the container can display based on its height
- Automatically triggers `loadMoreData()` when the container can show more rows than are currently loaded
- Continues loading batches until either the container is filled or no more data is available

## Changes
- Modified `DataTableEditor.tsx`:
  - Added effect to check container capacity vs. visible rows
  - Effect runs after each batch loads to determine if more data is needed
  - Self-regulating: stops when container is filled or `hasMore` becomes false

## Test Plan
- [x] Build completed successfully
- [ ] Test with various container heights (small, medium, large)
- [ ] Verify with different batch sizes (20, 50, 100)
- [ ] Test with filters and sorting applied
- [ ] Confirm scroll behavior still works correctly

## Screenshots
_N/A - behavior fix_

### DIFF
--- a/frontend/src/widgets/dataTables/DataTableEditor.tsx
+++ b/frontend/src/widgets/dataTables/DataTableEditor.tsx
@@ -117,7 +117,9 @@ export const DataTableEditor: React.FC<TableEditorProps> = ({
   });
   const [showSearch, setShowSearch] = useState(false);
   const [hoverRow, setHoverRow] = useState<number | undefined>(undefined);
+
   const scrollThreshold = 10;
+  const rowHeight = 38;
 
   // Generate header icons map for all column icons
   const headerIcons = useMemo(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/widgets/dataTables/DataTableEditor.tsx
```diff
@@ -117,7 +117,9 @@ export const DataTableEditor: React.FC<TableEditorProps> = ({
   });
   const [showSearch, setShowSearch] = useState(false);
   const [hoverRow, setHoverRow] = useState<number | undefined>(undefined);
+
   const scrollThreshold = 10;
+  const rowHeight = 38;
```
rorychatt, line 122: I'm curious does this logic break if using not 100% scaling? 
What happens on 50% scaling for example?

dcrjodle: what's scaling. height configuration?

rorychatt: browser/OS scaling

dcrjodle:
This PR is a small change that improves the batch behaviour. But there is a some sporadic behaviour where the bug still occurs. Let's merge this PR and monitor if it becomes an issue

PR #1332, opened 2025-11-03
[CodeInput]: fix copy button background visibility in different elements

ArtemKhvorostianyi:
Had an issue:
Screenshot 2025-11-04 at 00 11 14

@rorychatt , before mergind, tell me if it's good, I think it will be better to remove these new examples form CodeInputApp.cs, it seems to be redundant there. Before approving of PR, I won't remove it for case if we want some more changes about it

Fixed: 

Screenshot 2025-11-04 at 00 11 59

Screenshot 2025-11-04 at 00 12 17

On hover:

Screenshot 2025-11-04 at 00 12 37

When clicked:

Screenshot 2025-11-04 at 00 12 53

With basic black background:

Screenshot 2025-11-04 at 00 13 43

On hover:

Screenshot 2025-11-04 at 00 13 59

When clicked:

Screenshot 2025-11-04 at 00 14 45

### DIFF
--- a/Ivy.Samples.Shared/Apps/Widgets/Inputs/CodeInputApp.cs
+++ b/Ivy.Samples.Shared/Apps/Widgets/Inputs/CodeInputApp.cs
@@ -159,6 +159,32 @@ public class CodeInputApp : SampleBase
 
         var dataBinding = CreateStringTypeTests();
 
+        var cardCode = UseState(
+            """
+            public class Example
+            {
+                public void DoWork()
+                {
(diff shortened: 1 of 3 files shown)


dcrjodle:
- I see this is a custom button (why are we not using the existing button component variants?)
- Why are we doing `px-4` and `py-3`? Should it not be perfectly squared?

dcrjodle:
My comment does not refer to the changes, but rather the file that the changes are in

PR #1344, opened 2025-11-04
(url): attempt solve disconnects issue; resolve infinite sidebar render when on port 5010

rorychatt:
This pull request addresses a race condition in the AppHost connection logic and improves frontend connection stability and logging. The most significant changes involve how parent session IDs are resolved on reconnect, and how frontend hooks manage dependencies and log connection events.

**Backend: AppHost Parent Session Resolution**

* Added logic in `AppHub.cs` to resolve invalid or stale `parentId` values by searching for a valid Chrome session for the same machine, falling back to standalone mode if none is found. This prevents issues when AppHost reconnects with an outdated parent reference.

**Frontend: Connection Stability and Logging**

* Updated the dependency array in the `useBackend` hook (`use-backend.tsx`) to include `chrome`, ensuring the effect runs correctly when Chrome state changes, and clarified the rationale for excluding `appId` in favor of `stableAppId`.
* Modified the WebSocket connection established log to use `stableAppId` (or `null`) instead of `appId`, improving consistency with backend expectations and preventing misleading log output.

### DIFF
--- a/frontend/src/hooks/use-backend.tsx
+++ b/frontend/src/hooks/use-backend.tsx
@@ -288,7 +288,7 @@ export const useBackend = (
 
     const newConnection = new signalR.HubConnectionBuilder()
       .withUrl(
-        `${getIvyHost()}/messages?appId=${stableAppId ?? ''}&appArgs=${appArgs ?? ''}&machineId=${machineId}&parentId=${parentId ?? ''}&chrome=${chrome}`
+        `${getIvyHost()}/messages?appId=${appId ?? ''}&appArgs=${appArgs ?? ''}&machineId=${machineId}&parentId=${parentId ?? ''}&chrome=${chrome}`
       )
       .withAutomaticReconnect()
       .build();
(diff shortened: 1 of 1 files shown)


rorychatt:
Tested local build 5010, fe build 5173, tested from build .dll files with a package inject from localhost
Tested with .dll with basic auth

zachwolfe:
@rorychatt I need to test the frontend change a bit more, but I'm sorry to say I don't think the server changes are a good idea:
- There could be multiple open chrome sessions with the same machine ID, in the case of more than one open browser tab or window, so selecting an arbitrary chrome parent like this could cause bugs in the future.
- If there are no other open browser tabs/windows, and the chrome parent we want to find hasn't reconnected yet, we won't find anything and the `chrome=false` redirect bug will still happen.
- It is possible for a valid parent id to not point to a chrome session, if the app developer uses AppHost directly. So in theory we could find a chrome session when we actually shouldn't.

rorychatt:
Alright. Lets close this one then

PR #1345, opened 2025-11-04
Fix DBML Editor Table Layout and Connection Handle Visibility

### LINKED ISSUE #1337: (dbml editor): improvements and minor bug fixes
### Description

Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

Jeelislive:
Removed extra connection dots on unlinked fields, added spacing between field name and type, and increased table box width for better field visibility and alignment.

Fixes #1337 

Screenshot from 2025-11-04 19-06-35

### DIFF
--- a/frontend/src/widgets/internal/DbmlCanvasWidget.tsx
+++ b/frontend/src/widgets/internal/DbmlCanvasWidget.tsx
@@ -280,7 +280,7 @@ const nodeTypes: NodeTypes = {
 
 const calculateNodeDimensions = (tableData: DbmlTableData) => {
   // Fixed dimensions for all tables
-  const fixedWidth = 280;
+  const fixedWidth = 240;
   const minHeight = 120;
   const fieldHeight = 28; // Height per field row
   const headerHeight = 40; // Height for table header


rorychatt:
what if the field is super long in naming?

Like `email_super_long_field_that_should_overflow_or_be_on_newline` and then have on it a super long type like `a_really_long_type_that_maybe_should_not_exist`

frontend/src/widgets/internal/DbmlCanvasWidget.tsx
```diff
@@ -487,7 +489,7 @@ export const DbmlCanvasWidget: React.FC<DbmlCanvasWidgetProps> = ({
           sourceHandle: `${ref.endpoints[0].fieldNames[0]}-right`,
           targetHandle: `${ref.endpoints[1].fieldNames[0]}-left`,
           type: 'smoothstep',
-          style: { stroke: getConnectionLineColor() },
+          style: { stroke: getConnectionLineColor(), strokeWidth: 2 },
```
rorychatt: not sure if making the lines bolder is a good idea. We could propose this in a separate PR as separate feature and provide screenshots with before and after, so ppls can choose

(to me personally lines look super thicc now)

@dcrjodle

Jeelislive:
Screenshot from 2025-11-04 19-35-02
Screenshot from 2025-11-04 19-29-19

frontend/src/widgets/internal/DbmlCanvasWidget.tsx
Jeelislive: @rorychatt i also think line shoulld be bold. as of now removed that in this pr.

rorychatt:
@Jeelislive for on hover, see conventions within the codebase. We actually have some styling for tooltips with tooltip provider

rorychatt:
image
IMO we should revert to original width

Jeelislive:
Orignal width with overflow with tooltip property right? @rorychatt

Jeelislive:
@rorychatt it's only increase size when we have long fields. Ping me if we dont need this width

rorychatt:
I looked at this again, changes look good, we just need to do the following:

- remove responsiveness from the cards that you added, this brings inconsistencies when a lot of cards are rendered at the same time
- the overflow solution for `var name` looks good, does it also work with `var type`?
- we have tooltips that we use in the framework, look up usage of `tooltip` -> this brings the common styling that we use and it should also be applied in a similar way here too

Jeelislive:
yes it will works for both type and value, with static width instead of dynamic

frontend/src/widgets/internal/DbmlCanvasWidget.css
```diff
@@ -78,6 +82,7 @@
 .react-flow__node {
   /* Disable transitions during drag to prevent slow-motion effect */
   transition: none !important;
+  z-index: 1 !important;
```
ArtemKhvorostianyi, line 85: Tested out this, isn't it redundant change? It seems nothing to do
It's okay if we had some issue with it, but I didn't see any changes

frontend/src/widgets/internal/DbmlCanvasWidget.tsx
```diff
@@ -248,35 +279,21 @@ const nodeTypes: NodeTypes = {
 };
 
 const calculateNodeDimensions = (tableData: DbmlTableData) => {
-  // Base dimensions
-  const minWidth = 200;
+  // Fixed dimensions for all tables
+  const fixedWidth = 280;
```
ArtemKhvorostianyi: I think we need to revert old width policy with minWidth. New one is too long. However, work with tooltips is fine as for me.

PR #1346, opened 2025-11-04
(Fix): Reduced padding for child columns in DataTable

### LINKED ISSUE #1281: (datatable): Column group headers not aligned properly
## Description
Column group headers in DataTable are not properly aligned with their child columns.

**ALSO** try to add borer beneath column groups

Image

## Expected Behavior
When using column groups (grouped headers), the group header should:
- Be properly centered or aligned over its child columns
- Span the correct width to cover all child columns
- Maintain proper visual hierarchy

## Actual Behavior
Column group headers are misaligned, possibly:
- Not centered over their child columns
- Incorrect width/spacing
- Poor visual alignment with the column structure below

## Impact
- Reduced table readability
- Confusing visual hierarchy
- Professional appearance issues

## Reproduction
1. Create a DataTable with column groups (nested headers)
2. Observe the group header alignment relative to child columns
3. Note misalignment issues

## Additional Context
This issue affects tables with hierarchical column headers and may be related to CSS flexbox/grid layout or colspan handling.

KaiserReich95:
closes #1281 

Before (lines to point out padding issue):
ivy_groupheaders_not_aligned

After:
ivy_groupheaders_aligned

### DIFF
--- a/frontend/src/widgets/dataTables/DataTableEditor.tsx
+++ b/frontend/src/widgets/dataTables/DataTableEditor.tsx
@@ -101,7 +101,7 @@ export const DataTableEditor: React.FC<TableEditorProps> = ({
       borderColor: showVerticalBorders
         ? colors.border || (isDark ? '#404045' : '#d1d5db')
         : 'transparent',
-      cellHorizontalPadding: 16,
+      cellHorizontalPadding: 8,
       cellVerticalPadding: 8,
       headerIconSize: 20,
       // Add proper text colors for group headers and icons


dcrjodle:
In the issue there was another note saying to also see if we can add bottom borders to the column groups. 
Sorry it was unclear

Screenshot 2025-11-05 at 09 22 00

KaiserReich95:
Asked Claude to solve the problem with border beneath the group headers but it seems to be complicated.

Question: "So the horizontal line beneath the group headers are tied to the vertical lines in the library? So its not possible to change that?"

**Exactly! Yes, that's correct.**  
In the `glide-data-grid` library, the horizontal line beneath group headers is tied to the vertical borders because they both use the same `theme.borderColor` property.

---

# The Library's Design
Looking at the source code at data-grid-render.header.ts line 284:
ctx.strokeStyle = theme.borderColor;  // Used for BOTH vertical AND horizontal group header borders
The library uses borderColor for:
✅ Vertical borders between columns (everywhere in the grid)
✅ Horizontal border beneath group headers
✅ Vertical borders between group header sections

# Is It Possible to Change?
Without modifying the library itself: No, it's not possible to have the group header bottom border without also showing vertical column borders. 

### To truly fix this, you would need to: 

Fork the @glideapps/glide-data-grid library. Modify line 284 in data-grid-render.header.ts from: ctx.strokeStyle = theme.borderColor;

to:

ctx.strokeStyle = theme.horizontalBorderColor ?? theme.borderColor;
Use your forked version

# Current Workaround

The solution we implemented (showVerticalBorders || showGroups) is the best we can do without modifying the library. It's a reasonable compromise:
When groups are shown → you get both the group header border AND vertical borders
When groups are hidden → you can hide vertical borders if desired
This is a design limitation of the glide-data-grid library, not something we can work around in our code without either forking the library or using complex CSS hacks.

dcrjodle:
In glide data grid's own example on their front page they have the border beneath the column groups. https://grid.glideapps.com/ Let's investigate this some more

dcrjodle:
Actually let me merge this, since I'm really bothered by the misalignment and then I'll create a new issue

PR #1350, opened 2025-11-04
(colors): remove obsolete -light/-dark variants and chart-x colors

defymecobra:
Deleted unused colors + chart-x

### DIFF
--- a/frontend/src/widgets/charts/styles/colors.ts
+++ b/frontend/src/widgets/charts/styles/colors.ts
@@ -13,13 +13,13 @@ export type ColorScheme = 'Default' | 'Rainbow';
 const defaultColorVars = [
   'emerald', // Green-cyan, high visibility
   'red', // Red, high contrast
-  'blue', // Blue, distinct
-  'amber', // Yellow-orange, warm
+  'teal', // Blue, distinct
   'purple', // Purple, good contrast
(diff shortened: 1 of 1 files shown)


defymecobra:
- Replace chart-x variables with defaultColorVars
- Use existing chromatic colors 
- Update color scheme naming for clarity

Charts now look like this:

image

image

image

image

defymecobra:
@rorychatt what do you think?

rorychatt:
good job!

I think area charts look weird with their colors due to all colors being overlapped...

Especially these 3
image


-> can we make it so default colors are in a better order?

rorychatt (changes requested):
good job so far, see my comment aboce

defymecobra:
image
image
image
image
image

@rorychatt what do you think?

To be honest, I'm not the most knowledgeable person when it comes to designing multiple color objects, and choosing the right color palette can be difficult at times. If this option isn't good, maybe someone who knows about color design could give me some advice?

rorychatt:
@dcrjodle do you have any suggestions for this here? I would like to merge this asap as this is cleanup, but we need to come up with better colors for charts. Also seemed to me it could match somehow with the new design system?

Let me know and let's create a new issue for any css var adjustment

Issue #1342, opened 2025-11-04
(sidebar): inconsistent bg for onhover


rorychatt:
### Description

<img width="318" height="1432" alt="Image" src="https://github.com/user-attachments/assets/9932109e-1ff9-4114-8bd2-423616bb3a33" />

<img width="296" height="298" alt="Image" src="https://github.com/user-attachments/assets/f173e104-6a99-4ec0-8adc-8a7031f73c1b" />

### Code Samples

```csharp

```

rorychatt:
@KaiserReich95

KaiserReich95:
I look into it

PR #1351, opened 2025-11-04
(sidebar): Fixed so that Settings button and the dropdown menu in it uses --sidebar-accent

KaiserReich95:
Issue #1342: Fixed where the Settings button in the sidebar footer was not using the --sidebar-accent color on hover, unlike other sidebar menu items.

# Changes Made

## Root Cause
The **Settings** button uses a `Button` component with `ButtonVariant.Ghost`, which applies  
`hover:bg-accent` instead of the `hover:bg-sidebar-accent` used by sidebar menu items.

### CSS Changes (`index.css`)
- Added a CSS rule to make **rounded buttons** (`Button` components) in the sidebar use  
  `--sidebar-accent` on hover.
- Added a CSS rule for **dropdown menu items** to use `--sidebar-accent` when opened from the sidebar.
- Used the `.rounded-md` class selector to target `Button` components while **excluding plain text buttons**  
  (e.g., `"Dismiss"` in the `"Star on GitHub"` `MenuItem`).

### DropDownMenuWidget Changes (`DropDownMenuWidget.tsx`)
- Added logic to **detect if a dropdown is inside the sidebar**.
- Applied the `dropdown-in-sidebar` class to portal content when the dropdown is opened from the sidebar.
- This was necessary because dropdown menus **render in a portal** outside the sidebar’s DOM tree.

## Example

Before:

https://github.com/user-attachments/assets/24239fc2-e685-4f1d-acf8-573a992d25b6

After:

https://github.com/user-attachments/assets/6af5563f-c031-460d-888c-5d9eca6da515

### DIFF
--- a/frontend/src/widgets/DropDownMenuWidget.tsx
+++ b/frontend/src/widgets/DropDownMenuWidget.tsx
@@ -42,17 +42,6 @@ export const DropDownMenuWidget: React.FC<DropDownMenuWidgetProps> = ({
   const eventHandler = useEventHandler();
   const [open, setOpen] = useState(false);
   const triggerRef = useRef<HTMLButtonElement>(null);
-  const [isInSidebar, setIsInSidebar] = useState(false);
-
-  // Check if this dropdown is inside a sidebar
-  React.useEffect(() => {
-    if (triggerRef.current) {
-      const sidebar = triggerRef.current.closest(
-        '.bg-sidebar, [data-sidebar="sidebar"]'
-      );
-      setIsInSidebar(!!sidebar);
-    }
-  }, []);
 
   if (!slots?.Trigger) {
     return (
@@ -148,9 +137,7 @@ export const DropDownMenuWidget: React.FC<DropDownMenuWidgetProps> = ({
               )}
             </DropdownMenuSubTrigger>
             <DropdownMenuPortal>
-              <DropdownMenuSubContent
-                className={`m-2 ${isInSidebar ? 'dropdown-in-sidebar' : ''}`}
-              >
+              <DropdownMenuSubContent className="m-2">
                 {renderMenuItems(item.children)}
               </DropdownMenuSubContent>
             </DropdownMenuPortal>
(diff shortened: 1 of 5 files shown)


frontend/src/widgets/DropDownMenuWidget.tsx
```diff
@@ -42,6 +42,17 @@ export const DropDownMenuWidget: React.FC<DropDownMenuWidgetProps> = ({
   const eventHandler = useEventHandler();
   const [open, setOpen] = useState(false);
   const triggerRef = useRef<HTMLButtonElement>(null);
+  const [isInSidebar, setIsInSidebar] = useState(false);
+
+  // Check if this dropdown is inside a sidebar
+  React.useEffect(() => {
+    if (triggerRef.current) {
+      const sidebar = triggerRef.current.closest(
+        '.bg-sidebar, [data-sidebar="sidebar"]'
```
rorychatt: I am not sure if this is good practice. While it works, it can tank the performance

rorychatt:
this is not an issue in dark mode. I think that solution is to figure out why 

a) two vars have the same color in dark mode = one of them is not needed?
b) why in light mode they are different = maybe only 1 var should remain

KaiserReich95:
It seems to be that in darkmode both the values --accent (which the Settings button and many other items use) and --sidebar-accent (which only sidebar items use) have the same color but in lightmode they have different colors. Maybe we should remove --sidebar-accent since it could just use --accent?

defymecobra:
> It seems to be that in darkmode both the values --accent (which the Settings button and many other items use) and --sidebar-accent (which only sidebar items use) have the same color but in lightmode they have different colors. Maybe we should remove --sidebar-accent since it could just use --accent?

Try removing -sidebar-accent in both themes, and wherever it was used, replace it with the -accent (that contains your new color for hover).

KaiserReich95:
Should we remove --sidebar-accent totally or keep it? And there are alot of other colors used for the sidebar that has the same values as the preceding values. For example --sidebar-primary and --primary.

Preceding colors:

ivy_colors_pre

Sidebar colors:

ivy_colors

rorychatt:
@KaiserReich95 if it makes no sense in the world of tailwindcss, (and it does not) then yes, we should not use it

KaiserReich95:
I removed some sidebar colors that have the exact same values as the default counterpart. We no longer have to check if items are specifically in the sidebar anymore.

Light mode:

--sidebar-primary: #00cc92; Replace with --primary:
--sidebar-accent: #eaebeb; Replace with --accent
--sidebar-accent-foreground: oklch(0.33 0 0); Replace with --accent-foreground
--sidebar-border: #d1d5db; Replace with --border
--sidebar-ring: oklch(0.77 0 0); Replace with --ring

Dark mode:

--sidebar-accent: #eaebeb; Replace with --accent
--sidebar-ring: oklch(0.77 0 0); Replace with --ring

rorychatt (approved):

PR #1353, opened 2025-11-04
Fix unwanted redirects to `chrome=false` on reconnection

### LINKED ISSUE #1308: (URL SP): `chrome=false` sometimes automatically set after inactivity
### Description

When I leave an Ivy app in a browser tab open and inactive for several minutes, `?chrome=false` is sometimes added to the URL and the connection is lost. It could have something to do with [this](https://github.com/Ivy-Interactive/Ivy-Framework/pull/963#discussion_r2471747618), but that's just a guess.

zachwolfe:
Fixes #1308

@rorychatt Thought of this fix while responding to your PR. I don't know if it also solves the infinite sidebar issue, because I don't know what causes that (but probably not?)

### DIFF
--- a/Ivy/AppHub.cs
+++ b/Ivy/AppHub.cs
@@ -196,11 +196,13 @@ public class AppHub(
 
             var widgetTree = new WidgetTree(app, contentBuilder, serviceProvider);
 
+            var parentId = GetParentId(httpContext);
+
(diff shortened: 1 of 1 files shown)


rorychatt:
Looks good so far, let's merge this

PR #1355, opened 2025-11-04
(charts): display more space on top of charts for better readability

defymecobra:
Added optimal Y-axis calculation in `sharedUtils.ts`:
- Implemented `calculateOptimalInterval()` to determine clean interval values (1, 2, 3, 5, 10, etc.) based on data range
- Implemented `calculateOptimalMax()` to round the Y-axis maximum up to the next optimal division
- This ensures consistent spacing between all grid lines and adds visual breathing room at the top

Now they look like this:

image
image
image

### ALSO

Added proper maximum value calculation for stacked bar charts in `BarChartWidget.tsx`:
- Implemented `actualMaxValue` computation that correctly sums stacked values
- Previous implementation only considered individual series maximums, not the total stack height
- Now passes the correct maximum to `generateYAxis()`, ensuring proper scaling for stacked bars

### DIFF
--- a/frontend/src/widgets/charts/sharedUtils.ts
+++ b/frontend/src/widgets/charts/sharedUtils.ts
@@ -230,54 +230,6 @@ export const generateYAxis = (
 ) => {
   const safeTransform = transformValue ?? ((v: number) => v);
 
-  // Calculate optimal interval between Y-axis divisions
-  const calculateOptimalInterval = (dataMax: number, splitNum: number = 5) => {
-    // Handle zero or negative values using absolute value
-    const absDataMax = Math.abs(dataMax);
-    if (absDataMax <= 0) return 1;
-
-    // Calculate rough interval between divisions
-    const roughInterval = absDataMax / splitNum;
-
-    // Guard against zero or negative interval before log calculation
-    if (roughInterval <= 0) return 1;
-
-    // Find order of magnitude (1, 10, 100, 1000...)
-    const magnitude = Math.pow(10, Math.floor(Math.log10(roughInterval)));
-
-    // Normalize interval relative to magnitude
-    const normalized = roughInterval / magnitude;
-
-    // Choose optimal interval: 1, 2, 3, 5, or 10
-    let optimalInterval;
-    if (normalized < 1.5) optimalInterval = 1 * magnitude;
-    else if (normalized < 2.5) optimalInterval = 2 * magnitude;
-    else if (normalized < 4) optimalInterval = 3 * magnitude;
-    else if (normalized < 7.5) optimalInterval = 5 * magnitude;
-    else optimalInterval = 10 * magnitude;
-
-    return optimalInterval;
-  };
-
-  // Calculate optimal maximum for Y-axis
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

defymecobra:
@claude review this

defymecobra:
@rorychatt pls look at changes in `BarChartWidget.tsx`. With my new logic in `sharedUtils`, Area and Line charts work correctrly, but bar isn't, cause the logic of maxValue in it was broken until I've made my changes

rorychatt:
@defymecobra I think solution should be way simpler and most likely is some config we need to update

https://echarts.apache.org/examples/en/editor.html?code=PYBwLglsB2AEC8sDeAoWsDGwA2wBOAXLANoBEAxAMyUCcAHACIBCpAugDRqxjA6QhFU6dGDwQA5uICmhWKQCGADwgBnUp2GwlqgArAI0MDMFdNYAJ4gpRUioAW8gCbAA7qVOwAvl08bY4sUdBXy5FAEFlFSJiIWELKxsMeSNxfHN1D0dk-WjSAFkYdTkAFQBXKSLSAHUpR0riu1LKgDExSoBlZI7S6DYfDi5zCNVo2JFLazkAN3lscozNAHpF2AAjYB6svHMAcXkBEgAGdlhD1n6_FRkIKSiSMdhoeQBbSdJAF7dAUuNAfujAO9SFuITGyreR4AHoEF4KoQRxgOw2ABshwApODYFkwDkSABGY6wACsACYToTDnjqAAWE4U3Enah4wmk87oTznTxAA&_source=echarts-doc-preview

See this -> it renders properly in their examples

rorychatt (changes requested):
there has to be another solution that was made by the library

IDK maybe this is not it, you have to try
https://echarts.apache.org/en/option.html#yAxis.boundaryGap

defymecobra:
Deleted the previous complex solution and implemented a simpler solution using conditional max property:
- When largeSpread is false, the max property is not added to the axis configuration at all
- ECharts' native auto-scaling algorithm kicks in, automatically calculating optimal Y-axis range
- This provides the desired "more spacing" at the top of charts without custom logic
- ECharts ensures consistent intervals and rounds to nice numbers (100, 200, 300, etc.)
- Works correctly for all chart types including stacked bar charts

This approach leverages ECharts' built-in intelligence rather than reimplementing it, resulting in cleaner, more maintainable code.

Results (all works correclty):

image
image
image
image
image

defymecobra:
@claude review this

frontend/src/widgets/charts/sharedUtils.ts
```diff
@@ -258,7 +258,7 @@ export const generateYAxis = (
     },
     splitNumber: largeSpread ? 3 : 5,
     min: largeSpread ? safeTransform(minValue) : 0,
-    max: largeSpread ? safeTransform(maxValue) : 'dataMax',
```
rorychatt, line 261: good catch.

Maybe similar issue is with resizing?

PR #1378, opened 2025-11-05
fix(tabs): prevent duplicate history entries when switching tabs

defymecobra:
**Root Cause:**
The OnTabSelect handler was calling client.Redirect() on every tab click 
event without checking if the selected index actually changed.

**Solution:**
Added a check to verify that selectedIndex.Value != @event.Value before 
updating the state and calling Redirect. This ensures browser history is 
only updated when the active tab actually changes.

**Result:**
- Tab switching now creates only one history entry per actual change
- Browser back/forward navigation works correctly
- No duplicate entries in browser history


defymecobra:
@claude review this

PR #1390, opened 2025-11-06
(async select): add search feature

rorychatt:
This pull request improves the search experience and performance of the async select input widget by making the search case-insensitive and implementing debounced filtering with loading state feedback. The main changes are focused on both backend query logic and frontend UI responsiveness.

image

**Search improvements:**

* Changed the category name filtering in `AsyncSelectInputApp` to be case-insensitive by converting both the query and the category names to lower case before comparison.

**UI/UX enhancements:**

* Replaced the previous list filtering mechanism in `AsyncSelectListSheet<T>` with a debounced filter input using `Throttle`, so queries are only sent after the user pauses typing, reducing unnecessary requests and improving performance.
* Added a loading state indicator to the async select list, showing "Loading..." while results are being fetched.
* Updated the UI to use a vertical layout with a search input and either a loading indicator or the filtered list results.

**Dependency update:**

* Added a missing `System.Reactive.Linq` import to support throttling/filtering logic in the async select input.


rorychatt:
@claude review this

rorychatt:
@claude review this

nielsbosma (dismissed):

Issue #1389, opened 2025-11-06
(download): button does not work and reloads the same page.


ArtemLazarchuk:
### Description

When clicking the “Download” button, instead of downloading the file, the page just reloads or opens the same view again. The file download action is not triggered.
<img width="1343" height="912" alt="Image" src="https://github.com/user-attachments/assets/6af95013-b98d-4392-9df7-1f95c06d865b" />

### Code Samples

```csharp

```

PR #1403, opened 2025-11-06
fix(Download): resolve download button opening in new tab instead of downloading files

ArtemLazarchuk:
## Problem

1. **Download URLs opening in new tab**: When using `Button.Url()` with download URLs (e.g., `/download/{connectionId}/{downloadId}`), the button was opening the URL in a new tab (`target="_blank"`) instead of triggering a file download.

2. **Routing conflicts**: Download paths were being intercepted by `PathToAppIdMiddleware` and converted to `appId` query parameters, causing:
   - Download endpoints to be treated as app routes
   - Infinite redirect loops when accessing download pages
   - Files not being downloaded properly

## Solution

1. **Added `/download` to excluded paths**: Updated `routing-constants.json` to exclude `/download` paths from being converted to appId parameters by `PathToAppIdMiddleware`.

2. **Fixed ButtonWidget download handling**: Modified `ButtonWidget.tsx` to detect download URLs (starting with `/download/`) and avoid using `target="_blank"` for these links, allowing files to download directly.

## Changes

- `frontend/src/routing-constants.json`: Added `/download` to `excludedPaths` array
- `frontend/src/widgets/ButtonWidget.tsx`: Added logic to detect download URLs and conditionally apply `target="_blank"` only for non-download links

### DIFF
--- a/frontend/src/routing-constants.json
+++ b/frontend/src/routing-constants.json
@@ -4,6 +4,7 @@
     "/webhook",
     "/auth",
     "/upload",
+    "/download",
     "/assets",
     "/fonts",
     "/_framework",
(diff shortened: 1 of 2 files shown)


zachwolfe (approved):
Just tested, can confirm that #1389 is a problem, and that it is fixed by this PR (also, not opening in a new tab is a nice touch)

PR #1424, opened 2025-11-07
[Field]: implement validators for fields

ArtemKhvorostianyi:
Screenshot 2025-11-08 at 00 54 09

Added client-side validation in TextInputWidget so standalone Field wrappers surface the same email/password feedback as forms, while deferring to server-provided errors when present

### DIFF
--- a/frontend/src/widgets/inputs/TextInputWidget.tsx
+++ b/frontend/src/widgets/inputs/TextInputWidget.tsx
@@ -608,6 +608,8 @@ export const TextInputWidget: React.FC<TextInputWidgetProps> = ({
   const eventHandler = useEventHandler();
   const [localValue, setLocalValue] = useState(value);
   const [isFocused, setIsFocused] = useState(false);
+  // Indicates that we requested validation from backend and are waiting for response
+  const [pendingValidation, setPendingValidation] = useState(false);
   const inputRef = useRef<HTMLInputElement | HTMLTextAreaElement | null>(null);
 
   // Update local value when server value changes and control is not focused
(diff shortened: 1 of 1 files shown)


ArtemKhvorostianyi:
@claude review this

frontend/src/widgets/inputs/TextInputWidget.tsx
```diff
@@ -51,6 +51,12 @@ interface TextInputWidgetProps {
   'data-testid'?: string;
 }
 
+const EMAIL_VALIDATION_MESSAGE = 'Please enter a valid email address';
```
rorychatt: ban

frontend/src/widgets/inputs/TextInputWidget.tsx
```diff
@@ -664,21 +668,30 @@ export const TextInputWidget: React.FC<TextInputWidgetProps> = ({
 
   const handleBlur = useCallback(() => {
     setIsFocused(false);
+    // Mark field as pending validation so it shows invalid style until backend responds
+    setPendingValidation(true);
```
rorychatt, line 672: you are still handling state here in the frontend.

Right now the strategy was to never do this.

We are planning on a smarter system like this in the future, where you would get a "best guess" before the system's response arrives. But right now, we follow the flow

FE -> event to BE -> state update in BE -> event to FE -> FE applies patch

with this, you do a different scheme. you do

```text
FE |-> FE guess patch -> ...
     |-> Event to BE -> state update in BE -> event to FE -> FE applies patch
```

PR #1426, opened 2025-11-07
[FileInput]: make file input background transperent to fit different elements

ArtemKhvorostianyi:
We had an issue where the file input had a black background even when it wasn’t placed inside an element with a dark background, which caused a strange-looking UI.

Screenshot 2025-11-08 at 01 19 12

Fixed:
Screenshot 2025-11-08 at 01 19 51

Screenshot 2025-11-08 at 01 20 17


ArtemKhvorostianyi:
@claude review this

PR #1435, opened 2025-11-08
fix(supabase): allow user metadata fields to be null

zachwolfe:
This fixes an issue I just ran into with Sign In with Apple using Supabase auth.


rorychatt:
@claude review this

PR #1439, opened 2025-11-09
[Chat]: refactor docs

ArtemLazarchuk:
### Summary

- Fix chat docs to show user messages before responses. Updated each demo’s handler so the user’s message is added to state immediately, guaranteeing it renders in the transcript before any assistant response or status indicator.
- Clarify the async loading flow. Expanded the loading-state description to outline how the chat appends the user message, inserts a ChatStatus “thinking” indicator during async work, and then swaps that status for the assistant’s final answer once processing completes

image

image

image


rorychatt:
(chat): docs needs a check if it's valid

PR #1446, opened 2025-11-10
fix: make kanban swaps consistent

defymecobra:
- Fix backend HandleMove logic so drag-and-drop swaps are consistent within and across columns
- Preserve original priorities when cards move, respecting backend state
- Swap cards when dropping onto another column’s card; add safeguard to avoid duplicates when moving between columns

https://github.com/user-attachments/assets/3d38e2ad-527a-45c0-b8b9-73b61fcd5496

### DIFF
--- a/Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
+++ b/Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Diagnostics;
 using System.Linq.Expressions;
 using Ivy.Shared;
 using Ivy.Views.Kanban;
@@ -124,12 +125,15 @@ public class KanbanApp : SampleBase
     private static KanbanState AddTask(KanbanState state, string columnKey, Task task)
     {
         var columns = state.Columns;
-        var column = columns.TryGetValue(columnKey, out var items)
-            ? items.Add(task)
+        var columnExists = columns.TryGetValue(columnKey, out var items);
+        var column = columnExists
+            ? items!.Add(task)
             : ImmutableArray.Create(task);
 
         var updatedColumns = columns.SetItem(columnKey, column);
-        var updatedOrder = BuildColumnOrder(updatedColumns, state.ColumnOrder);
+        var updatedOrder = columnExists
+            ? state.ColumnOrder
+            : BuildColumnOrder(updatedColumns, state.ColumnOrder);
 
         return state with { Columns = updatedColumns, ColumnOrder = updatedOrder };
     }
@@ -152,7 +156,7 @@ public class KanbanApp : SampleBase
                 ? existingTarget.ToBuilder()
                 : ImmutableArray.CreateBuilder<Task>();
 
-        var originalIndex = IndexOfTask(sourceBuilder, taskId);
+        var originalIndex = IndexOfTask(sourceBuilder, taskId, moveData.FromColumn);
         var movingTask = sourceBuilder[originalIndex];
 
         if (sameColumn && moveData.TargetIndex.HasValue)
@@ -189,7 +193,9 @@ public class KanbanApp : SampleBase
                 .SetItem(moveData.FromColumn, sourceBuilder.ToImmutable())
                 .SetItem(moveData.ToColumn, targetBuilder.ToImmutable());
 
-        var updatedOrder = BuildColumnOrder(updatedColumns, state.ColumnOrder);
+        var updatedOrder = ShouldRebuildColumnOrder(state.Columns, updatedColumns)
+            ? BuildColumnOrder(updatedColumns, state.ColumnOrder)
+            : state.ColumnOrder;
 
         return state with { Columns = updatedColumns, ColumnOrder = updatedOrder };
     }
@@ -205,7 +211,9 @@ public class KanbanApp : SampleBase
         builder.RemoveAt(index);
 
         var updatedColumns = state.Columns.SetItem(columnKey, builder.ToImmutable());
-        var updatedOrder = BuildColumnOrder(updatedColumns, state.ColumnOrder);
+        var updatedOrder = ShouldRebuildColumnOrder(state.Columns, updatedColumns)
+            ? BuildColumnOrder(updatedColumns, state.ColumnOrder)
+            : state.ColumnOrder;
 
         return state with { Columns = updatedColumns, ColumnOrder = updatedOrder };
     }
@@ -226,7 +234,7 @@ public class KanbanApp : SampleBase
         (column[boundedTarget], column[originalIndex]) = (column[originalIndex], column[boundedTarget]);
     }
 
-    private static int IndexOfTask(ImmutableArray<Task>.Builder column, string taskId)
+    private static int IndexOfTask(ImmutableArray<Task>.Builder column, string taskId, string columnKey)
     {
         for (var i = 0; i < column.Count; i++)
         {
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

defymecobra:
@claude review this

defymecobra:
@claude review this

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -70,39 +73,45 @@ public class KanbanApp : SampleBase
...
-                        Description = taskToMove.Description,
-                        Assignee = taskToMove.Assignee
-                    };
+                        return;
+                    }
+
+                    var originalIndex = sourceList.FindIndex(t => t.Id == taskId);
+                    if (originalIndex < 0)
+                    {
+                        return;
```
rorychatt: it feels weird that this needs to be done. maybe the type should be uint in original class? Because card order is an array index?

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -70,39 +73,45 @@ public class KanbanApp : SampleBase
...
-                        // Insert at specific position
-                        var insertIndex = moveData.TargetIndex.Value;
-                        updatedTasks.InsertRange(insertIndex, new[] { taskToMove });
+                        targetList = new List<Task>();
+                        columns[moveData.ToColumn] = targetList;
                     }
-                    else
+
+                    var handled = moveData.FromColumn == moveData.ToColumn && moveData.TargetIndex.HasValue
+                        ? TrySwapWithinColumn(targetList, originalIndex, moveData.TargetIndex.Value)
```
rorychatt: how is tryswapwithin column and moveacross columns different, if the only event the BE receives has info of both the column the card should be placed at, and it's position?

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
         ).Width(Size.Rem(32));
     }
 
+    private static Dictionary<string, List<Task>> BuildColumnMap(Task[] source)
+    {
```
rorychatt, line 126: there was a selector that defines which property is the column name

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
...
+        if (boundedTarget == originalIndex)
+        {
+            return false;
+        }
+
+        var sourceTask = column[originalIndex];
+        var targetTask = column[boundedTarget];
+
+        column[boundedTarget] = CloneTask(sourceTask, sourceTask.Status);
+        column[originalIndex] = CloneTask(targetTask, targetTask.Status);
```
rorychatt: maybe we should have used immutable arrays here
we need to make this thread safe too (hinted by you using something called "clone")

-> meaning what if same card is moved by two clients?

rorychatt: Swap naming suggests that A is to be replaced by B, but here we only place at an index

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
...
+        return true;
+    }
+
+    private static bool MoveAcrossColumns(
```
rorychatt: arguably move across columns is the only handler we need, and thus could have a better naming

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
...
+        List<Task> targetList,
+        Task movingTask,
+        int originalIndex,
+        int? targetIndex,
+        string fromColumn,
+        string toColumn)
+    {
+        if (originalIndex < 0 || originalIndex >= sourceList.Count)
+        {
+            return false;
```
rorychatt: same here, maye we need to implement something so these checks are unneccessary (from the event itself)

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
...
+        string targetColumn)
+    {
+        var orderedStatuses = original
+            .Select(t => t.Status)
+            .Concat(new[] { targetColumn })
+            .Distinct()
+            .OrderBy(GetStatusOrder)
+            .ThenBy(t => t);
+
+        var reorderedTasks = new List<Task>();
```
rorychatt: I am afraid that in general in this solution we create a new array too many times which could be bad...

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -150,6 +159,117 @@ private object BuildTaskSheet(IState<string?>? selectedTaskId, IState<Task[]> ta
...
+            if (columns.TryGetValue(status, out var list))
+            {
+                reorderedTasks.AddRange(list);
+            }
+        }
+
+        return reorderedTasks.ToArray();
+    }
+
+    private static Task CloneTask(Task task, string status) => new Task
```
rorychatt: also maybe a hint that we are doing something wrong

rorychatt: have you tried `with` keyword?

defymecobra:
- Switch kanban task state to ImmutableArray and treat Task as a record
- Collapse move logic into a single handler that covers swaps and cross-column inserts
- Rebuild columns via the original selector and avoid extra array copies

defymecobra:
@claude review this

defymecobra:
- rework kanban storage into `KanbanState` backed by immutable column map
- consolidate move/add/delete handlers to mutate only touched columns
- render board via `FlattenTasks` without rebuilding full task arrays

defymecobra:
@claude review this

defymecobra:
@rorychatt what you think

defymecobra:
@claude review

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
```diff
@@ -1,43 +1,40 @@
...
-            new Task { Id = "10", Title = "Unit Tests", Status = "Done", Priority = 1, Description = "Write comprehensive test suite", Assignee = "Bob" },
-            new Task { Id = "11", Title = "Deploy to Production", Status = "Done", Priority = 2, Description = "Configure CI/CD pipeline", Assignee = "Charlie" },
-            new Task { Id = "12", Title = "User Training", Status = "Done", Priority = 3, Description = "Train users on new features", Assignee = "Alice" },
-        });
+        var tasks = UseState(CreateInitialState(CreateInitialTasks()));
+        Expression<Func<Task, string>> columnSelector = task => task.Status;
 
-        var kanban = tasks.Value
+        var kanban = FlattenTasks(tasks.Value)
                 .ToKanban(
```
rorychatt, line 36: .ToKanban should work on IEnumerable (for example, arrays)

Ivy.Samples.Shared/Apps/Widgets/KanbanApp.cs
rorychatt: I feel like a bunch of logic should maybe be moved to the kanban widget.

End user should be able to do just this (API):

example
```csharp

data = array<SomeType>

data.ToKanban(
      groupBySelector: ... 
      idSelector: ...
      titleSelector: ...
      descriptionSelector: ...
   )
   .HandleDelete()
   .HandleMove()
   .HandleClick()

```

If we need some internal state, it should be implemented inside kanban widget so users don't have to do all this over and over.

rorychatt (changes requested):

rorychatt:
I am doing a refactor that might fix this. Put this on hold for now

PR #1454, opened 2025-11-11
Attempt to use Squirrel

sudipto80:


rorychatt:
We will have a discussion on Friday about this. But we are super careful in which dependencies we use in the Ivy-Framework because we are used by enterprises. We need to address any potential security concerns.

rorychatt:
We can not merge another dependency in the core framework right now

PR #1483, opened 2025-11-12
[Callout]: make links presented in callouts clickable

ArtemKhvorostianyi:
Links presented in callouts just weren't recognized as links
Screenshot 2025-11-13 at 01 00 31

Screenshot 2025-11-13 at 01 01 50

### DIFF
--- a/Ivy/Widgets/Primitives/Callout.cs
+++ b/Ivy/Widgets/Primitives/Callout.cs
@@ -85,10 +85,6 @@ public static class CalloutExtensions
         return callout with { Icon = icon };
     }
 
-    /// <summary>Sets link click event handler for Callout widget by setting it on the internal Markdown widget.</summary>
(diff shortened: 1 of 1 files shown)


Ivy/Widgets/Primitives/Callout.cs
```diff
@@ -82,4 +84,20 @@ public static Callout Icon(this Callout callout, Icons icon)
     {
         return callout with { Icon = icon };
     }
+
+    /// <summary>Sets link click event handler for Callout widget by setting it on the internal Markdown widget.</summary>
+    /// <param name="callout">Callout widget to configure.</param>
+    /// <param name="onLinkClick">Event handler for link clicks.</param>
+    /// <returns>Callout widget with link click handler set on its Markdown child.</returns>
```
rorychatt: we dont need excessive xml docs. In this case you can understand what the function does based on its signature and xml docs add no value

PR #1500, opened 2025-11-13
(feat): routing new way

defymecobra:
## Summary
- move Ivy’s internal endpoints (upload, download, webhook, auth token, health, SignalR, assets) under a unified `/ivy` prefix
- update routing constants and docs to reflect the new reserved paths

## Testing
- File uploads and downloads in samples (`/ivy/upload`, `/ivy/download`)
- OAuth docs: webhook URL now `http://localhost:5010/ivy/webhook`
- Auth token flow (`PATCH /ivy/auth/set-auth-token`)
- SignalR handshake over `/ivy/messages`
- Health check at `/ivy/health`
- Docs assets load through `/ivy/assets`

### DIFF
--- a/Ivy.Samples.Shared/Apps/Concepts/AssetsApp.cs
+++ b/Ivy.Samples.Shared/Apps/Concepts/AssetsApp.cs
@@ -7,6 +7,6 @@ public class AssetsApp : ViewBase
 {
     public override object? Build()
     {
-        return new Image("/assets/tagline.png");
+        return new Image("/ivy/assets/tagline.png");
     }
 }


defymecobra:
In `frontend/src/routing-constants.json` I still exclude `/assets`, `/fonts`, `/_framework`. Vite’s dev server serves static resources on these paths, so if we drop them the middleware (and `getAppId`) will start rewriting asset requests into `?appId=` and break resource loading. The new `/ivy` entry covers our internal endpoints; the legacy exclusions stay for compatibility with dev mode.

nielsbosma:
@rorychatt  Let's say I add a Controller in an app with some routes. How can I get these so routing-constants.json isn't blocking, and I don't want to put the route under /ivy/.

defymecobra:
> @rorychatt Let's say I add a Controller in an app with some routes. How can I get these so routing-constants.json isn't blocking, and I don't want to put the route under /ivy/.

Right now it rewrites every request that isn’t under `/ivy` (or another entry from `frontend/src/routing-constants.json`) into `/?appId=…`, so any controller that exposes `/my-controller` gets interpreted as an app route and never executes. To keep a custom controller reachable, the quickest fix is to list its root segment (e.g. `/my-controller`) in `excludedPaths` and restart the backend so the embedded routing constants pick up the change. If we prefer to keep the controller outside of `/ivy` without touching the shared JSON, we can add a configuration-driven escape hatch. The middleware could merge `routing-constants.json` with an `appsettings` section (or similar) so individual apps can declare their reserved paths, and the CLI template could accept a `--route` flag to populate that automatically. That would let add service endpoints without editing frontend assets.

zachwolfe:
@nielsbosma @defymecobra Another idea that may or may not work is to ask ASP.NET for all routing endpoints on startup or hot reload. Then we could use those in `PathToAppIdMiddleware` and pass them to the frontend via `WebApplicationExtensions.UseFrontend()`. That way, Ivy app developers would not need to manually configure routes.

This also highlights the need for explicit handling of collisions between app IDs and routes, I think. E.g., if the app developer defines a `"/homepage"` route and a `HomepageApp` app with its default app ID of "homepage", we should probably just fail to launch the server, because bad things will surely happen otherwise.

I don't think any of the above needs to be added to this PR, though. :)

zachwolfe (approved):

PR #1502, opened 2025-11-14
[HotFix]: revert braking changes for Docs

ArtemKhvorostianyi:
We had a problem that a lot of examples in docs stopped to work (no errors, no warnings, it just stopped to be clickable)
Broken Doc pages:
Sheet doesn't work
Job Scheduller doesn't work
Kanban -> all actions with kanban don't work (drag and drop, delete etc)
Floating panel doesn't work 
...
All the pages that had interactive examples didn't work. 
The problem was in the Tabs refactor, I just reverted changes from TabsLayoutWidget.tsx to its old version, nothing more. 
I tested out, tabs seem to work fine (as for new refactored and redesigned verstion). 
@dcrjodle Can you please have a look on tabs from this branch ? I reverted Your changes in the TabsLayoutWidget.tsx, it seems to be the same, however, nothing is broken

### DIFF
--- a/frontend/src/widgets/layouts/tabs/TabsLayoutWidget.tsx
+++ b/frontend/src/widgets/layouts/tabs/TabsLayoutWidget.tsx
@@ -1,19 +1,173 @@
 import React from 'react';
-import type { TabsLayoutWidgetProps } from './types';
+import { Tabs, TabsList, TabsTrigger, TabsContent } from '@/components/ui/tabs';
+import Icon from '@/components/Icon';
+import { Button } from '@/components/ui/button';
 import {
-  filterTabWidgets,
-  orderTabWidgets,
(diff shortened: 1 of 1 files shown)


dcrjodle:
I'm trying out the docs project on my local with frontend built and running, and I'm unable to reproduce this

ArtemKhvorostianyi:
@dcrjodle , try to run project from the main branch and have a look at some doc page. For example, Docs -> Sheet, and try to interact with examples. This and many others pages are broken (examples aren't clickable).

I asked your review because changes touch tabs widget that you've changed, so I ask if I broke something about tabs

rorychatt:
I like the brave move of just reverting the commit :D

But we want the new design, solution is to fix the bug by fixing it not removing

PR #1505, opened 2025-11-14
refactor(tabs): remove stableChildren stabilization logic

dcrjodle:
Removing a fix for tabs re-rendering content, because it caused a bug in content tabs in Docs


dcrjodle:
@ArtemKhvorostianyi this is the revert we need to do

ArtemKhvorostianyi:
That works, thanks!

PR #1513, opened 2025-11-15
(xml-docs): minification

rorychatt:
Some part of this I ran through an automation. 
Some part of this I had removed myself.

More PR-s regarding this might be created soon. Idea is that we only create XML docs if they add value. If something can be generated just by looking at the function signature, don't add it.

Good xml docs to add are:

- notes about side effects (like `this also deletes your system OS`
- clarification of returns (like `this returns null also when an error is caught`)
- throws (like `throws InvalidException when ...`
- anything that basically is not signature
- come clues on general use case, most frequent use cases, etc


rorychatt:
@claude review this

defymecobra:
I think all done

PR #1517, opened 2025-11-16
[Field]: implement proper validation

ArtemKhvorostianyi:
Screenshot 2025-11-16 at 02 55 54

About password validation: we don't have implemented default password validation (for example, in form fields we always implement custom validation on form creation, it's not handled in BE by defailt)

Should we implement default password validation? However, email validation is fine now


ArtemKhvorostianyi:
@claude review this

Ivy/Widgets/Inputs/Field.cs
```diff
@@ -1,14 +1,106 @@
...
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 
 
 namespace Ivy;
+public class FieldView : ViewBase
```
rorychatt, line 16: don't we already have a field widget?

ArtemKhvorostianyi:
Yes, we already have a Field widget. This PR keeps Field and introduces FieldView plus a WithField() helper to enable auto email/password validation and fluent config when wrapping inputs. The previous backend implementation of the field couldn’t handle validation, so I refactored it. The changes are tested and don’t break anything.

rorychatt:
> Yes, we already have a Field widget. This PR keeps Field and introduces FieldView plus a WithField() helper to enable auto email/password validation and fluent config when wrapping inputs. The previous backend implementation of the field couldn’t handle validation, so I refactored it. The changes are tested and don’t break anything.

I am not sure if this is what we want as this will bring a lot of confusion.

PR #1519, opened 2025-11-16
feat(routing): allow custom routes to be defined by applications

defymecobra:
# Allow custom routes to be defined by applications

## Problem

Developers cannot define custom controller routes (e.g., `/my-controller`) without manually updating `routing-constants.json`. Custom routes are incorrectly treated as application IDs by `PathToAppIdMiddleware` and get converted to query parameters (e.g., `/my-controller` → `/?appId=my-controller`).

This prevents creating custom API endpoints or controllers without modifying the framework's routing configuration.

## Solution

Implemented a minimal refactoring that automatically discovers and excludes custom controller routes from `appId` conversion:

### Changes

1. **`PathToAppIdMiddleware.cs`**:
   - Added check for already matched endpoints using `context.GetEndpoint()` (after `UseRouting()` executes)
   - Added fallback check using `EndpointDataSource` to discover registered routes before routing executes
   - Added `GetRoutePath()` helper method to extract route paths from `RoutePattern` (handles both `RawText` and segment-based routes)
   - Optimized by replacing LINQ `Any()` with `foreach` loops

2. **`Server.cs`**:
   - Registered controllers from the entry assembly (user's application) in addition to Ivy assembly
   - Moved `app.UsePathToAppId()` middleware after `app.UseRouting()` so routing can match endpoints first
   - Added comments explaining the middleware order

## How It Works

1. When a request comes in, `PathToAppIdMiddleware` first checks if routing has already matched an endpoint via `context.GetEndpoint()`
2. If not found, it checks `EndpointDataSource` to see if the path matches any registered controller route
3. If a match is found, the middleware skips `appId` conversion and lets the request proceed to the matched controller
4. If no match is found, it falls back to the original behavior (checking excluded paths, static file extensions, etc.)

## Benefits

- Developers can define custom routes without modifying `routing-constants.json`
- Automatic route discovery - no manual configuration needed
- Backward compatible - existing behavior for app IDs remains unchanged
- Minimal changes - only 2 files modified
- Performance optimized - early exits and efficient route matching

## Testing

To verify this works:

1. Create a custom controller in your application:
   ```csharp
   [ApiController]
   [Route("my-custom-route")]
   public class MyCustomController : ControllerBase
   {
       [HttpGet]
       public IActionResult Get() => Ok(new { message = "Success!" });
   }
   ```

2. Access `http://localhost:5010/my-custom-route` - it should work correctly without being converted to `?appId=my-custom-route`

3. Existing app IDs (e.g., `/my-app`) should still work as before, being converted to `/?appId=my-app`

## Example

**Before**: 
- `/my-controller` → `/?appId=my-controller` ❌
- Required manual update to `routing-constants.json`

**After**:
- `/my-controller` → handled by `MyController` ✅
- `/my-app` → `/?appId=my-app` ✅ (still works)
- No configuration needed!

### DIFF
--- a/Ivy/Ivy.csproj
+++ b/Ivy/Ivy.csproj
@@ -58,7 +58,7 @@
     <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.10.2" />
     <PackageReference Include="Isopoh.Cryptography.Argon2" Version="2.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="9.0.9" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="9.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR" Version="1.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.SignalR" Version="1.2.0" />
     <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.10.2-preview.1.25552.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="9.0.10" />


defymecobra:
@claude review this

rorychatt:
@claude review this

rorychatt:
Look at claude's review, some valid points

defymecobra:
Honestly, I need to redo this PR completely. Will do it as soon as other issues with higher priority will be solved.

PR #1530, opened 2025-11-17
(datatable): Support Lucide react icons in the column headers

### LINKED ISSUE #1496: (datatable): Remove Glide Data grid custom implementations from the API
for example the Glide Data Grid icon enum

dcrjodle:
closes #1496

### DIFF
--- a/Ivy/Widgets/DataTables/DataTableColumn.cs
+++ b/Ivy/Widgets/DataTables/DataTableColumn.cs
@@ -114,28 +114,3 @@ public class LabelsDisplayRenderer : IDataTableColumnRenderer
 {
     public bool IsEditable => false;
 }
-
-/// <summary>
-/// Glide Data Grid compatible icon names for DataTableColumn.
-/// These icon names correspond to the icons defined in the frontend headerIcons.ts file.
-/// </summary>
-public static class DataTableIcons
-{
-    public const string User = "User";
-    public const string Mail = "Mail";
-    public const string Hash = "Hash";
-    public const string Calendar = "Calendar";
-    public const string Clock = "Clock";
-    public const string Activity = "Activity";
-    public const string Flag = "Flag";
-    public const string Zap = "Zap";
-    public const string Info = "Info";
-    public const string ChevronUp = "ChevronUp";
-    public const string ChevronDown = "ChevronDown";
-    public const string Filter = "Filter";
-    public const string Search = "Search";
-    public const string Settings = "Settings";
-    public const string MoreVertical = "MoreVertical";
-    public const string HelpCircle = "HelpCircle";
-}
-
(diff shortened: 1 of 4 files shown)


frontend/src/widgets/dataTables/types/types.ts
```diff
@@ -95,16 +93,71 @@ export enum SelectionModes {
   Columns = 'Columns',
 }
 
+/**
+ * Glide Data Grid compatible icon names for DataTableColumn.
+ * These icon names correspond to the icons defined in the frontend headerIcons.ts file.
+ */
+export enum DataTableIcon {
+  User = 'User',
```
rorychatt, line 101: this looks super sus

Ivy/Widgets/DataTables/DataTableColumn.cs
```diff
@@ -117,36 +108,9 @@ public enum LinkDisplayType
 public class ProgressDisplayRenderer : IDataTableColumnRenderer
 {
     public bool IsEditable => false;
-    public ColType ColType => ColType.Number;
 }
```
rorychatt, line 122: PR states that you moved icon enum to backend, but you did the opposite?

Ivy/Widgets/DataTables/DataTableColumn.cs
```diff
@@ -117,36 +108,9 @@ public enum LinkDisplayType
...
 
 public class LabelsDisplayRenderer : IDataTableColumnRenderer
 {
```
rorychatt, line 124: this should have used Icons.Mail, Icons.User and so on instead?

Ivy/Widgets/DataTables/DataTableColumn.cs
```diff
@@ -117,36 +108,9 @@ public enum LinkDisplayType
...
     public bool IsEditable => false;
```
rorychatt, line 125: And I don't really understand why we don't just support all Icons from Icons enum

dcrjodle:
oops sorry, we're moving them to the frontend because we want the backend to be completely agnostic

rorychatt:
> oops sorry, we're moving them to the frontend because we want the backend to be completely agnostic

so there will now be no compile time error if this is misused, a runtime one?
image

image


I am not so sure if this is how we want. Maybe solution is to use the enums instead of strings where possible in the backend, and then in the rendererer in the frontend use the svg of the icon library in the canvas?

@nielsbosma

Ivy/Widgets/DataTables/DataTableColumn.cs
dcrjodle: glide data grid have built in icons that aren't part of what the framework support

dcrjodle:
I will think of a better solution for the icons. Let's try to use lucide icons with custom renderers in the column headers

PR #1538, opened 2025-11-17
deleted sub directory from root

dcrjodle:


rorychatt:
@dcrjodle I am just curious (others too)
What had happened here?

dcrjodle:
@rorychatt claude code multi branching chaos. won't happen anymore since I've switched to cursor

Issue #1295, opened 2025-10-30
[Links]: links presented in callouts aren't clickable


ArtemKhvorostianyi:
### Description

<img width="1273" height="537" alt="Image" src="https://github.com/user-attachments/assets/3a4828dc-52af-4200-add5-bca4eddedd14" />

### Steps to Reproduce

Docs - > Program -> Authentication

### Code Samples

```csharp

```

rorychatt:
This broke after our URL state changes. 
Make this work with new URL changes

rorychatt:
https://github.com/Ivy-Interactive/Ivy-Framework/pull/1535

@ArtemKhvorostianyi confirm this is fixed in this PR

KaiserReich95:
It seems that the links where clickable after the #1535 fix but didnt lead to the right page.

rorychatt:
Something is off with the routing/where links lead to. 
Since two people are working on this and @ArtemKhvorostianyi has a lot on his table, @KaiserReich95 will you be able to solve this?

KaiserReich95:
Sure, it seems all the test are successful now. I will check if we need more tests to cover some code/links we are missing and try to fix the problem with the links.

rorychatt:
@KaiserReich95 true, but none of the docs stuff is tested

ArtemKhvorostianyi:
This issue was closed with this merged PR https://github.com/Ivy-Interactive/Ivy-Framework/pull/1483

Issue #1529, opened 2025-11-17
[Links]: In docs all the links redirect to Introduction to Ivy page


ArtemKhvorostianyi:
### Description

All the links in docs redirect user to default page  Introduction to Ivy
<img width="681" height="176" alt="Image" src="https://github.com/user-attachments/assets/27dd9faf-890a-4169-9336-40a62ffa9f48" />

<img width="1440" height="788" alt="Image" src="https://github.com/user-attachments/assets/be8ac230-2bf4-4f99-9f83-94c079d40312" />

### Code Samples

```csharp

```

rorychatt:
@ArtemKhvorostianyi can we close this?

PR #1559, opened 2025-11-18
(fix): links presented in callouts leads to auth overview

KaiserReich95:
https://github.com/Ivy-Interactive/Ivy-Framework/issues/1295

The link in callouts are now clickable and leads to Authentication Overview.

https://github.com/user-attachments/assets/1a5c7593-6c1e-46ee-8a8e-7839bd251e4a

### DIFF
--- a/Ivy.Docs.Tools.Test/UtilsTests.cs
+++ b/Ivy.Docs.Tools.Test/UtilsTests.cs
@@ -39,7 +39,7 @@ public class UtilsTests
 
     [Theory]
     [InlineData("Onboarding.InstallationApp", "onboarding/installation")]
-    [InlineData("Onboarding.GettingStarted.FirstAppApp", "onboarding/gettingstarted/firstapp")]
+    [InlineData("Onboarding.GettingStarted.FirstAppApp", "onboarding/getting-started/first-app")]
     public void GetAppIdFromTypeName_ReturnsExpected(string typeName, string expectedAppId)
     {
         var result = Utils.GetAppIdFromTypeName(typeName);


rorychatt:
@KaiserReich95 maybe this is a good example that we need to test that links work. We already have some e2e test core structure...
I notice that we fix this in 1 place, other spot breaks and so on.

KaiserReich95:
Add unit tests for link path resolution

- Add Windows backslash path handling test for GetPathForLink
- Add tests for GetTypeNameFromPath and GetAppIdFromTypeName
- Tests cover order prefix stripping, path segment conversion, and URL generation
- Changed so some tests have abstract placeholder names (Foo/Bar/Baz) instead of specific paths.

I think this will cover the problems we have with links redirecting to the introduction page. One of the tests failed on the current main branch, which is related to this issue #1295, and it now passes with the changes we made in this branch before. The other issue #1529 we had should also be covered by the tests now.

I'm not the expert on testing, so if there is anything more we need to test let me know.

KaiserReich95:
Fixed the abstractions since they are too abstract.

rorychatt:
@ArtemKhvorostianyi I want you to test out this branch. So it does not break anything that you did before on the same topic,

Ivy.Docs.Tools/Utils.cs
```diff
@@ -18,7 +18,8 @@ public static string GetPathForLink(string source, string link)
 
         // Normalize the source path to handle cross-platform separators
         var normalizedSource = source.Replace('\\', '/');
-        var sourceDir = Path.GetDirectoryName(normalizedSource) ?? "";
+        var lastSlashIndex = normalizedSource.LastIndexOf('/');
```
rorychatt, line 21: hard-coding to windows?
I think in this case we are OK due to previous line.

ArtemKhvorostianyi:
> @ArtemKhvorostianyi I want you to test out this branch. So it does not break anything that you did before on the same topic,

@rorychatt  , It works from branch main and doesn't need any fixes. 
I fixed this issue with links in callouts before, here is link https://github.com/Ivy-Interactive/Ivy-Framework/pull/1483
It was already fixed.

rorychatt:
I have tested this on current latest release.

This is not an issue there. I am closing this PR.

Let me know if this is a mistake.

PR #1560, opened 2025-11-19
(routing): Implement 404 error for unrecognized apps

defymecobra:
image

* Created `NotFoundView`: Added a new, lightweight view component to display a user-friendly "App Not Found" message.
* Implemented Virtual App Handling: Updated `AppHub` to detect invalid app IDs and dynamically generate a virtual `AppDescriptor` for the 404 view.
* Added `ScopedAppRepository`: Introduced a helper class to intercept app lookups, enabling the Chrome shell to render the 404 content seamlessly within a tab.
* Preserved Navigation State: Modified routing logic to keep the invalid URL in the browser address bar, preventing automatic redirection to the home page.
* Integrated with Chrome Shell: Configured the 404 error to appear as a standard tab with a "App Not Found" title and an alert icon.

### DIFF
--- a/Ivy/AppHub.cs
+++ b/Ivy/AppHub.cs
@@ -28,15 +28,15 @@ public class AppHub(
     IQueryableRegistry queryableRegistry
     ) : Hub
 {
-    public AppArgs GetAppArgs(string connectionId, string appId, string? navigationAppId, HttpContext httpContext)
+    private AppArgs GetAppArgs(string connectionId, string appId, string? navigationAppId, HttpContext httpContext)
     {
         string? appArgs = null;
-        if (httpContext!.Request.Query.TryGetValue("appArgs", out var appArgsParam))
+        if (httpContext.Request.Query.TryGetValue("appArgs", out var appArgsParam))
         {
             appArgs = appArgsParam.ToString().NullIfEmpty();
         }
 
-        HttpRequest request = httpContext.Request;
+        var request = httpContext.Request;
         return new AppArgs(connectionId, appId, navigationAppId, appArgs ?? server.Args?.Args, request.Scheme, request.Host.Value!);
     }
 
@@ -47,10 +47,7 @@ public class AppHub(
             var appServices = new ServiceCollection();
 
             var httpContext = Context.GetHttpContext()!;
-
-            var chrome = GetChrome(httpContext);
-            var parentId = GetParentId(httpContext);
-            var (appId, navigationAppId) = GetAppId(server, httpContext, chrome);
+            var parentId = AppRouter.GetParentId(httpContext);
 
             var clientProvider = new ClientProvider(new ClientSender(clientNotifier, Context.ConnectionId));
 
@@ -111,21 +108,31 @@ public class AppHub(
                 {
                     clientProvider.SetAuthToken(authToken, reloadPage: parentId != null && authToken == null);
                 }
-
-                if (authToken == null)
-                {
-                    appId = AppIds.Auth;
-                }
             }
 
             var appRouter = new AppRouter(server);
-            var routeResult = appRouter.Resolve(appId, navigationAppId, parentId, chrome);
+            var routeResult = appRouter.Resolve(httpContext);
+
+            // Override to Auth app if authentication failed
+            if (server.AuthProviderType != null)
+            {
+                var authService = appServices.BuildServiceProvider().GetService<IAuthService>();
+                if (authService?.GetCurrentToken() == null)
+                {
+                    var authApp = server.AppRepository.GetAppOrDefault(AppIds.Auth);
+                    routeResult = routeResult with
+                    {
+                        AppId = AppIds.Auth,
+                        AppDescriptor = authApp
+                    };
+                }
+            }
 
             appServices.AddSingleton(typeof(IAppRepository), routeResult.AppRepository);
 
-            var appArgs = GetAppArgs(Context.ConnectionId, appId, navigationAppId, httpContext);
+            var appArgs = GetAppArgs(Context.ConnectionId, routeResult.AppId, routeResult.NavigationAppId, httpContext);
 
-            logger.LogInformation("Connected: {ConnectionId} [{AppId}]", Context.ConnectionId, appId);
+            logger.LogInformation("Connected: {ConnectionId} [{AppId}]", Context.ConnectionId, routeResult.AppId);
 
             appServices.AddSingleton(appArgs);
             appServices.AddSingleton(routeResult.AppDescriptor);
@@ -141,8 +148,8 @@ public class AppHub(
 
             var appState = new AppSession
             {
-                AppId = appId,
-                MachineId = GetMachineId(httpContext),
+                AppId = routeResult.AppId,
+                MachineId = AppRouter.GetMachineId(httpContext),
                 ParentId = parentId,
                 AppDescriptor = routeResult.AppDescriptor,
                 App = app,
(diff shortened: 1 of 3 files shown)


defymecobra:
@claude review this

dcrjodle (approved):

Ivy/Views/NotFoundView.cs
```diff
@@ -0,0 +1,19 @@
...
+public class NotFoundView : ViewBase
+{
+    public override object? Build()
+    {
+        return Layout.Center()
+               | (Layout.Vertical()
+                   .Gap(4)
+                   .Center()
+                   | Text.H1("Ouch! :|").Bold()
+                   | Text.H3("Sorry, this app does not exist.")
```
nielsbosma: Remove the H3

rorychatt:
@defymecobra I will look at this together with Niels when we have some time for this. Maybe this is not the way we should do this. I will keep you updated, until then, don't do any refactors

defymecobra:
> @defymecobra I will look at this together with Niels when we have some time for this. Maybe this is not the way we should do this. I will keep you updated, until then, don't do any refactors
> 
> 

okay

rorychatt:
not quite happy with the API.

Can we instead try a solution with a "default app" that shows for any url (that can be overridden with a custom 404)?

zachwolfe:
> Can we instead try a solution with a "default app" that shows for any url

@rorychatt I kind of don't like the current behaviour on main, where the default app gets shown if the URL isn't valid (unless you meant something different from the current behaviour?). Being able to override the 404 app would be great though.

rorychatt:
While this may look good, let's try with the suggestion by Niels that I wrote @defymecobra 

> Can we instead try a solution with a "default app" that shows for any url (that can be overridden with a custom 404)?

defymecobra:
- Added AppIds.NotFound constant for the system 404 app identifier.
- Registered NotFoundView as a standard system app in the main Server constructor.
- Introduced Server.UseNotFoundApp method to allow developers to override the default 404 page.
- Refactored AppHub to use the registered 404 app instead of dynamic descriptor creation.
- Enhanced ScopedAppRepository to intercept invalid app IDs and map them to the 404 app while preserving the original URL.
- Removed redundant H3 header from NotFoundView for cleaner UI.
- Ensured IAppRepository is always registered in all code paths to prevent runtime errors.

defymecobra:
@claude review this

rorychatt (dismissed):

zachwolfe (changes requested):
This is working great for me with applications that use a chrome, but applications that don't use chrome end up redirecting to the default app, as they did before this PR. See video below:

https://github.com/user-attachments/assets/78f7f227-fd10-4c3f-9bf1-a243c2020d9a

defymecobra:
> This is working great for me with applications that use a chrome, but applications that don't use chrome end up redirecting to the default app, as they did before this PR. See video below:
> 
>  Screen.Recording.2025-11-28.at.10.49.34.PM.mov

okay i will see

zachwolfe:
@defymecobra FYI I still get the same behaviour shown in the video above after your latest commit (apologies if you already know this)

defymecobra:
Chrome shell was automatically redirecting URLs when opening non-existent apps in tabs, causing the 404 page to not display properly in Safari. Fixed by blocking redirects for NotFound app in all Chrome navigation points (opening new tabs, switching tabs, page navigation).

defymecobra:
@claude review this

defymecobra:
@zachwolfe pls try it now, it should work

Ivy/Chrome/DefaultSidebarChrome.cs
```diff
@@ -107,7 +111,9 @@ void OpenApp(NavigateArgs navigateArgs, bool replaceHistory = false)
 
                         // Update browser URL when switching to existing tab
                         var tab = tabs.Value[tabIndex];
-                        if (navigateArgs.Purpose is NavigationPurpose.NewDestination)
+                        bool isNotFoundTab = tab.AppId == AppIds.NotFound;
```
rorychatt: I am completely fine with bool here, but using var is recommended

rorychatt:
@defymecobra @zachwolfe 
I had tested this with both `?chrome=true` and `?chrome=false` and everything seems to work. Can you see if the issue still persists for you, @zachwolfe ?

zachwolfe:
@rorychatt Sorry, I should’ve been more clear. I was talking about applications that don’t have a chrome at all (e.g., where `UseChrome` is never called), not ones that do have a chrome which was then disabled with `?chrome=false`.

zachwolfe:
@rorychatt @defymecobra Here is my test app with no chrome. I'm unsure if `UseDefaultApp` or `UseAuth` are necessary to reproduce this problem or not. I have not yet tested with the latest commit (now reverted).

```csharp
CultureInfo.DefaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US");
var server = new Server();
server.UseAuth<BasicAuthProvider>();
#if DEBUG
server.UseHotReload();
#endif
server.AddAppsFromAssembly();
server.AddConnectionsFromAssembly();
server.UseDefaultApp(typeof(MainApp));
await server.RunAsync();
```

MainApp (nothing special):
```csharp
[App(path: ["Apps"])]
public class MainApp : ViewBase
{
    {
        return Layout.Horizontal().Align(Align.Center) |
            new HeaderLayout(Text.Large("Main App"), Layout.Vertical()
                        | Text.Small("Welcome to the Main App")
                        | new Button("Click Me")
            ).Width(Size.Full().Max(300));
    }
}
```

defymecobra:
> @rorychatt @defymecobra Here is my test app with no chrome. I'm unsure if `UseDefaultApp` or `UseAuth` are necessary to reproduce this problem or not. I have not yet tested with the latest commit (now reverted).
> 
> ```cs
> CultureInfo.DefaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US");
> var server = new Server();
> server.UseAuth<BasicAuthProvider>();
> #if DEBUG
> server.UseHotReload();
> #endif
> server.AddAppsFromAssembly();
> server.AddConnectionsFromAssembly();
> server.UseDefaultApp(typeof(MainApp));
> await server.RunAsync();
> ```
> 
> MainApp (nothing special):
> 
> ```cs
> [App(path: ["Apps"])]
> public class MainApp : ViewBase
> {
>     public override object? Build()
>     {
>         return Layout.Horizontal().Align(Align.Center) |
>             new HeaderLayout(Text.Large("Main App"), Layout.Vertical()
>                         | Text.Small("Welcome to the Main App")
>                         | new Button("Click Me")
>             ).Width(Size.Full().Max(300));
>     }
> }
> ```

I'm sorry, it's my mistake. I was testing this implementation in Samples, not in app. Sorry for confusion, now I see the problem clearly

defymecobra:
## Changes

Modified the app resolution logic in `AppHub.OnConnectedAsync` to handle 404 pages correctly when Chrome is not registered. When a navigation app is not found and Chrome app is not configured, the `appDescriptor` is now set directly to the NotFound app instead of falling back to the default app. Made `AppDescriptor` nullable to enable conditional assignment based on Chrome registration status.

## Testing

Tested with a test application without `UseChrome` - invalid routes correctly display the 404 page in default and `chrome=false` modes . Also verified with Ivy.Samples that has `UseChrome` configured - invalid routes still work correctly in both `chrome=true` (default) and `chrome=false` modes, with Chrome properly rendering the 404 page through IAppRepository.

defymecobra:
@claude review this

defymecobra:
@claude review this. I think you were wrong about all 3 critical issues.

defymecobra:
@zachwolfe now you can test it to ensure that everything works correctly

zachwolfe:
@defymecobra Great! I will test again shortly.

rorychatt:
hey sorry if you get pinged a lot, Zach, in the future will try to make it so there are less turns here and there.

zachwolfe:
@rorychatt hey, it’s no problem at all. 🙂

zachwolfe (dismissed):

rorychatt:
@defymecobra Can you test this solution now?

defymecobra:
> @defymecobra Can you test this solution now?

I've just tested. Everything works correctly.

rorychatt:
okay.

@zachwolfe You can merge this PR if you agree with the refactor that Niels did. Everything seems to work as expected now

rorychatt:
UPD: nevermind I'll just merge it we need this

PR #1561, opened 2025-11-19
fix: enable width and height support for field widget

defymecobra:
- Fixed an issue where `Width` and `Height` properties on `Field` widgets were ignored by the frontend.
- Updated `FieldWidget.tsx` to accept and apply `width` and `height` props to the component's container.
- Modified the container styles to conditionally remove the `flex-1` class when explicit dimensions are provided, ensuring the specified size takes precedence.
- Added a verification case in `FieldApp.cs` to demonstrate a `Field` with a constrained width.

Without width and height:
image

With width and height:
image


defymecobra:
@claude review this

PR #1565, opened 2025-11-19
fix: kanban column width serialization and CSS conflicts

defymecobra:
### **Problem Context**

*   The `ColumnWidths` property was defined as a dictionary with object keys, causing unpredictable JSON serialization behavior where keys were automatically converted to camelCase (e.g., "Todo" became "todo").
*   The frontend logic performed strict case-sensitive lookups for column widths, failing to match the serialized lowercase keys with the capitalized column IDs.
*   A default Tailwind CSS class `min-w-70` imposed a minimum width of approximately 280px, overriding any explicit width values smaller than this threshold.

### **Solution & Changes**

*   Changed the `ColumnWidths` property type in `Kanban.cs` from `Dictionary<object, Size>` to `Dictionary<string, string>` to ensure consistent serialization.
*   Updated `KanbanBuilder.cs` to explicitly convert `Size` values to their string representation during the build process.
*   Implemented a case-insensitive lookup strategy in `useKanbanData.ts` to reliably match column widths regardless of key casing differences.
*   Modified `kanban.tsx` to conditionally apply the `min-w-70` class only when no explicit width is defined, allowing for narrower custom columns.

Example of correct behavior:
image
image

### DIFF
--- a/Ivy/Views/Kanban/KanbanBuilder.cs
+++ b/Ivy/Views/Kanban/KanbanBuilder.cs
@@ -303,7 +303,7 @@ public class KanbanBuilder<TModel, TGroupKey> : ViewBase, IStateless
         }).ToArray();
 
         var columnWidthsDict = _columnWidths.Any()
-            ? _columnWidths.ToDictionary(kvp => kvp.Key, kvp => kvp.Value)
+            ? _columnWidths.ToDictionary(kvp => (object)kvp.Key, kvp => kvp.Value)
             : null;
 
         var kanban = new Ivy.Kanban(cards) with


defymecobra:
@claude review this

Ivy/Widgets/Kanban/Kanban.cs
```diff
@@ -18,7 +18,7 @@ public Kanban(params KanbanCard[] cards) : base([.. cards.Cast<object>()])
 
     [Prop] public bool AllowDelete { get; set; }
 
-    [Prop] public Dictionary<object, Size>? ColumnWidths { get; set; }
+    [Prop] public Dictionary<string, string>? ColumnWidths { get; set; }
```
rorychatt, line 22: this is our API.

In it, it should be Size for consistency with other API-s.

defymecobra:
- Revert ColumnWidths value type to Size in `Kanban.cs` to maintain API consistency with other widgets.
-  Update `KanbanBuilder.cs` to correctly serialize dictionary keys as strings while keeping values as Size objects.
- Optimize `useKanbanData.ts` by memoizing normalized column widths to improve performance.

defymecobra:
@claude review this

defymecobra:
Done

rorychatt:
@defymecobra Is that API change you did needed?
I guess the dict keys should be the same like everywhere else in widgets.

defymecobra:
> @defymecobra Is that API change you did needed?

no. sorry when I made the commit, I forgot to remove it

rorychatt:
@defymecobra okay. fix the branch and request another review when you are ready. Sorry if I broke it btw

defymecobra:
oh I see build isn't done. i'll fix it

defymecobra:
@claude review this

rorychatt:
@defymecobra I think I have already fixed this in another PR. Can you check if this issue is resolved on main?

defymecobra:
> @defymecobra I think I have already fixed this in another PR. Can you check if this issue is resolved on main?

No, it isn't resolved on main. The API for setting individual column widths is missing. The current implementation only supports a single ColumnWidth property that applies the same width to all columns globally.

PR #1583, opened 2025-11-23
[Chrome]: for ?chrome=false fix redirect to prev and next page links 

ArtemKhvorostianyi:
Screenshot 2025-11-23 at 08 57 09

Fixed

### DIFF
--- a/frontend/src/widgets/article/ArticleFooter.tsx
+++ b/frontend/src/widgets/article/ArticleFooter.tsx
@@ -1,6 +1,7 @@
 import { InternalLink } from '@/types/widgets';
 import { Github } from 'lucide-react';
 import React from 'react';
+import { convertAppUrlToPath, getChromeParam } from '@/lib/utils';
 
 interface ArticleFooterProps {
   id: string;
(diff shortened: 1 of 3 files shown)


frontend/src/widgets/article/ArticleFooter.tsx
```diff
@@ -17,17 +18,31 @@ export const ArticleFooter: React.FC<ArticleFooterProps> = ({
   documentSource,
   onLinkClick,
 }) => {
+  const isChromeFalse = !getChromeParam();
```
rorychatt, line 21: very bad naming IMO

frontend/src/widgets/article/ArticleFooter.tsx
```diff
@@ -17,17 +18,31 @@ export const ArticleFooter: React.FC<ArticleFooterProps> = ({
...
+
+  const handleLinkClick = (
```
rorychatt, line 23: is this the only thing that broke?

Are there other events that broke?

rorychatt:
@ArtemKhvorostianyi 

I mistakenly merged this.

Can you do the small refactors we talked about in a separate PR?

PR #1606, opened 2025-11-26
[Time]: style improvements

ArtemKhvorostianyi:
I found an issue with the time input’s background color. When the widget was placed inside an element with a different background, the time field kept its own background, which caused a visual mismatch:

Screenshot 2025-11-26 at 04 36 23

I also noticed that the time input had its icon positioned differently compared to the other date-related widgets. This made the layout feel inconsistent across the date/time components.

Both issues are now fixed:

Screenshot 2025-11-26 at 04 35 08

Fixed:
Screenshot 2025-11-26 at 04 33 57

And the background color behaves correctly as well. The last screenshot is just a small test example I wrote for myself to verify everything visually.

Screenshot 2025-11-26 at 04 40 53


rorychatt:
time frotnend code is becoming to big. we need to split it up.

PR #1562, opened 2025-11-19
feat: add cross-tab logout synchronization using Broadcast Channel API

defymecobra:
https://github.com/user-attachments/assets/696a16e7-ccb5-4739-94b1-b9613558b234

- Added `authChannelRef` to store the Broadcast Channel instance in `use-backend.tsx`.
- Extended `handleSetAuthToken` to broadcast logout events when `authToken` is `null`.
- Initialized Broadcast Channel on component mount with a message handler that reloads the page on logout.
- Added proper cleanup to close the Broadcast Channel on component unmount.
- Included browser compatibility check with fallback warning for unsupported browsers.

PR #1607, opened 2025-11-26
(auth): get tokens out of frontend, add `IAuthSession` interface with checked access

zachwolfe:
### Changes
- Auth tokens are no longer sent to the frontend. Instead, tokens are serialized and temporarily stored in a cookie registry on the backend, indexed by cryptographically secure ID. This ID is passed to the frontend and used to make an API call to the server, which then responds with the token as an HTTP-only cookie. Thus, the frontend never gets to see the token.
- All browser tabs with the same machine ID are now refreshed whenever the user logs _in_ (as opposed to #1562, which addressed the logging out case).
- A new `IAuthSession` interface is added which stores an `AuthToken` and a string that can be used to persist arbitrary session data into cookies. Most `IAuthProvider` methods now take this instead of a token.
- Due to the extreme flexibility provided by the `IAuthSession` interface (it allows getting and setting the token and session data at any time), extensive checks have been added and enabled in debug mode to aid with development. These take the form of the wrapper classes `CheckedAuthSession` and `CheckedAuthProvider`.
- `CheckedAuthSession` implements `IAuthSession` by wrapping another instance, but limiting access to `ReadOnly`, `WriteOnly`, or `ReadWrite` for each property (defaulting to no access allowed).
- `CheckedAuthProvider` implements `IAuthProvider` by wrapping another instance, but enforces `IAuthSession` access rules for each method using `CheckedAuthSession`. It also enforces some additional pre- and post-conditions.
- `IAuthProvider.SetHttpContext` has been refactored into `IAuthProvider.InitializeAsync`


zachwolfe:
@claude review please

zachwolfe:
I am going to expand the scope of this refactor so it can be more helpful for Clerk auth #1452

zachwolfe:
@claude review please

zachwolfe:
Now that I have tried these changes with the Clerk branch, I believe they are ready for review.

nielsbosma:
@zachwolfe why do we need the cookie concept? Why not use AppSessionStore?

Not sure I understand this PR fully. 

Also I think we should put everything that isn't a public API under Ivy.Core.

zachwolfe:
@nielsbosma
> why do we need the cookie concept?

I'll split this into two questions:

### Why do we need cookie jars?
I have plans to set other cookies in the future (such as for stateless verification of webhook IDs, as previously mentioned on Slack). Registering cookie jars instead of just tokens enables us to reuse the same pattern for securely setting other kinds of cookies as well.

### Why do we need cookie registries?
Cookie jars are stored globally (in `IGlobalCookieRegistry`) so that we can just have a single async task server-wide which garbage-collects old ones, instead of one task per-app session. `ICookieRegistry` (non-global) just exists to remove from the global registry all jars associated with an app session when it dies. In practice, of course, we're very unlikely to have a bunch of cookie jars in-flight at once, let alone per-app session. This is just protecting against extreme cases and malicious clients.

---

> Why not use AppSessionStore?

Would you like me to replace `IGlobalCookieRegistry` with just a `ConcurrentDictionary` property on `AppSessionStore`, and `ICookieRegistry` with a property on `AppSession`? I think that would work fine. (_edit: actually that doesn't work, but I am currently simplifying the cookie registry stuff in a slightly different way_)

---

> Also I think we should put everything that isn't a public API under Ivy.Core.

Sure. I will move a few things to `Ivy.Core`.

---

> Not sure I understand this PR fully.

Let me know if you have any more questions!

zachwolfe:
@nielsbosma I moved some things to `Ivy.Core` and moved all cookie registration functionality to `AppSessionStore`. The non-global `ICookieRegistry` stuff has been removed; the global timer will clean them up anyway.

Note that I had to add `AppSessionStore` as a service on the `Server` so that I could access it from `DefaultSidebarChrome` and `DefaultAuthApp`. Let me know if that is not OK with you.

nielsbosma:
@zachwolfe

So I moved all the auth files to Ivy.Auth. Makes more sense. 

We want to make it easy to create custom AuthApps. I'm not sure we should expose internals as AppSessionStore. 

I'm thinking that instead of exposing AppSessionStore how about we put the logic for SetAuthCookies and  SetAuthSessionData in IAuthService - IAuthService internally can access sessionStore and IClientProvider?

zachwolfe:
@nielsbosma
> I'm thinking that instead of exposing AppSessionStore how about we put the logic for SetAuthCookies and SetAuthSessionData in IAuthService - IAuthService internally can access sessionStore and IClientProvider?

Yes, I like that idea.

zachwolfe:
@nielsbosma I moved all the cookie-setting logic to IAuthService as `internal` methods. I also moved `TimeoutHelper` usage to `IAuthService`, so the call-sites for `IAuthService` are a lot cleaner in general now. In my testing, everything seems to still work for basic auth and OAuth.

nielsbosma:
@zachwolfe looks good! Great work!

PR #1616, opened 2025-11-26
[Charts]: show tools only on hover if enabled

ArtemLazarchuk:
### Summary
Changed the default Toolbox property value from new Toolbox() to null in all chart widgets (LineChart, BarChart, AreaChart, PieChart) to enable toolbox functionality by default.

### Changes
Widget Classes
LineChart.cs: Toolbox default changed from new Toolbox() to null

### Sample Applications
Updated sample apps to demonstrate toolbox:
LineChartApp: Remove.Toolbox() for 3 examples
BarChartApp: Remove.Toolbox() for 3 examples
AreaChartApp: Remove.Toolbox() for 3 examples
PieChartApp: Remove.Toolbox() for 3 examples

### Impact
Toolbox is disabled by default on all charts.

image
image
image
image

### DIFF
--- a/Ivy.Samples.Shared/Apps/Widgets/Charts/AreaChartApp.cs
+++ b/Ivy.Samples.Shared/Apps/Widgets/Charts/AreaChartApp.cs
@@ -44,7 +44,6 @@ public class AreaChart0View : ViewBase
                 .Dimension("Month", e => e.Month)
                 .Measure("Desktop", e => e.Sum(f => f.Desktop))
                 .Measure("Mobile", e => e.Sum(f => f.Mobile))
-                .Toolbox()
         ;
     }
 }
@@ -69,7 +68,6 @@ public class AreaChart1View : ViewBase
                 .Measure("Desktop", e => e.Sum(f => f.Desktop))
                 .Measure("Mobile", e => e.Sum(f => f.Mobile))
                 .Measure("Tablet", e => e.Sum(f => f.Tablet))
-                .Toolbox()
         ;
     }
 }
@@ -93,7 +91,6 @@ public class AreaChart2View : ViewBase
                 .Dimension("Month", e => e.Month)
                 .Measure("Revenue", e => e.Sum(f => f.Revenue))
                 .Measure("Expenses", e => e.Sum(f => f.Expenses))
-                .Toolbox()
         ;
     }
 }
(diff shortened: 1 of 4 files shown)


rorychatt:
Is there really no way to just hide the controls?

Why is the isHovered even needed?

rorychatt:
I dont also like that when you leave the chart, it resets the state

ArtemLazarchuk:
@rorychatt ECharts does not provide a built-in way to automatically hide the toolbox without manually changing the `show` property.  
That's why the `isHovered` state is used.

### What is `isHovered`?

`isHovered` indicates whether the mouse cursor is currently over the chart area:

- **true** — the cursor is hovering over the chart  
- **false** — the cursor is outside the chart

### How it works

- **On `mouseenter`:**  
  `isHovered = true` → `toolbox.show = true` → the toolbox appears

- **On `mouseleave`:**  
  `isHovered = false` → `toolbox.show = false` → the toolbox disappears

frontend/src/widgets/charts/BarChartWidget.tsx
```diff
@@ -214,13 +215,34 @@ const BarChartWidget: React.FC<BarChartWidgetProps> = ({
     ]
   );
 
+  // Show/hide toolbox on hover (directly via ECharts, no React re-renders)
+  const onChartReady = (instance: ECharts) => {
+    if (!toolbox) return;
+    const dom = instance.getDom();
+    if (!dom) return;
```
ArtemKhvorostianyi: Are we sure that dom based solution is a cleanest way to implement toolbox? Most of input widgets has toolbox implementation, maybe have a look how it is implemented there. I'm not sure that this is an issue, however

frontend/src/widgets/charts/sharedUtils.ts
```diff
@@ -331,7 +334,7 @@ export const generateEChartToolbox = (toolbox?: ToolboxProps) => {
   }
 
   return {
-    show: true,
+    show: isHovered !== false, // Show only when hovered (or when isHovered is undefined/true)
```
ArtemKhvorostianyi: Can't we just use isHovered == true instead of isHovered !== false?

rorychatt: why not just use isHovered itself. it itself returns a value.

frontend/src/widgets/charts/BarChartWidget.tsx
rorychatt: yeah this is always a bad solution

rorychatt:
@ArtemLazarchuk 

Please make this PR simple and just make it so ToolBox is not shown.

By default in all charts it should not be enabled -> FE does not render them
If user does .Toolbox() -> it is enabled -> FE reads the prop and enables it.

No need for hover/on hover.

End goal is to make that dashboards don't get the unneccessary noice.

PR #1623, opened 2025-11-28
(DemoBox): (removal): replace with Box.Plain and smart TOC

defymecobra:
- Added `Box.Plain()` extension method as a reusable preset applying DemoBox styling (border=1px, padding=16px, neutral color, top-left alignment)
- Updated `MarkdownConverter.cs` to generate `Box().Plain()` instead of `DemoBox()` for documentation examples
- Enhanced `TableOfContents.tsx` to intelligently detect and exclude headings from tabs named "Demo" instead of relying on `data-demo-box` attribute
- Removed redundant `DemoBox.cs` class and `DemoBoxWidget.tsx` component
- Removed `Ivy.DemoBox` registration from `widgetMap.ts`

### DIFF
--- a/Ivy/Widgets/Primitives/Box.cs
+++ b/Ivy/Widgets/Primitives/Box.cs
@@ -49,16 +49,6 @@ public static class BoxExtensions
 
     public static Box ContentAlign(this Box box, Align? align) => box with { ContentAlign = align };
 
-    public static Box Plain(this Box box) => box with
-    {
-        BorderThickness = new(1),
-        Padding = new(4),
-        BorderRadius = Ivy.Shared.BorderRadius.Rounded,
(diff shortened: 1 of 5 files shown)


defymecobra:
@claude review this

Ivy/Widgets/Primitives/Box.cs
rorychatt, line 1: this is only used in Docs... can we move this to another place? like into Ivy.Docs.Tools for example?

rorychatt: I am thinking about the Plain extension

defymecobra: Done

PR #1625, opened 2025-11-29
(form): FE api refactoring

defymecobra:
This pull request refactors the way icon sizing and positioning is handled in the `AsyncSelectInputWidget` component to improve code maintainability and readability. Instead of using inline conditional logic to determine class names for icon containers and icons, the code now uses dedicated variant objects for each scale.

Refactoring for maintainability:

* Added three variant objects—`asyncSelectIconContainerVariants`, `asyncSelectIconVariants`, and `asyncSelectInvalidIconVariants`—to centralize the mapping of scale values to class names for icon containers, icons, and invalid icon positioning, respectively. (`frontend/src/widgets/inputs/AsyncSelectInputWidget.tsx`)

Code simplification and consistency:

* Updated the rendering logic in `AsyncSelectInputWidget` to use the new variant objects instead of inline ternary expressions, making the code easier to read and maintain. (`frontend/src/widgets/inputs/AsyncSelectInputWidget.tsx`)

### DIFF
--- a/Ivy/Shared/Size.cs
+++ b/Ivy/Shared/Size.cs
@@ -177,52 +177,7 @@ public class SizeJsonConverter : JsonConverter<Size>
 {
     public override Size Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        var s = reader.GetString();
-        if (string.IsNullOrEmpty(s))
-        {
-            return null!;
-        }
-
-        var parts = s.Split(',');
-        var main = ParseSize(parts[0]);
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt (changes requested):
I never saw us get this exception in any of the apps. So naturally, I would say this is not the implementation we need

defymecobra:
> I never saw us get this exception in any of the apps. So naturally, I would say this is not the implementation we need

I removed it

PR #1640, opened 2025-12-01
[Kanban]: clear column highlight after drag ends

defymecobra:
## Problem

After dragging and dropping a Kanban card, the previous column remained highlighted. This happened because the local `isDragOver` state in each `KanbanColumn` component wasn't being reliably reset when the drag operation ended. The `handleDragLeave` event doesn't always fire correctly during fast mouse movements, leaving columns in a highlighted state.

## Solution

Moved highlight state management from local component state to global context:

- Added `dragOverColumn` state to `KanbanContext` to track which column is currently highlighted (stores column ID instead of boolean in each column)
- Modified `KanbanColumn` to use the global `dragOverColumn` state instead of local `isDragOver` state
- Added `setDragOverColumn(null)` to `KanbanCard.handleDragEnd` to ensure highlight is always cleared when drag operation completes

This approach guarantees that highlight state is reset when the drag ends, regardless of whether `handleDragLeave` fired correctly. The centralized state management also prevents synchronization issues between columns.

### DIFF
--- a/frontend/src/components/ui/shadcn-io/kanban.tsx
+++ b/frontend/src/components/ui/shadcn-io/kanban.tsx
@@ -33,6 +33,8 @@ interface KanbanContextType {
   columns: Column[];
   draggedCardColumn: string | null;
   setDraggedCardColumn: (column: string | null) => void;
+  dragOverColumn: string | null;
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

PR #1660, opened 2025-12-02
[Grid]: improve text contrast with opacity in dark mode

defymecobra:
image
image

In dark mode, text was unreadable on boxes with opacity because `color-mix` always mixed towards white, making backgrounds lighter while text remained white. Added CSS variable `--opacity-mix-color` that switches between `white` (light mode) and `black` (dark mode), ensuring text stays readable in both themes.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -116,16 +116,12 @@
     --shadow-xl:
       0 1px 3px 0px hsl(0 0% 0% / 0.1), 0 8px 10px -1px hsl(0 0% 0% / 0.1);
     --shadow-2xl: 0 1px 3px 0px hsl(0 0% 0% / 0.25);
-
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

rorychatt:
can we instead use somehow the "transparent" or "background" color var? 
bg might be of different colors if its in root of app, inside a card and so on

dcrjodle:
use the background variable instead of creating a new one :)

defymecobra:
Removed all custom opacity mix color variables from CSS. Now color-mix for opacity always blends with the current --background variable, providing context-aware transparency without extra global or theme-specific variables. The appearance is the same as in the screenshots above.

defymecobra:
@claude review this

PR #1680, opened 2025-12-03
[Details]: implement scale api 

ArtemLazarchuk:
image

## Summary

Implemented full scale API support for Detail components with the ability to choose size (Small, Medium, Large).

### Backend (C#):

- Added scaling methods to `DetailsBuilder.cs` (Small(), Medium(), Large())
- Updated `DetailsApp.cs` with demonstration of all component sizes

### Frontend (React/TypeScript):

- Created new UI component detail.tsx with size variant support
- Implemented `DetailContext` for size management through React Context
- Added detail-variants.ts with style definitions for each size
- Created `useDetailScale` hook for easy scale access
- Updated `DetailWidget.tsx `and `DetailsWidget.tsx` to integrate new functionality

### DIFF
--- a/frontend/src/components/ui/detail/detail-variants.ts
+++ b/frontend/src/components/ui/detail/detail-variants.ts
@@ -21,9 +21,9 @@ export const detailLabelSizeVariants = cva(
 export const detailValueSizeVariants = cva('align-middle min-w-0', {
   variants: {
     scale: {
-      Small: 'p-2 pl-1 text-xs',
-      Medium: 'p-3 pl-2 text-sm',
-      Large: 'p-4 pl-3 text-base',
+      Small: 'p-2 pl-1 pr-0 text-xs',
+      Medium: 'p-3 pl-2 pr-0 text-sm',
+      Large: 'p-4 pl-3 pr-0 text-base',
     },
   },
   defaultVariants: {


Ivy/Views/Builders/DetailsBuilder.cs
```diff
@@ -91,6 +93,24 @@ public DetailsBuilder<TModel> RemoveEmpty()
         return this;
     }
 
+    public DetailsBuilder<TModel> Large()
```
rorychatt, line 96: this is already in widget base.

Ivy/Views/Builders/DetailsBuilder.cs
```diff
@@ -47,6 +48,7 @@ private class Item(string label, IBuilder<TModel> builder, FieldInfo fieldInfo,
     private bool _removeEmpty;
     private readonly Dictionary<string, Item> _items;
     private readonly TModel _model;
+    private Scale _scale = Scale.Medium;
```
rorychatt, line 51: this is already in widget base

rorychatt:
What happens when this is inside a card that has a scale?

Did this not get styled before by the parent? (before your commits, try this out)

aka Card.Large(content: object.ToDetails())

ArtemLazarchuk:
@rorychatt 
### In this PR:
if you use a format like Card.Large(content: object.ToDetails()), everything will render correctly — the text size for ToDetails() will adjust based on the container size. The spacing does not change yet, but this can be fixed. For example, with a Card.Large card, ToDetails().Large() will be applied implicitly and automatically inside.

image
image

### Before this PR:
if you try to repeat the same action, you get the following result: the label doesn’t change its text size, and the spacing also doesn’t change, but the values do change their size.

image
image

### Summary:
the size for details can now be applied both explicitly and implicitly.

ArtemLazarchuk:
> see my feedback.
> 
> Scale api is already implemented in widget base.

### The Issue
DetailsBuilder inherits from `ViewBase`, not `WidgetBase<T>`. This means the generic extension methods from `WidgetBaseExtensions` cannot be applied to `DetailsBuilder` because it doesn't satisfy the where `T : WidgetBase<T>` constraint.

### The Pattern
The methods in `DetailsBuilder` are wrapper methods that store the scale and apply it when building the `Details` widget.
This is the same pattern used in:
`FormBuilder.cs`
`TableBuilder.cs`

ArtemLazarchuk:
@rorychatt when the details are inside the card, the font visually changes, but the size shown in the inspector does not. This happens because the cards use a selector like `[&_*]:text-...` It applies only the font-size to all nested elements, but doesn’t affect their spacing, so those remain unchanged.

PR #1688, opened 2025-12-04
(tooltip): multiline text in tooltip if it's long

ArtemKhvorostianyi:
Had an issue:
Screenshot 2025-12-04 at 09 40 23
Screenshot 2025-12-04 at 09 40 56

Added max-w-sm param for tooltips, solved
Screenshot 2025-12-04 at 09 38 18
Screenshot 2025-12-04 at 09 38 24

Handled very long string without spaces (for example, urls) with adding break-all param
Screenshot 2025-12-04 at 09 47 30

These changes should fix all possible issues linked to cropped tooltips


ArtemKhvorostianyi:
@claude review this

rorychatt:
(tooltip): limit height too

While whe made it so new lines can appear and set max width of a tooltip to `sm` (which might be too small) we still need to limit the amount of lines that can be created

PR #1689, opened 2025-12-04
(stepper): implement docs

ArtemKhvorostianyi:
Screenshot 2025-12-04 at 10 42 30
Screenshot 2025-12-04 at 10 42 41


rorychatt:
API section has this info. No?

image

ArtemKhvorostianyi:
> API section has this info. No?
> 
> image

Yes, but basic usage contains not clear (null, 1...) parametres, I think it will be better to separatly explain what it means right under basic usage example

Issue #1667, opened 2025-12-03
(tabs): content variant underline is over... content


rorychatt:
### Description

<img width="935" height="510" alt="Image" src="https://github.com/user-attachments/assets/ce57dbd1-b0e2-465c-9b6e-a3cc883e5c10" />

### Code Samples

```csharp

```

PR #1692, opened 2025-12-04
(chore): Added Form apps into one menu folder. Removed FromSizeApp

KaiserReich95:
After changes:

https://github.com/user-attachments/assets/f8268339-34a7-4f29-a673-a76b4d998904

Removed FromSizeApp since that functionallity already exists in From app under "Form Size Demonstration".

### DIFF
--- a/Ivy.Test/FormValidationAttributesTest.cs
+++ b/Ivy.Test/FormValidationAttributesTest.cs
@@ -67,7 +67,7 @@ public class ValidationAttributesTestModel
 /// <summary>
 /// Model to test Display attribute properties.
 /// </summary>
-public class DisplayAttributeTestModel
+public class FormDisplayAttributeTestModel
 {
     [Display(Name = "Custom Name")]
     public string NameField { get; set; } = string.Empty;
@@ -471,7 +471,7 @@ public class FormValidationAttributesTest
     [Fact]
     public void Display_ShouldExtractName()
     {
-        var displayInfo = GetDisplayInfo<DisplayAttributeTestModel>(nameof(DisplayAttributeTestModel.NameField));
+        var displayInfo = GetDisplayInfo<FormDisplayAttributeTestModel>(nameof(FormDisplayAttributeTestModel.NameField));
 
         Assert.Equal("Custom Name", displayInfo.Name);
     }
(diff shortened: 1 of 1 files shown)


rorychatt:
I have a better suggestion for implementation.

Instead of having a folder in sidebar, let's experiment with giving this a similar look like in `ivy-samples/kanban`, where we have inside the app tabs variant content.

rorychatt (changes requested):
see comment above

rorychatt:
also for form scaffolding, we might need to make sure that all the supported descriptors (tags) are actually working. we have a list of what works somewhere in the code... I think the docs forms example had them

KaiserReich95:
I used the same implementation as the kanban app which seems to have issues with the line under the different tabs. It is sometimes hidden under some cards and there is a transparent gap between the tab title and the line. Is that something that should be fixed in a separate issue?

ivy-kanban-issue1
ivy-kanban-issue2

rorychatt:
@KaiserReich95 we already have an issue about this

https://github.com/Ivy-Interactive/Ivy-Framework/issues/1667

Introduced via tabs refactor, I think.

KaiserReich95:
The difference with kanban is that I divided up the forms into their different variations. The code for Kanban is quite large and maybe could be broken up aswell into the different variants? I still have the same problem as the #1667 issue in this aswell. 

Added unit tests (FormValidationAttributesTest.cs) to verify all form validation attributes listed in the Forms documentation work correctly.

A minor change was made when the tests failed here in github: DisplayAttributeTestModel renamed to FormDisplayAttributeTestModel

Also wrapped the horizontal layout in a Layout.Vertical() to enable scrolling:

The change was in FormValidationExample.cs: 

Before:
```
return Layout.Horizontal()
    | new Card(form)...
```

After:
```
return Layout.Vertical()
    | (Layout.Horizontal()
        | new Card(form)...
```

Forms with tabs:
https://github.com/user-attachments/assets/dc97bbcc-fdcb-4fde-a412-31a89241ba9a

dcrjodle (approved):
I think it looks pretty good. 
nitpick: I would remove the file summary comments since we don't use them elsewhere
@rorychatt what's the difference between `Ivy.Samples.Shared/Apps/Tests` and `Ivy.Test` location for test. Doesn't seem to be a clear structure

rorychatt:
@KaiserReich95 @dcrjodle we use Ivy.Samples.Shared/Test for "e2e" testing (currently just developers visually confirming) some functionality of some apps.
Most likely we will have a separate Ivy.Samples.Tests for this in the future.
Ivy.Test is a xUnit (unit) test project

Issue #1592, opened 2025-11-25
(text-input): file structure refactor


rorychatt:
maybe for hooks we should have a separate file.

In this case, it makes sense to create a folder called `TextInputWidget`, in there have the `index.ts`, `TextInputWidget.tsx` and the hook file

_Originally posted by @rorychatt in https://github.com/Ivy-Interactive/Ivy-Framework/pull/1518#discussion_r2556130906_

KaiserReich95:
Should we also have different files for the variants? When I was working with affixes it was tiresome to go though all the code for all the variants to find the one to change.

<img width="271" height="300" alt="Image" src="https://github.com/user-attachments/assets/d9ec6edd-4d94-4752-85e5-9e1f058288d8" />

rorychatt:
@KaiserReich95 I don't have a preference. See how it is done for other structured components.

If you do separate files for each variant, then yes it is a good pattern where all variants are in `.widget/variants` module  with named exports through `index.js`

PR #1696, opened 2025-12-04
(chore): Refactor of TextInput

KaiserReich95:
Based on this issue #1592 

Still works after refactor:

https://github.com/user-attachments/assets/86d700b1-71e0-485e-8bd0-e545560476da

We structured the TextInput to the pattern used by dataTables and layout/tabs but we added variants instead of components to reflect the types of files we have there. DataTables uses one file for all its hooks while Tabs uses multiple. I think it was better to follow the DataTables pattern since some hook files become quite small otherwise.


frontend/src/widgets/inputs/TextInputWidget/types.ts
```diff
@@ -0,0 +1,55 @@
...
+  suffix?: PrefixSuffix;
+  maxLength?: number;
+  'data-testid'?: string;
+}
+
+/**
+ * Renders either text or icon for prefix/suffix display.
+ * Uses discriminated union type to ensure only one type can be set.
+ */
+export const renderPrefixSuffix = (
```
dcrjodle: create utils folder and put this function there instead, it doesn't belong to types.ts as it is not an interface or a type

frontend/src/widgets/inputs/TextInputWidget/TextInputWidget.tsx
```diff
@@ -0,0 +1,192 @@
...
+  maxLength,
+  'data-testid': dataTestId,
+}) => {
+  const eventHandler = useEventHandler();
+  const [localValue, setLocalValue] = useState(value);
+  const [isFocused, setIsFocused] = useState(false);
+  const inputRef = useRef<HTMLInputElement | HTMLTextAreaElement | null>(null);
+
+  // Update local value when server value changes and control is not focused
+  useEffect(() => {
```
dcrjodle: can we get the two useEffects out of the main component and in to the hook file

frontend/src/widgets/inputs/TextInputWidget/TextInputWidget.tsx
```diff
@@ -0,0 +1,192 @@
...
+    setIsFocused(false);
+    if (events.includes('OnBlur')) eventHandler('OnBlur', id, []);
+  }, [eventHandler, id, events]);
+
+  const handleFocus = useCallback(() => {
+    setIsFocused(true);
+    if (events.includes('OnFocus')) eventHandler('OnFocus', id, []);
+  }, [eventHandler, id, events]);
+
+  const commonProps = useMemo(
```
dcrjodle, line 65: can you explain why this is needed?

frontend/src/widgets/inputs/TextInputWidget/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type PrefixSuffix =
+  | { type: 'text'; value: string }
+  | { type: 'icon'; value: string };
+
+export interface TextInputWidgetProps {
+  id: string;
+  placeholder?: string;
+  value?: string;
+  variant:
```
dcrjodle: in this file, create an enum called variants that we put all these keys in. then use that enum in TextInputWidgetProps and in TextInputWidget.tsx where we have loose string keys

frontend/src/widgets/inputs/TextInputWidget/TextInputWidget.tsx
```diff
@@ -0,0 +1,192 @@
...
+          inputRef={inputRef}
+          isFocused={isFocused}
+          scale={scale}
+        />
+      );
+    default:
+      return (
+        <DefaultVariant
+          type={
+            variant.toLowerCase() as Lowercase<TextInputWidgetProps['variant']>
```
dcrjodle: use enum, see prev comment

frontend/src/widgets/inputs/TextInputWidget/hooks/useTextInput.ts
```diff
@@ -0,0 +1,171 @@
+import React, { useRef, useEffect, useCallback } from 'react';
+
+// Utility to detect Mac platform
```
dcrjodle: can you put all the utility functions in a utils folder. the utility file name should be what it contains, not just `util.ts`

frontend/src/widgets/inputs/TextInputWidget/TextInputWidget.tsx
KaiserReich95: From claude:

Let me explain what it does: useMemo is a React hook that caches the result of a calculation between re-renders. Here it creates a commonProps object that's passed to all the variant components (DefaultVariant, PasswordVariant, etc.). What it does:

- Creates an object with all the shared props that every variant needs
- Only recalculates this object when one of the dependencies in the array changes
- If none of the dependencies change, React reuses the cached object from the previous render

Why it matters: Without useMemo, a new commonProps object would be created on every render. Since objects are compared by reference in JavaScript, this would cause the child variant components to re-render even if nothing actually changed. The dependency array lists all values used inside the memoized function. React will only recompute commonProps when one of these values changes:

PR #1698, opened 2025-12-04
[Docs]: fix dialog form width in few app examples in Forms

defymecobra:
The issue arose because in some sample applications on the Forms page in Docs, the width `Size.Units(500)` was set for dialog forms, which actually looks like 2000 pixels.
To solve this, I set the value in units equal to 500 pixels, i.e. `Size.Units(125)`.

Now it looks as expected:
image
image
image


defymecobra:
@claude review this

Issue #1672, opened 2025-12-03
(embed): github codespaces on blur gets green border


rorychatt:
### Description

this should not be green IMO

<img width="306" height="1011" alt="Image" src="https://github.com/user-attachments/assets/e60d02bb-04ac-4976-9559-f040dee614ed" />

### Code Samples

```csharp

```

KaiserReich95:
Where can I find the button "Open in Codespaces"? Cant seem to find it in Ivy.Samples or Ivy.Docs. Is it in some other project?

rorychatt:
in any of the deployed package demos in Ivy-Examples.

it's an embed element

PR #1699, opened 2025-12-04
(fix): Removed focus:ring (green color) from git codespaces button

KaiserReich95:
Relates to this issue #1672 

I removed the focus:ring all together, it seems to only appear after the user has clicked the button and doesnt provide much for user experience.

Before:
https://github.com/user-attachments/assets/168d478d-418f-4421-b9e7-b1da51e6467d

After:
https://github.com/user-attachments/assets/978fcd36-e1aa-4295-91c6-efc2bb0fe143


rorychatt (approved):

rorychatt:
LGTM

@dcrjodle merge if you agree with this change

dcrjodle:
can you add a small hover effect to the button that basically reduces the background color opacity with 20-30%, or something similar

PR #1702, opened 2025-12-05
[tabs]: move underline inside container bounds

defymecobra:
The underline indicator in Content variant was positioned outside the container (bottom: -6px), causing it to be overlapped.

Changes:
- Added pb-[6px] padding to tab container
- Changed underline position from bottom-[-6px] to bottom-0

image


rorychatt:
this will have side-effects in docs. I will need to check it out there

Issue #1686, opened 2025-12-04
(docs): description should be more consistent/clear


joshuauaua:
<img width="1271" height="874" alt="Image" src="https://github.com/user-attachments/assets/1a1513f5-9316-4957-a6c2-1c560b1bdad3" />

The descriptions under the docs example (JSON) could be more cohesive and consistent, especially the text highlighted in screenshot.  

It would be helpful to align them in voice/phrasing, structure, and clarity.

joshuauaua:
Happy to take this on

rorychatt:
@joshuauaua sure give this a try with a better phrasing/format

PR #1717, opened 2025-12-08
(docs): rewrote description for better consistency

joshuauaua:
Screenshot 2025-12-08 at 09 55 20

Issue #1686


rorychatt:
@joshuauaua do the change as we discussed with a big cleanup

PR #1721, opened 2025-12-08
(chore): Ability to set new sidebar width

### LINKED ISSUE #1557: (sidebar): allow adjusting width
### Description

_No response_

### Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Impact

_No response_

### Code Examples

```csharp

```

### Additional Context

_No response_

KaiserReich95:
Closes #1557 

The user creating a new application can set the sidebar to a new width inside of the DefaultSidebarChrome.cs file.

It was difficult to understand the code correctly but I did my best to make it as efficent as possible. Some explainations (comments) are added to the file changes here in Github to explain some of the changes made.

This is just an extreme example when we make the sidebar very small:

The text was at first centerd:
ivy-sidebar-tabs-no-centre

So i left-aligned the text for it to look better:
ivy-sidebar-small

To change the size of the sidebar one has to configure the chrome settings for a project. Here is an example of that in Ivy.Samples project:

ivy-change-sidebar-width-example

Also test for refactor of sidebar toggle button positioning:

https://github.com/user-attachments/assets/9b0a0949-dc70-4cbf-bea6-0bb099f856a3

### DIFF
--- a/frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
+++ b/frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
@@ -64,8 +64,7 @@ export const SidebarLayoutWidget: React.FC<SidebarLayoutWidgetProps> = ({
   width,
 }) => {
   // Get sidebar width from the width prop (default set in backend)
-  const sidebarWidthStyles = getWidth(width);
-  const sidebarWidth = sidebarWidthStyles.width as string;
+  const sidebarWidth = getWidth(width).width as string;
   // Initialize sidebar state based on current window width (only for main app sidebar)
   const getInitialSidebarState = () => {
     if (!mainAppSidebar) return true;


frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -59,7 +61,11 @@ export const SidebarLayoutWidget: React.FC<SidebarLayoutWidgetProps> = ({
   autoCollapseThreshold = 768,
   mainAppSidebar = false,
   mainContentPadding,
+  width,
 }) => {
+  // Get sidebar width from the width prop (default set in backend)
+  const sidebarWidthStyles = getWidth(width);
+  const sidebarWidth = sidebarWidthStyles.width as string;
```
KaiserReich95: To execute the code further we need the sidebarWidthStyles.width as a string

frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -146,15 +152,16 @@ export const SidebarLayoutWidget: React.FC<SidebarLayoutWidgetProps> = ({
       ref={containerRef}
       className="grid h-full w-full remove-parent-padding"
       style={{
-        gridTemplateColumns: isSidebarOpen ? '16rem 1fr' : '0 1fr',
+        gridTemplateColumns: isSidebarOpen ? `${sidebarWidth} 1fr` : '0 1fr',
```
KaiserReich95, line 154: Here we need to use sidebarWidth as a string to extract the value. The gridTemplateColumns divides the whole screen into sidebar to the left and the main content to the right. I think it's still the most efficent way of sizing the sidebar and the main content instead of main content resizing itself to fill the space to the right of the sidebar. The blade widget has an adjusted size to fit between other components but the sidebar has one size set at the start so we can just divide up the space between sidebar and the main content. Write a comment if you guys have other ideas.

frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -177,36 +184,28 @@ export const SidebarLayoutWidget: React.FC<SidebarLayoutWidgetProps> = ({
...
+            style={{ marginTop: '3px' }}
+            aria-label={isSidebarOpen ? 'Close sidebar' : 'Open sidebar'}
+          >
+            {isSidebarOpen ? (
+              <PanelLeftClose className="h-4 w-4" />
+            ) : (
+              <PanelLeftOpen className="h-4 w-4" />
+            )}
+          </button>
+        )}
```
KaiserReich95, line 207: We moved the toogle button for the sidebar into main content instead. Then we dont have to use this calculation:
left: isSidebarOpen ? 'calc(16rem + 4px)' : '4px'. to position the button.

frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -59,7 +61,10 @@ export const SidebarLayoutWidget: React.FC<SidebarLayoutWidgetProps> = ({
   autoCollapseThreshold = 768,
   mainAppSidebar = false,
   mainContentPadding,
+  width,
 }) => {
+  // Get sidebar width from the width prop (default set in backend)
+  const sidebarWidth = getWidth(width).width as string;
```
KaiserReich95, line 67: Most widgets spread the whole object into styles: style={{ ...getWidth(width) }} The sidebar only needs the width value as a raw string ('16rem') to use in the CSS Grid template: `${sidebarWidth} 1fr`. You can't put an object there - it needs a string for the code later down the line.

It looks stupid and I'm not sure if its the best solution here since I dont think we need a min and max size for the sidebar that is required for other widgets. The user creating an app with the framework can easily discover if the sidebar width is to large or small.

frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -465,7 +463,7 @@ export const SidebarMenuWidget: React.FC<SidebarMenuWidgetProps> = ({
         return (
           <li key={item.tag}>
             <button
-              className={`flex w-full items-center gap-2 rounded-lg p-2 text-sm hover:bg-accent hover:text-accent-foreground cursor-pointer h-8 ${
+              className={`flex w-full items-center gap-2 rounded-lg p-2 text-sm hover:bg-accent hover:text-accent-foreground cursor-pointer h-8 text-left ${
```
rorychatt: we have Text.Align in the backend (or something like that)
-> another customization feature for the sidebar perhaps?

Could you share more screenshots of what sidebar looks like when it is super narrow or wide?

PR #1724, opened 2025-12-09
(security): fix Incomplete string escaping or encoding

ArtemKhvorostianyi:
Screenshot 2025-12-09 at 03 49 18
Wrote some tests about it (and deleted them, feauture is too small to create external file for testing)
Screenshot 2025-12-09 at 03 52 00

Tested out copying of markdown doc page and extracting of tables. For example: 
Screenshot 2025-12-09 at 03 55 09

Screenshot 2025-12-09 at 03 54 49

Everything looks correct — the change doesn’t affect table extraction, and the behavior remains the same.
This update is minor and safe.

Screenshot 2025-12-09 at 03 56 45


ArtemKhvorostianyi:
@claude review this

ArtemKhvorostianyi:
Screenshot 2025-12-09 at 06 20 48

1. We don’t have similar issues in other sections of the codebase, checked. 
2. I tested this functionality locally (screenshot attached). I didn’t commit the test because it would be redundant for such a small addition.

PR #1758, opened 2025-12-09
(fix): Fixed inconsistencies in the Forms app (Ivy.Samples)

### LINKED ISSUE #1706: (select): large has weird styling
### Description

Maybe other scales too IDK (see full green toggle field)

Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

KaiserReich95:
Closes #1706 

The inconsistency in the issue was the boolean field that was to wide:

Before:
ivy-form-org

After:
ivy-form-bool1

Then I looked for more inconsistencies and found more of them in DateTime:

- Icons for Date Field and Date Time Field did not get larger by size.
- Time Field had different padding by size.
- Time Field input field did not get larger by size.

ivy-form-main

So I fixed that aswell:
ivy-form-fixes

The changes to the Time Field does require changes that are unique it to work. Date Field and Date Time Field uses a button as a container which brings with it styling changes, Time Field doesnt. We could add a variant of input field called Time Field but we only use it here so I dont know if its needed.

I was also thinking about the input times and dates being higher or lower then the center of the icons but it seems to be hard to solve in a clean way since C# tries to center itself but it cant do that in a perfect way everytime. Trying to center the text myself makes many sections of the code very specific and abritrary.

### DIFF
--- a/frontend/src/components/ui/input/date-time-input-variants.ts
+++ b/frontend/src/components/ui/input/date-time-input-variants.ts
@@ -19,9 +19,9 @@ export const dateTimeInputVariants = cva(
 export const dateTimeInputIconVariants = cva('', {
   variants: {
     scale: {
-      Small: 'h-3 w-3',
-      Medium: 'h-4 w-4',
-      Large: 'h-5 w-5',
+      Small: '!size-3',
+      Medium: '!size-4',
+      Large: '!size-5',
     },
   },
   defaultVariants: {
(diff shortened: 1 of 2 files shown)


dcrjodle:
@rorychatt are icons suppose to scale? I don't think icons usually scale in other widgets

rorychatt:
@dcrjodle yes icons are supposed to scale. SVG-s have no troubles with that

PR #1802, opened 2025-12-11
Refactored Affixes

### LINKED ISSUE #1586: (prefix-suffix): big refactor
### Description

Current solution in the backend does not suit us well.

We need to at least remove everything regarding custom json converter. And not register that in abstract widget base.

Also we are not so happy about the API (PreffixSuffix), as it looks confusing.

We talked about the API, and figured out that simpler solution with a record type that has nullable icon or string is just better.

And if both are given, just show only one in the frontend.

### Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Impact

_No response_

### Code Examples

```csharp

```

### Additional Context

_No response_

KaiserReich95:
Closes #1586 

Changes made:

-PrefixSuffix naming changed to Affix instead.
-Removed PrefixSuffixJsonConverterFactory()
-Added record for Affixes


Ivy/Widgets/Inputs/TextInput.cs
```diff
@@ -1,16 +1,23 @@
...
 
 // ReSharper disable once CheckNamespace
 namespace Ivy;
 
+public record PrefixSuffix
+{
+    public Icons? Icon { get; init; }
+    public string? Text { get; init; }
+
+    public static PrefixSuffix ToIcon(Icons icon) => new() { Icon = icon };
```
rorychatt: this should just be Icon, and Text.

Also, let's bring back the Affix since it was a better name (discussed with Niels)

rorychatt: ^ @KaiserReich95 sorry about that, a bit of disinformation happened

KaiserReich95:
Fixed it

rorychatt:
pipeline is red

KaiserReich95:
It seems to be problem with setting the name to Icon and Text since Affix record already has properties named Icon and Text. They have to be renamed to something else. Any suggestion?

KaiserReich95:
@rorychatt?

rorychatt:
@KaiserReich95 I am aware. we are working on something else right now, will get back to you when possible

Ivy/Widgets/Inputs/TextInput.cs
```diff
@@ -1,16 +1,23 @@
...
 
 // ReSharper disable once CheckNamespace
 namespace Ivy;
 
+public record Affix
+{
+    public Icons? Icon { get; init; }
+    public string? Text { get; init; }
+
+    public static Affix Icon(Icons icon) => new() { Icon = icon };
```
rorychatt: try with an extension method. Or simply remove this

KaiserReich95:
How about this?

dcrjodle (approved):

nielsbosma (approved):

PR #1803, opened 2025-12-11
(chore): Readded Form app with scaffolding after compile errors

### LINKED ISSUE #1593: (codex): forms grouping in samples
Image

I am loosing the point of having so many `form` related files -> we should merge stuff together maybe, or have them be in their own folder

_Originally posted by @rorychatt in https://github.com/Ivy-Interactive/Ivy-Framework/pull/1518#discussion_r2556158517_

Also maybe hide this too?

Image

KaiserReich95:
Closes #1593 

Compile errors occured when merging the new form app (with tab scaffolding). Probably because the old PR was dormant for to long and main was updated during that time. The Form Scaffolding tab had code example blocks that used example code that was removed between the last commit and the merge into main. It least I think that was the issue. I removed xml comments aswell.

Video for test (no need to watch it, just shows I did some tests):
https://github.com/user-attachments/assets/c2a71818-84d2-41e4-8828-95cea41dc332

Discovered that the invalid input still seem to have the old issue of being highlighted with grey and red on the outline. Maybe there is a issue on this problem already just wanted to point it out:
image

### DIFF
--- a/Ivy.Samples.Shared/Apps/Concepts/FormApp.cs
+++ b/Ivy.Samples.Shared/Apps/Concepts/FormApp.cs
@@ -6,8 +6,6 @@ using Ivy.Views.Forms;
 
 namespace Ivy.Samples.Shared.Apps.Concepts;
 
-#region Enums
-
 public enum Gender
 {
     Male,
@@ -55,10 +53,6 @@ public enum ViewState
     Success,
     Error
 }
-
-#endregion
-
-#region Models
 
 public record AppSpec(string Name, string Description);
 
@@ -135,10 +129,6 @@ public record UserModel(
     List<Fruits> FavoriteFruits = null!
 );
 
-#endregion
-
-#region Scaffolding Models
-
 public class DisplayExample
 {
     [Display(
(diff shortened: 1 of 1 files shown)


Ivy.Samples.Shared/Apps/Concepts/FormApp.cs
```diff
@@ -1,22 +1,27 @@
 using System.Collections.Immutable;
+using System.ComponentModel.DataAnnotations;
 using Ivy.Shared;
 using Ivy.Views.Builders;
 using Ivy.Views.Forms;
 
 namespace Ivy.Samples.Shared.Apps.Concepts;
 
+#region Enums
```
rorychatt: let's not use regions so that other ai agents don't pick this as pattern.

`#region` was a failure as concept and historically lead to unmaintainable codebases

rorychatt (changes requested):
LGTM, just 1 minor change needed

Ivy.Samples.Shared/Apps/Concepts/FormApp.cs
KaiserReich95: Removed #regions

PR #1804, opened 2025-12-11
[FileInput]: enable tooltip for invalid icon

defymecobra:
## Problem

The validation icon tooltip in FileInput component was not working when hovering over the red invalid icon. The parent wrapper div had `pointer-events-none` CSS class which blocked all pointer interactions, preventing the tooltip from being triggered.

## Solution

Added an inner wrapper with `pointer-events-auto` around the InvalidIcon component to restore pointer events specifically for the icon. This allows the tooltip to work while keeping the surrounding area click-through for the file drop zone functionality.

How it looks now:
image

### DIFF
--- a/frontend/src/widgets/inputs/FileInputWidget.tsx
+++ b/frontend/src/widgets/inputs/FileInputWidget.tsx
@@ -381,9 +381,7 @@ export const FileInputWidget: React.FC<FileInputWidgetProps> = ({
       {/* Invalid icon in top right corner - only for required field validation */}
       {invalid && (
         <div className="absolute top-2 right-2 z-20 pointer-events-none">
-          <div className="pointer-events-auto">
-            <InvalidIcon message={invalid} />
-          </div>
+          <InvalidIcon message={invalid} />
         </div>
       )}
       <div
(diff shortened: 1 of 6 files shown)


frontend/src/widgets/inputs/FileInputWidget.tsx
```diff
@@ -381,7 +381,9 @@ export const FileInputWidget: React.FC<FileInputWidgetProps> = ({
       {/* Invalid icon in top right corner - only for required field validation */}
       {invalid && (
         <div className="absolute top-2 right-2 z-20 pointer-events-none">
-          <InvalidIcon message={invalid} />
+          <div className="pointer-events-auto">
+            <InvalidIcon message={invalid} />
```
rorychatt: why not just make it so all invalid icons have the pointer-events-auto instead?

so we don't need to do this custom div.

rorychatt (changes requested):
see my comment. let me know if it is valid or not

defymecobra:
## Changes

Moved `pointer-events-auto` logic directly into the `InvalidIcon` component instead of requiring wrapper divs in every usage.

## Benefits

- Removed 5 wrapper divs across FileInputWidget and TextInput variants
- Simplified code and improved maintainability
- InvalidIcon now works automatically regardless of parent's pointer-events settings

defymecobra:
@claude review this

PR #1807, opened 2025-12-12
(fix): Added width inline style for number input

### LINKED ISSUE #1800: (numberInput): setting width doesn't work
### Description

There is a width setting in the documentation, but the width is not set
Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

KaiserReich95:
Closes #1800 

The user of the framework can now create number input with different widths simular to how Text input does it.

ivy-number-input-width

### DIFF
--- a/frontend/src/widgets/inputs/NumberInputWidget.tsx
+++ b/frontend/src/widgets/inputs/NumberInputWidget.tsx
@@ -3,7 +3,7 @@ import { useEventHandler, EventHandler } from '@/components/event-handler';
 import NumberInput from '@/components/NumberInput';
 import { Slider } from '@/components/ui/slider';
 import { cn } from '@/lib/utils';
-import { inputStyles } from '@/lib/styles';
+import { inputStyles, getWidth } from '@/lib/styles';
 import { InvalidIcon } from '@/components/InvalidIcon';
 import { X } from 'lucide-react';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


rorychatt:
@ArtemKhvorostianyi test this in different existing apps

ArtemKhvorostianyi:
> @ArtemKhvorostianyi test this in different existing apps

We actually haven't any number input with used custom width in existing apps in samples/docs. 

I tested out functionality, so we have this pattern to set width to number input:

- for number input returning statements

```
var value = UseState(0);
        return new NumberInput<double>(value)
                     .Min(-10)
                     .Max(10)
                     .Width(Size.Fraction(0.5f));
```

- for vars with .ToNumberInput() and .WithField()

 ```
var num = UseState(3.14);

        return num
            .ToNumberInput()
            .FormatStyle(NumberFormatStyle.Decimal)
            .Width(Size.Fraction(0.5f))
            .WithField()
            .Label("Number");
```

Works fine. 
Should I update docs with using these patterns of setting custom width in this branch or create new one?

rorychatt:
No need to explain users how to set a property on a widget. Size is referenced in docs in API section.

PR #1809, opened 2025-12-12
[App]: StrictMode causes page re-render while dragging and dropping tab

defymecobra:
## Problem

Tab content re-renders when dragging and dropping tabs in development mode (`npm run dev`, localhost:5173). This issue **does not occur** in production builds or on the backend server (localhost:5010).

### Root Cause

React's `StrictMode` in development:
1. Intentionally unmounts/remounts components to detect bugs
2. This disconnects the WebSocket connection
3. On reconnection, backend sends **REFRESH** (full tree) instead of **UPDATE** (patch)
4. Full tree replacement causes tab content to re-render

### Console logging

Tested dragging and dropping "Hello" tab.

#### With StrictMode (active tab):
image

#### With StrictMode (inactive tab) (you can see confetti in upper left corner):
image 

#### Without StrictMode (active tab):
image

#### Without StrictMode (inactive tab):
image

### Recommendation: No action required

This issue should be **closed as "By Design"** or **"Won't Fix"** for the following reasons:

**1. Development-Only Issue**
- StrictMode only runs in development mode
- Production builds are unaffected
- End users never experience this

**2. Not A Bug**
- StrictMode's behavior is intentional
- Component logic is correct
- No state management issues
- This is expected behavior for WebSocket connections in StrictMode

**3. Alternatives Are Not Worth It**
- Disabling StrictMode: Lose valuable development checks
- Deep memoization: Adds complexity for a dev-only issue
- Keep as-is: Document as known behavior

**4. Widely Accepted Pattern**
- Many WebSocket/SSE applications have this behavior
- React team acknowledges this trade-off
- Not indicative of production problems

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -1,4 +1,3 @@
-import { StrictMode } from 'react';
 import { createRoot } from 'react-dom/client';
 import './index.css';
 import { App } from './components/App';
@@ -16,8 +15,4 @@ if (!root) {
   (window as WindowWithReactRoot).__reactRoot = root;
 }
 
-root.render(
-  <StrictMode>
-    <App />
-  </StrictMode>
-);
+root.render(<App />);


defymecobra:
@claude review this

rorychatt:
@dcrjodle @nielsbosma 

cc
>A long running tab-rerender issue that happens only in strict mode.
A lot of debugging later, seems as if it is indeed intended behavior for remounting ws connections.
defymecobra wrote a good draft that explains the problem, and I need help.
My suggestion is that we just ignore this problem. First thoughts were that issue is in state handling, but we can't find it in code. Deep memoization might also be a bad solution. But let me know what you think

PR #1814, opened 2025-12-14
[tabs]: cache widget content to prevent remount on parent re-render

defymecobra:
## Problem
When the backend sends any state update via SignalR (e.g., switching tabs in one section or opening a Sheet), the entire widget tree is re-rendered. The `renderWidgetTree` function creates new React element instances on every call, even when the underlying data hasn't changed. Although these elements have stable `key` props, React still unmounts and remounts child components because the element references themselves are new objects. This causes heavy widgets like `DataTable` to lose their internal state (scroll position, sorting, filters) whenever an unrelated UI interaction triggers a backend update.

## Solution
Introduced a `useWidgetCache` hook that caches React elements by their unique id on first render. The cache uses `useState` with useLayoutEffect to comply with React's rendering rules — no refs are accessed during render, and the Map is updated synchronously before paint. When parent components re-render with new widget instances, the cached (original) elements are returned instead, preserving component identity and preventing unnecessary remounts. Stale entries are automatically removed when widgets are deleted. This approach is localized to the `Variants.tsx` file, requires minimal code changes, and doesn't affect other parts of the system.


defymecobra:
@claude review this

PR #1823, opened 2025-12-14
[forms]: fix multiple boolean input validation error

defymecobra:
### Problem
The issue was caused by the default behavior of the `[AllowedValues]` attribute when applied to collection properties (e.g., `string[]`). The validation logic was comparing the entire collection object against the individual allowed string values defined in the attribute. Since an array instance is never equal to a string value, the validation invariably failed with an error like "The field does not equal any of the values...", making it impossible to submit forms with multi-select inputs that used this attribute.

### Solution
I updated `FormHelpers.cs` to introduce specific handling for `AllowedValuesAttribute` when it is detected on collection types (like arrays or lists). Instead of using the default validation, the system now iterates through the collection and independently verifies that each selected item is present in the allowed values list. This ensures that the attribute works intuitively for multi-select scenarios, validating that all selected options are permitted, rather than checking the collection object itself.

### DIFF
--- a/Ivy/Views/Forms/FormHelpers.cs
+++ b/Ivy/Views/Forms/FormHelpers.cs
@@ -22,22 +22,9 @@ public static class FormHelpers
         {
             var capturedAttr = attr; // Capture for closure
 
-            if (attr is AllowedValuesAttribute && propertyInfo.PropertyType != typeof(string) && typeof(IEnumerable).IsAssignableFrom(propertyInfo.PropertyType))
+            if (TryCreateCollectionValidator(attr, propertyInfo.PropertyType, out var collectionValidator))
             {
-                validators.Add(value =>
-                {
-                    if (value is IEnumerable collection)
-                    {
-                        foreach (var item in collection)
-                        {
-                            if (!capturedAttr.IsValid(item))
-                            {
-                                return (false, capturedAttr.ErrorMessage ?? $"Value '{item}' is not allowed.");
-                            }
-                        }
-                    }
-                    return (true, "");
-                });
+                validators.Add(collectionValidator!);
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
@defymecobra write unit tests, both happy and sad path

defymecobra:
> @defymecobra write unit tests, both happy and sad path

Done

Ivy/Views/Forms/FormHelpers.cs
```diff
@@ -20,6 +21,26 @@ public static bool IsRequired(PropertyInfo propertyInfo)
         foreach (var attr in attributes)
         {
             var capturedAttr = attr; // Capture for closure
+
+            if (attr is AllowedValuesAttribute && propertyInfo.PropertyType != typeof(string) && typeof(IEnumerable).IsAssignableFrom(propertyInfo.PropertyType))
```
rorychatt: this code is super hard to understand.
I see that the test indeed fails on main before your change -> but is there a better way of making the test pass?
there is a bunch of ifs in foreach, this can tank the performance a lot in the long run

Ivy/Views/Forms/FormHelpers.cs
```diff
@@ -53,6 +74,26 @@ public static bool IsRequired(PropertyInfo propertyInfo)
         foreach (var attr in attributes)
         {
             var capturedAttr = attr; // Capture for closure
+
+            if (attr is AllowedValuesAttribute && fieldInfo.FieldType != typeof(string) && typeof(IEnumerable).IsAssignableFrom(fieldInfo.FieldType))
```
rorychatt: repetitive code. Create a helper function

rorychatt (changes requested):
left some feedback ->

try to make logic simpler, also break out repetitive code.

defymecobra:
Refactored the logic as requested. I extracted the `AllowedValues` validation into a separate `TryCreateCollectionValidator` helper method. This eliminated the code duplication between fields/properties and significantly simplified the main loop readability.

defymecobra:
@claude review this

PR #1829, opened 2025-12-14
[fonts]: prevent font flicker by changing font-display to fallback

defymecobra:
## Problem

When loading pages, text initially renders with a fallback font (typically Times New Roman) before switching to the Geist font family, causing a visible "flicker" effect. This issue occurs because `font-display: swap` instructs the browser to immediately display the fallback font while the custom font loads in the background, then swap to the custom font once it's available. This creates a noticeable visual transition that degrades the user experience, especially on slower network connections where the font loading delay is more pronounced.

## Solution

Changed `font-display` from `swap` to `fallback` for all Geist font family definitions in `index.css`. The `fallback` strategy provides a short blocking period (~100ms) for the font to load. If the font loads within this window, it's displayed immediately without any flicker. If it doesn't load in time, the fallback font is used for the current page load, but subsequent navigations will use the cached font. This approach eliminates the visible font switching while maintaining good performance characteristics.

Testing with network throttling confirmed that `fallback` eliminates the flicker:

https://github.com/user-attachments/assets/d150c3c9-83a0-4002-b5df-3fdd079b41bc

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -6,7 +6,7 @@
   src: url('/fonts/Geist/webfonts/Geist-Regular.woff2') format('woff2');
   font-weight: 400;
   font-style: normal;
-  font-display: fallback;
+  font-display: swap;
 }
 
 @font-face {
@@ -14,7 +14,7 @@
   src: url('/fonts/Geist/webfonts/Geist-Medium.woff2') format('woff2');
   font-weight: 500;
   font-style: normal;
-  font-display: fallback;
+  font-display: swap;
 }
 
 @font-face {
@@ -22,7 +22,7 @@
   src: url('/fonts/Geist/webfonts/Geist-SemiBold.woff2') format('woff2');
   font-weight: 600;
   font-style: normal;
-  font-display: fallback;
+  font-display: swap;
 }
 
 @font-face {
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
What worries me is this state you got
image

rorychatt:
We need to figure out how to make the fonts load first. so fonts load before all js or something.

then we will never have the flicker or the missing text state.

rorychatt (changes requested):
maybe the fonts are also heavy, that's why it takes time. 

We need to some how load the font files before anything else.

rorychatt:
image

defymecobra:
Implemented early font loading by adding `<link rel="preload">` tags for all 8 Geist font files (Regular, Medium, SemiBold, Bold for both Geist Sans and Geist Mono) in `index.html`, ensuring fonts are requested before any JavaScript or React code executes. Combined with `font-display: swap`, this approach guarantees that font files start downloading immediately during HTML parsing, significantly reducing the time window where fallback fonts are displayed. The preload strategy eliminates the flicker by ensuring fonts are available much earlier in the page load lifecycle, while `swap` maintains text visibility throughout the loading process. Testing confirms that this combination completely eliminates the font flicker issue.

Here how it looks with preload + `font-display: swap`:

https://github.com/user-attachments/assets/30a3e89b-342a-4a00-9921-0bcececc5767

defymecobra:
@claude review this

PR #1835, opened 2025-12-15
(fix): Fixed text cutoff in charts

### LINKED ISSUE #1813: (chart): tooltip out of bounds
### Description

Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

KaiserReich95:
Closes #1813 

Original issue was that text description was cutoff by container boundery for Piecharts. It seems that also happens for all other charts so I fixed them aswell.

Bar and Piecharts before:
ivy-barchart-before
ivy-piechart-before

Bar and Piecharts after:
ivy-barchart-after
ivy-piechart-after

### DIFF
--- a/frontend/src/widgets/charts/PieChartWidget.tsx
+++ b/frontend/src/widgets/charts/PieChartWidget.tsx
@@ -6,6 +6,7 @@ import {
   getColors,
   generateTextStyle,
   generateEChartToolbox,
+  generateTooltip,
 } from './sharedUtils';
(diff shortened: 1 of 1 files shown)


KaiserReich95:
I changed it now so pie chart uses the generateTooltip like all other charts. No text description is out of bounds now for all charts since all of them use the updated generateTooltip function.

image

PR #1837, opened 2025-12-15
(callout): use text-foreground on callouts

dcrjodle:
Changed to only use default text-foreground on callouts. I don't see a reason to use chromatic-foreground specific colors when we have opacitated text
Screenshot 2025-12-15 at 16 21 02


rorychatt:
if it works it works :D

Allthough for semantics, maybe `text-destructive-foreground ` and so on are the correct naming.

PR #1840, opened 2025-12-16
(fix): Implemented sizing to TextArea

### LINKED ISSUE #1815: (textInput): scale doesn't work with ToTextArea()
### Description

Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

KaiserReich95:
Closes #1815 

Now the Text area can have different sizes like all the other text inputs. Defaults to medium size:

image

Renamed textarea and Textarea to textArea and TextArea aswell. If we dont want that we can just revert latest commit.


Ivy/Widgets/Inputs/TextInput.cs
rorychatt: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea

Image

I think this change does not add value, so let's revert this

frontend/src/components/ui/textarea.tsx
```diff
@@ -1,22 +1,28 @@
...
-        'flex min-h-[60px] w-full rounded-md border border-input bg-transparent px-3 py-2 text-sm shadow-sm placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50',
-        className
-      )}
-      ref={ref}
-      {...props}
-    />
-  );
-});
-Textarea.displayName = 'Textarea';
+export interface TextAreaProps
```
dcrjodle, line 7: can you explain why this exported interface is needed

KaiserReich95: We use "export interface" in the input.tsx aswell. The code we had before had no support for scale props so thats why we couldnt scale the textarea.

KaiserReich95:
Could we merge this now? @dcrjodle

dcrjodle (approved):

PR #1844, opened 2025-12-17
[datatable]: use stable connectionKey instead of object reference to prevent rerender

defymecobra:
### Problem
The DataTable component was unnecessarily reloading data when switching between tabs (Demo/Code) in the documentation or when opening the "Show Code" sheet in Samples. This happened because the useEffect hook in `useDataLoading.ts` depended on the connection object reference. When the parent component re-mounted (which happens during tab switches), React created a new connection object in memory — even though the actual connection values (connectionId, sourceId, port, path) remained the same. Since JavaScript compares objects by reference, React considered the connection "changed" and triggered data reload.

### Solution
Replaced the object reference dependency with a stable string key connectionKey composed of connectionId and sourceId. This ensures that the useEffect hooks only trigger when the connection actually changes logically, not when the object reference changes due to component re-mounting.

Current look - it works without rerender now


defymecobra:
@claude review this

Issue #1812, opened 2025-12-12
(samples): The card with the code is not displayed at full height


ArtemLazarchuk:
### When the Show Code button is clicked, a sheet with the code opens, and the card’s height is hard-coded.
<img width="1906" height="956" alt="Image" src="https://github.com/user-attachments/assets/d7170993-bfa8-49ea-8c79-fbf79eaf3e99" />

<img width="1918" height="912" alt="Image" src="https://github.com/user-attachments/assets/923d64f5-89e5-4e27-8565-642964444ec4" />

### Code Samples

```csharp

```

KaiserReich95:
@rorychatt @ArtemLazarchuk  I changed it now so every code snippet in show code extends to the bottom of the page. Should it be like that or should short code snippets only extend as far as the code?

Short code snippet:
<img width="977" height="937" alt="Image" src="https://github.com/user-attachments/assets/a036b9d6-f5e9-403f-a039-411b19724fcf" />

Long code snippet:
<img width="973" height="940" alt="Image" src="https://github.com/user-attachments/assets/de7a8ba0-a2c0-4daf-8ffc-b02d0e6bb72e" />

rorychatt:
I think for short ones it should extend as much is actually needed.

original issue was that it did not extend, and instead user got a scrollbar (max heigth reached?)

rorychatt:
Here is a good demo of the issue where we don't set a max height value -> press show code when page loads

@KaiserReich95 https://ivy-projectdemos-snowflake-dashboard.sliplane.app/snowflake-dashboard/dashboard

source code in Ivy-Examples

ArtemKhvorostianyi:
<img width="1440" height="863" alt="Image" src="https://github.com/user-attachments/assets/8e3a18ea-5c14-4051-a656-d35a6768c453" />

PR #1855, opened 2025-12-17
(code): remove hardcoded height

dcrjodle:
Screenshot 2025-12-17 at 16 33 00
Screenshot 2025-12-17 at 16 32 52


dcrjodle:
@rorychatt  The issue @KaiserReich95 was working on #1812
We solved it together

rorychatt:
@dcrjodle @KaiserReich95

We should come up with a different solution. We should keep the max height constraint for default widgets, and this case simply override the given value. So the max height is 100% of remaining space, and there should be a Min() set, and where the code can grow from Min to Max

KaiserReich95:
I made some changes and submit them in another PR. Closing this for now.

PR #1858, opened 2025-12-17
(fix): Fixed so that code snippets fills sheet as much as needed

### LINKED ISSUE #1812: (samples): The card with the code is not displayed at full height

KaiserReich95:
Closes #1812 

Fixed so that the code snippets fills out the appropriate space needed in the sheet.

Before:
image

Short code snippet:
image

Long code snippet:
image

### DIFF
--- a/Ivy.Samples.Shared/Helpers/CodeView.cs
+++ b/Ivy.Samples.Shared/Helpers/CodeView.cs
@@ -24,7 +24,6 @@ public class CodeView(Type type) : ViewBase
             }
         }
 
-        return new Code(code, Languages.Csharp)
-            .Height(Size.MinContent().Max(Size.Full()));
+        return new Code(code, Languages.Csharp);
     }
 }


Ivy.Samples.Shared/Helpers/CodeView.cs
```diff
@@ -24,6 +24,7 @@ public class CodeView(Type type) : ViewBase
             }
         }
 
-        return new Code(code, Languages.Csharp);
+        return new Code(code, Languages.Csharp)
+            .Height(Size.MinContent().Max(Size.Full()));
```
KaiserReich95: The default Code widget has Height = Size.MaxContent().Max(Size.Px(800)) which caps height at 800px. For the "Show Code" sheet, we need the code to make it so short code snippets doesnt take over the whole space and large snippets will fill the whole page.

frontend/src/widgets/primitives/CodeWidget.tsx
```diff
@@ -84,11 +84,12 @@ const CodeWidget: React.FC<CodeWidgetProps> = memo(
     const dynamicTheme = useMemo(() => createPrismTheme(), []);
 
     return (
-      <div className="relative">
+      <div className={cn('relative', height && 'h-full pb-2')}>
```
KaiserReich95: When a height is set, the code widget fills its container completely. Adding pb-2 creates a small gap at the bottom so the code border doesn't touch the bottom of the sheet.

frontend/src/widgets/sheetWidget.tsx
```diff
@@ -53,7 +53,7 @@ export const SheetWidget: React.FC<SheetWidgetProps> = ({
       <SheetContent
         style={styles}
         className={cn(
-          'h-full flex flex-col p-0 gap-0',
+          'h-auto max-h-full flex flex-col p-0 gap-0',
```
KaiserReich95: h-full made the sheet always fill the entire viewport height. h-auto makes the sheet size to its content.  max-h-full caps it at viewport height so long content scrolls instead of overflowing.

frontend/src/widgets/sheetWidget.tsx
```diff
@@ -66,7 +66,7 @@ export const SheetWidget: React.FC<SheetWidgetProps> = ({
             {description && <SheetDescription>{description}</SheetDescription>}
           </SheetHeader>
         )}
-        <div className="flex-1 pb-0 pt-1 pl-4 pr-4 mt-4 overflow-y-auto">
+        <div className="min-h-0 pb-0 pt-1 pl-4 pr-4 mt-4 overflow-y-auto">
```
KaiserReich95: When content exceeds the available space, it stops at the sheet's max-h-full boundary when using min-h-0.

rorychatt:
@ArtemLazarchuk test this out

rorychatt:
@ArtemKhvorostianyi @ArtemLazarchuk

KaiserReich95:
The problem is that if the base class sheet has h-full like before the code snippets seem to only conform to two modes:

Either:

- (Scenario 1): Longer code snippets will look ok but the shorter code snippets will extend their bottom border to the bottom.

image

- (Scenario 2): Shorter code snippets will have the border just under them but longer code snippets border is not visible when scrolling to the top.

ivy-large-code-snippet

I cant seem to find another solution to the problem than choosing one of these scenarios above or changing the index file.

We can have a specific index.css rule that applies to all elements that want to change the sheets layout (like footer) that fixes the footer problem:

```
/* Sheet should be full height when containing FooterLayout */
.ivy-sheet:has(.remove-parent-padding) {
  height: 100%;
  max-height: 100%;
}

.ivy-sheet:has(.remove-parent-padding) > .ivy-sheet-content {
  flex: 1;
}
```

It can be even more specific to only apply to the footer if we want to.

The padding I added fixed the problem with the border of the Code snippet touching the bottom of the page but it added padding on other places we dont want it.

It can also be solved via having a specific CSS pattern when code widget are used within a sheet.

```
/* Code widget padding only when inside a sheet */
.ivy-sheet .ivy-code {
  padding-bottom: 1rem;
}
```

What do you guys think is the best solution? @rorychatt @ArtemKhvorostianyi

rorychatt:
I can not decide based on the feedback.

I think since this is getting tricky now, best is if I look at the issue myself and propose a solution. Doing something with (has) will lead us to hard-coding behavior in the FE.

We should have been able to define the responsiveness of the code window like we wanted with just using the Heigth API.

Best solution right now is to make it always fill the full available height, with header at the top, footer at the bottom, no scrollbar on main div, but scrollbar inside the code block. And once you do that, we merge this PR and reassign the issue to me

KaiserReich95:
Removed the changes that made shorter codesnippets to cutoff after the end of the code since that interfered with the footer. Also padding for longer code snippets is removed so they will touch the bottom of the sheet. This was because the padding will also effect code snippets not inside of a sheet aswell.

Now all codesnippets extend to the end of the sheet:
image

And footer looks ok again:
ivy-sheet-footer

rorychatt:
@ArtemKhvorostianyi pls test this

rorychatt:
image
BTW, is this a horizontal scrollbar?

KaiserReich95:
Yes it is. The problem is with the padding there, which effected the other code snippets we have on other places. We could implement a specific CSS rule to only fix padding for code snippets within sheets but since we want to avoid specific CSS rules it doesnt seem possible to implement without effecting footers or features that are close to the bottom of the sheet.

rorychatt:
@KaiserReich95 propose a solution

ArtemKhvorostianyi:
Screenshot 2026-01-11 at 17 31 25

We have very small padding at the bottom of the code snippet, and there’s also unnecessary horizontal scrolling. Does the code block have a hardcoded width?

KaiserReich95:
@rorychatt : I could add buttom padding to the sheet instead of the codewidget so all items except footer is affected. I didnt know footer had its own padding before so I thought it would be affected by padding changes in the sheet itself. 

Other widgets like TextWidget, MarkdownWidget, ButtonWidget, InputWidget, CardWidget, ImageWidget will be affected by this aswell. They are simular to CodeWidget in that they should not be touching the buttom of the sheet so adding padding for all of them would be OK I think.

New sheet padding:
image

Footer has the same padding still:
image

@ArtemKhvorostianyi : It seems the horizontal scrolling was like that from the beginning. The code snippet extends as far as the longest code line. Should I try to fix it so it wraps the code if the lines get to long? Maybe it should be made into a new issue then instead of fixing it here?

Main branch sheet horizontal scrolling:
https://github.com/user-attachments/assets/9cfebd3c-c647-4c60-ba05-0ed706ba92cc

PR #1871, opened 2025-12-19
[charts]: remove automatic alphabetical sorting in PivotTable

defymecobra:
## Problem

When using `.ToLineChart()` (and other chart builder methods), the x-axis values were sorted alphabetically instead of preserving the original data order. This caused months to appear in wrong order (e.g., "Apr → Feb → Jan → Mar" instead of "Jan → Feb → Mar → Apr"), breaking chronological visualization. The issue occurred because `PivotTable.ExecuteAsync()` applied `Comparer.Default.Compare` sorting on the first dimension, which performs lexicographic comparison on strings.

## Solution

Removed the automatic alphabetical sorting from `PivotTable.cs`. The data now preserves the order in which it was passed by the user, allowing proper chronological or custom ordering when using `.OrderBy()` before chart creation.

Before:
image
After:
image

### DIFF
--- a/Ivy/Views/Charts/PivotTable.cs
+++ b/Ivy/Views/Charts/PivotTable.cs
@@ -112,8 +112,6 @@ public class PivotTable<T>
             calculation.Calculation(result);
         }
 
-        //sort by first dimension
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

Issue #1787, opened 2025-12-10
(docs): error page content box is broken


defymecobra:
### Description

<img width="299" height="397" alt="Image" src="https://github.com/user-attachments/assets/3115a49d-a5fa-4df0-be80-65b3a4167b74" />

### Code Samples

```csharp

```

ArtemKhvorostianyi:
It probably was fixed with  https://github.com/Ivy-Interactive/Ivy-Framework/pull/1883
However, it works fine for now

<img width="1621" height="671" alt="Image" src="https://github.com/user-attachments/assets/63cfd767-7144-482c-91e0-01672361f9ec" />

PR #1872, opened 2025-12-20
[Docs]: fix leaking headings from Box widgets to the TOC

defymecobra:
Problem
Headings inside Box widgets (like "Revenue" from MetricView) were incorrectly appearing in the Table of Contents sidebar on documentation pages.

Cause
PR #1623 removed the DemoBox widget with its `data-demo-box` attribute that was used for TOC filtering, replacing it with `Box.Plain()` (was deleted later) without any marker attribute.

Solution
- Added `data-box` attribute to `BoxWidget.tsx` to mark all Box containers
- Updated `TableOfContents.tsx` to exclude headings that are inside `[data-box]` elements

### DIFF
--- a/frontend/src/widgets/primitives/BoxWidget.tsx
+++ b/frontend/src/widgets/primitives/BoxWidget.tsx
@@ -62,7 +62,7 @@ export const BoxWidget: React.FC<BoxWidgetProps> = ({
 
   return (
     <>
-      <div data-box style={styles} className={cn(className)}>
+      <div style={styles} className={cn(className)}>
         {children}
       </div>
     </>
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

defymecobra:
This PR also fixes [this issue](https://github.com/Ivy-Interactive/Ivy-Framework/issues/1787)
image

frontend/src/widgets/primitives/BoxWidget.tsx
```diff
@@ -62,7 +62,7 @@ export const BoxWidget: React.FC<BoxWidgetProps> = ({
 
   return (
     <>
-      <div style={styles} className={cn(className)}>
+      <div data-box style={styles} className={cn(className)}>
```
rorychatt: I don't think this is a good solution because it affects all boxes

rorychatt:
-> 
image

PR #1876, opened 2025-12-21
[GridView]: add Height

defymecobra:
What I mentioned in example that Artem provided in issue:
1 - you set 3 values for .RowHeights, but actually you have 4 rows
2 - without setting .Height() for layout, fraction won't work at all and this is correct behaviour (it don't know from what value fraction has to be taken)

About 2 punct, GridView don't have height at all, so I added it in this PR. Now fractions will work.

Here you can see the correct one:
image
image

### DIFF
--- a/Ivy/Views/GridView.cs
+++ b/Ivy/Views/GridView.cs
@@ -52,6 +52,12 @@ public class GridView : ViewBase, IStateless
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

defymecobra:
@claude look at WidgetBase, it already has this props.

Issue #1799, opened 2025-12-10
(docs): sheet page example refactor


defymecobra:
### Description

This kanban example isn't working

<img width="625" height="470" alt="Image" src="https://github.com/user-attachments/assets/d4c805f2-eb30-495d-ab52-da13888ade39" />

Also in content box we have kanban columns displayed. that's not an expected behaviour

<img width="295" height="405" alt="Image" src="https://github.com/user-attachments/assets/df2f4bcd-5aff-404c-a116-eeef3d9f21f5" />

### Code Samples

```csharp

```

PR #1877, opened 2025-12-21
[Kanban]: fix examples in docs

defymecobra:
Added correct .HandleMove() logic for 2 and 3 example that I've mentioned in issue. It solves the issue and kanban examples work as expected now.

For 1 one (Basic Usage example) I decided not to change it, because it shows to user, how the default usage is provided. If I add the .HandleMove() logic here, I'll have to make significant edits to the example code, which will make it not quite "Basic Usage" (in my opinion).


defymecobra:
@claude review this

defymecobra:
I've found out that [this issue](https://github.com/Ivy-Interactive/Ivy-Framework/issues/1799) also has the same problem, so I added the same solution here

defymecobra:
@claude review this

Issue #1881, opened 2025-12-22
(cards): fix null warning


rorychatt:
### Description

dotnet watch ⚠ /Users/rorychatt/git/ivy/Ivy-Framework/Ivy/Widgets/Card.cs(56,31): warning CS8604: Possible null reference argument for parameter 'anything' in 'LayoutView LayoutExtensions.WithLayout(object anything)'. [/Users/rorychatt/git/ivy/Ivy-Framework/Ivy/Ivy.csproj]

<img width="1653" height="1012" alt="Image" src="https://github.com/user-attachments/assets/f02838d2-bdcb-4160-8d5a-fdbccd49b8b0" />

### Code Samples

```csharp

```

PR #1884, opened 2025-12-22
(build): fix warning

dcrjodle:
#1881

### DIFF
--- a/Ivy/Widgets/Card.cs
+++ b/Ivy/Widgets/Card.cs
@@ -53,7 +53,7 @@ public static class CardExtensions
     {
         object? header = Layout.Vertical().Gap(0)
                          | (Layout.Horizontal().Align(Align.Center)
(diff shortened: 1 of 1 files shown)


Ivy/Widgets/Card.cs
```diff
@@ -53,7 +53,7 @@ public static Card Header(this Card card, object? title = null, object? descript
     {
         object? header = Layout.Vertical().Gap(0)
                          | (Layout.Horizontal().Align(Align.Center)
-                            | title.WithLayout().Grow()
+                            | title?.WithLayout().Grow()
                             | icon!)
```
rorychatt, line 57: these 2 lines in general might need better handling. What happens if the value passed is null?

What happens if in runtime the icon is undefined?

dcrjodle: niels pushed a fix for this to main

Issue #1888, opened 2025-12-22
(charts) Weird casing that is applied


nielsbosma:
### Description

<img width="721" height="340" alt="Image" src="https://github.com/user-attachments/assets/e508ab6c-1c1a-46b9-af7f-0b33abb31c45" />

### Code Samples

```csharp

```

PR #1890, opened 2025-12-22
[ECharts]: upgrade to v6

defymecobra:
## Overview

This PR updates the ECharts charting library from v5 to v6, along with the React wrapper `echarts-for-react`.

### Dependency Changes

| Package | Before | After |
|---------|--------|-------|
| `echarts` | - | `6.0.0` |
| `echarts-for-react` | `3.0.2` | `3.0.5` |

ECharts v6 is fully backward compatible with v5 at the API level.

## Issues Discovered After Upgrade

During testing, two visual issues were identified that appeared after the ECharts v6 update.

---

### Issue 1: Toolbox Position Changed

**Problem:** The chart toolbox (containing Data View, Magic Type, Save as Image buttons) moved from the top of the chart to the bottom.

**Before Fix:**
image

**After Fix:**
image

**Root Cause:** ECharts v6 changed the default positions and sizes of several components as part of their new default theme. According to the [V6 Upgrade Guide](https://echarts.apache.org/handbook/en/basics/release-note/v6-upgrade-guide):

> "The new styles in v6 not only change the theme colors, but also optimize and adjust the default positions and sizes of some components."

The string value `'top'` for the toolbox `top` property is now interpreted differently in ECharts v6.

**Solution:** Changed the toolbox `top` property from the string `'top'` to an explicit numeric value `-10` in `sharedUtils.ts`. This ensures the toolbox is positioned at the top of the chart area regardless of ECharts version.

---

### Issue 2: Data View Rendered Outside Chart Container

**Problem:** When clicking the Data View button in the toolbox, the data table appeared in the top-left corner of the page instead of inside the chart container. The Close and Refresh buttons were also displaced.

**Before Fix**:
image

**After Fix**:
image

**Root Cause:** ECharts Data View uses `position: absolute` for its overlay elements. These elements position themselves relative to the nearest positioned ancestor. The chart widget containers did not have position: relative set, causing ECharts elements to position relative to a higher-level ancestor (potentially the body).

Solution: Added `position: relative` to the container styles object in all chart widgets. This change ensures that all absolutely positioned ECharts elements (Data View overlay, buttons) are contained within their respective chart containers.

### DIFF
--- a/Ivy/Core/WidgetSerializer.cs
+++ b/Ivy/Core/WidgetSerializer.cs
@@ -12,7 +12,6 @@ public static class WidgetSerializer
     private static readonly JsonSerializerOptions SerializerOptions = new()
     {
         PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-        DictionaryKeyPolicy = JsonNamingPolicy.CamelCase,
         TypeInfoResolver = new DefaultJsonTypeInfoResolver(),
         Converters =
         {


rorychatt:
Can you also fix #1888 ?

rorychatt:
image
Can you make it so toolbox has the same font color like before?

rorychatt:
image
remove bg fill color, make it transparent so it matches bg in different themes

defymecobra:
> Can you also fix #1888 ?

yes

defymecobra:
Now it looks exactly as before:
image
image
image
image

rorychatt:
@defymecobra great. Let's also fix the casing issue in the same PR
Maybe also you could give the #1161 a try?

defymecobra:
# Fix: Chart Legend Casing (#1888)

## Problem
Chart legends displayed measure names with corrupted casing when using `SCREAMING_SNAKE_CASE` or similar naming:
- Input: `FRIENDS_AND_FAMILY`
- Output: `friendS_AND_FAMILY` ❌

## Root Cause
`WidgetSerializer.cs` used `DictionaryKeyPolicy = JsonNamingPolicy.CamelCase` which incorrectly transformed dictionary keys (including chart data measure names) to camelCase.

The CamelCase policy doesn't handle `SCREAMING_SNAKE_CASE` properly, resulting in corrupted output.

## Solution
Removed `DictionaryKeyPolicy` from `WidgetSerializer.cs`. This ensures that dictionary keys (measure names specified in `.Measure("...")`) are preserved exactly as defined by the user.

### Changed Files
- `Ivy/Core/WidgetSerializer.cs` — removed `DictionaryKeyPolicy = JsonNamingPolicy.CamelCase`

## Verification
After the fix, chart legends correctly display measure names as defined:
- Input: `FRIENDS_AND_FAMILY`
- Output: `FRIENDS_AND_FAMILY` ✅

## Impact
This change only affects dictionary key serialization. Other widgets don't use dictionaries in their props, so there's no risk of breaking changes.

defymecobra:
@claude review this

PR #1897, opened 2025-12-22
(theme): add popover to theme presets

dcrjodle:
Screenshot 2025-12-22 at 14 15 21

### DIFF
--- a/Ivy/Themes/ThemeService.cs
+++ b/Ivy/Themes/ThemeService.cs
@@ -31,7 +31,7 @@ public class ThemeService : IThemeService
 
         // Generate :root (light theme) variables
         sb.AppendLine(":root {");
-        AppendThemeColors(sb, _currentTheme.Colors.Light);
+        AppendThemeColors(sb, _currentTheme.Colors.Light, ThemeColors.DefaultLight);
         AppendNeutralColors(sb);
         AppendChromaticColors(sb);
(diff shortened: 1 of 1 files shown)


Ivy.Samples.Shared/Apps/Demos/ThemeCustomizer.cs
```diff
@@ -331,7 +333,9 @@ private string GenerateCSharpCode(Theme theme)
                 Accent = "#1A3A5C",
                 AccentForeground = "#E8F4FD",
                 Card = "#0F2A4A",
-                CardForeground = "#E8F4FD"
+                CardForeground = "#E8F4FD",
```
rorychatt, line 333: This leaves me curious.

What happens if a user provides a theme with not all colors set?
Do the remaining colors just disappear? Should we maybe just apply the default color in this cases?

Like in this case where Popover was not defined, I was still expecting the default color (from Ivy theme) to appear. What do you think?

dcrjodle: https://github.com/user-attachments/assets/5c93281f-bae7-46eb-a723-5488bd39f8b3

small recording to demonstrate the default fallbacks

dcrjodle: just pushed a commit

PR #1901, opened 2025-12-24
[frontend]: fix font flickering using ivy-design-system fonts

defymecobra:
## Problem

The `@tailwindcss/vite` plugin triggers CSS re-evaluation on DOM changes (e.g., sidebar collapsing during resize). Without proper Cache-Control headers from Vite's dev server, this caused repeated font requests and visible flickering.

When combined with the `<link rel="preload">` directives for fonts in `index.html`, this resulted in a noticeable font flickering. The browser would briefly display text in Times New Roman before switching to the intended Geist font, creating a jarring visual experience during development.

The root cause was identified through DevTools Network analysis: font files returned with status 200 (full reload) instead of 304 (cached) on refresh, and the Response Headers lacked any caching directives.

image

## Solution

Added a custom Vite plugin `fontCacheHeaders` that intercepts requests to font files and sets appropriate caching headers:

```typescript
const fontCacheHeaders = (): Plugin => ({
  name: 'font-cache-headers',
  configureServer(server) {
    server.middlewares.use((req, res, next) => {
      if (req.url?.includes('/fonts/') && req.url?.endsWith('.woff2')) {
        res.setHeader('Cache-Control', 'public, max-age=86400, immutable');
      }
      next();
    });
  },
});
```

The `Cache-Control: public, max-age=86400, immutable` header tells the browser:
- `public` — the resource can be cached
- `max-age=86400` — cache for 24 hours (sufficient for a dev session)
- `immutable` — the resource will not change, skip revalidation

This eliminates redundant font requests during development, preventing the font flickering issue while maintaining proper caching behavior.
 
image

Also I want to admit that this is a dev-only problem, without and with this fix it works on a production build.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -22,7 +22,7 @@
         "@dnd-kit/sortable": "10.0.0",
         "@dnd-kit/utilities": "3.2.2",
         "@glideapps/glide-data-grid": "6.0.4-alpha24",
-        "@ivy-interactive/ivy-design-system": "1.1.16",
+        "@ivy-interactive/ivy-design-system": "1.1.17",
         "@microsoft/signalr": "9.0.6",
         "@radix-ui/react-accordion": "1.2.12",
         "@radix-ui/react-alert-dialog": "1.1.15",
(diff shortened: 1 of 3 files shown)


defymecobra:
@claude review this

rorychatt:
@defymecobra I think we should move fonts to src/assets. Can you try this, if it solves the issue without this plugin?

Otherwise, if src/assets is a better location, we can also inlcude the plugin. But we need to be more direct in what gets the immutable flag. So we should place fonts in its separate folder and be more specific on what gets the header.

rorychatt:
@defymecobra Can you try a solution where we create an npm package and load the fonts using that?

defymecobra:
## Solution

Created a local npm package `@ivy/fonts` that contains the Geist font files and `@font-face` definitions. Fonts are now loaded through the package instead of directly from the public folder.

### Result:
- No font flickering on page load (dev + production)
- No font flickering on window resize
- Fonts properly cached

defymecobra:
@claude review this

defymecobra:
## Summary

Migrate font loading from local `@ivy/fonts` package to shared `ivy-design-system` npm package.

Here is [PR](https://github.com/Ivy-Interactive/Ivy-Design-System/pull/5) for Ivy-Design-System

defymecobra:
@claude review this

defymecobra:
Waiting for update in Ivy-Design-System

PR #1906, opened 2025-12-27
[PieChart]: disable magicType in toolbox

defymecobra:
# Problem
The PieChart toolbox displayed "switch to line chart" and "switch to bar chart" buttons that didn't work when clicked.

# Cause
ECharts magicType feature only works for chart series with compatible data structures (cartesian coordinates with xAxis/yAxis). PieChart uses radial data without axes, so type conversion is not supported by the library.

# Solution
Disabled magicType in the toolbox for PieChart component.
image


defymecobra:
@claude review this

PR #1913, opened 2025-12-28
[widgetRenderer]: add scale inheritance for nested widgets

defymecobra:
## Problem

When using `Card.Small()` or `Card.Large()`, the nested `Details` widget does not inherit the Card's scale. Instead, Details always renders with its default Medium scale.

**Root cause:**

1. `widgetRenderer.tsx` renders slot children recursively **before** they reach the parent widget
2. Each widget receives its props independently without knowledge of parent's scale
3. `DetailsBuilder` on the backend sets `scale = Medium` by default
4. There was no mechanism to propagate scale from parent widgets to their children during the rendering tree traversal

**Example:**
```csharp
new Card(details.ToDetails())
    .Small()  // Card gets Small scale
    // But Details inside still renders with Medium (default)
```

---

## Solution

Modified `widgetRenderer.tsx` to implement universal scale propagation:

1. Added `inheritedScale` parameter to `renderWidgetTree()` function
2. Each widget with a `scale` prop now passes it to all child widgets during rendering
3. Inherited scale takes priority over widget's default scale

**Key changes:**
- `widgetRenderer.tsx`: Added scale inheritance during tree traversal
- `DetailsWidget.tsx`: Removed default `scale = Scales.Medium` to allow inheritance

**Result:**
```
Card.Small() → Details → DetailItem (text-xs) ✓
Card.Large() → Details → DetailItem (text-base) ✓
```

Before:
image

After:
image

This solution is universal — any widget with scale automatically propagates it to children without requiring per-widget modifications.

### DIFF
--- a/Ivy.Docs.Shared/Docs/02_Widgets/03_Primitives/AudioPlayer.md
+++ b/Ivy.Docs.Shared/Docs/02_Widgets/03_Primitives/AudioPlayer.md
@@ -14,14 +14,14 @@ searchHints:
 Play audio content with browser controls. Supports common audio formats and provides customizable playback options.
 </Ingress>
 
-The `Audio` [widget](../../../01_Onboarding/02_Concepts/Widgets.md) displays an audio player with browser controls in your [app](../../../01_Onboarding/02_Concepts/Apps.md). This widget is for playing audio files, not recording them.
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
@nielsbosma check DM-s

PR #1842, opened 2025-12-16
(chore): Fixed expandable to have extention for disabled with toggle

KaiserReich95:
Closes #1820 

Expandable widgets does now have an extention for a switch that toggles off or on the disabled stance.

https://github.com/user-attachments/assets/e01d3fae-a324-4b99-aa72-8873a0b806a6

Issue #1820, opened 2025-12-14
(expandable): disabled does not get proper cursor disabled


rorychatt:
### Description

<img width="626" height="237" alt="Image" src="https://github.com/user-attachments/assets/3298d38e-f444-4bf2-9d8a-26a0dbb90665" />

### Code Samples

```csharp

```

KaiserReich95:
It should look like this right?

https://github.com/user-attachments/assets/e53fa7c0-982f-45bb-8c80-84e6924a99ac

rorychatt:
Yes, but if you have a switch on it, the switch needs to still be functional
https://samples.ivy.app/widgets/expandable

<img width="1124" height="329" alt="Image" src="https://github.com/user-attachments/assets/34c47554-ff4e-4495-b464-32c689385d5d" />
meaning that when you hover on switch ehre, even though select is disabled, switch should be active

ArtemKhvorostianyi:
Closed by: https://github.com/Ivy-Interactive/Ivy-Framework/pull/2128

PR #1925, opened 2025-12-29
Toggle for disabled in expandable

KaiserReich95:
Releates to this issue: #1820 

Joel pointed out that the last solution to the problem had to much code to solve the problem: #1842. I tried to fix that in this version.

I have some questions about how the expandables should look like:

Question 1: 
What should we do with the extention "disabled" when we have switch on it. Before we had disabled with a boolean but why does it need to have a boolean? If we add the extention we obviously want it disabled right, why put disabled to false there? In this version the boolean is for the toggle instead. So false is disabled without toggle and true is disabled with toggle. Otherwise we can have another extention on top of the disabled one if we want to add a toggle.

Question 2:
Should the only option for toggling disabled be a switch or other types of toggles aswell (for example checkbox)? If we want other types of toggles then we cant just have a boolean for disabled and it also makes the code more complicated.

I added an example of the expandable widget to show the disabled toggle switch and another switch just added to the expandble that does nothing:

https://github.com/user-attachments/assets/cdfcfd77-7a26-4941-a23b-907b2ec58429

### DIFF
--- a/Ivy.Samples.Shared/Apps/Widgets/ExpandableApp.cs
+++ b/Ivy.Samples.Shared/Apps/Widgets/ExpandableApp.cs
@@ -37,12 +37,12 @@ public class ExpandableApp : SampleBase
                 "Generous spacing gives the content breathing room and improves readability.")
         ).Large();
 
+        var notificationSwitch = UseState(false);
         var toggleableExpandable1 = new Expandable(
             Layout.Horizontal()
-            | Text.Block("Apps")
-            | new Icon(Icons.ChevronRight)
-            | new Icon(Icons.Paperclip)
-            | Text.Block("Attachments"),
+            | Text.Block("Attachments")
+            | notificationSwitch.ToSwitchInput()
+            | Text.Block("Enable notifications"),
             Text.Block("This is the content for Attachments")
         ).Disabled(true);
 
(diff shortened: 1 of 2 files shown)


rorychatt:
The goal is: `expandable` can be disabled due to state, meaning it should not be able to expand.

However, you can have a switch on top of it that can mark the `expandable` as valid, so it can then expand.

Currently in `ivy db generate` you can see the use case for this for selecting apps you want to generate -> the expandable expands to show the per-app prompt, but if you switch the switch to off, the expandable becomes `invalid` and the app is not then generated.

To answer the question if we need to support more inputs I need to look at the code myself. So far right now there are no other such use cases, but I think that yes we should support any input.

but this is likely just a "leaking" issue where invalid state on parent for some reason in the FE is also applied to the child.
For case of expendables, it seems we need to simply stop propagation

rorychatt:
It is also okay if the switching the value on/off In switch does not affect the disabled state for expandable itself as the logic for this in other example suggests that the disabled state is handled externally.

In the Ivy-samples example, we just need to make sure that even though the expandable is disabled/invalid, you can still switch on and off the switch.

KaiserReich95:
I tried to fix so that the opacity applied for disabled expandables specifically doesnt apply to any toggles but that seems to not be possible. I could make the toggles be outside of the expandable and have absolute positioning to put them in the correct place. That solution requires alot of new code and is quite complicated to get right. Getting the spacing between the different items within the expandable when we dont know how many toggles we want to be absolutly positioned requires alot of targeted code. I cant seem to find a good solution to the problem then so I dont know what to do. Maybe someone else should try to fix this issue instead.

PR #1926, opened 2025-12-29
[Datatables]: lazy initialize GrpcTableService to avoid overhead for apps without datatables

defymecobra:
## Problem

The `GrpcTableService` singleton was being instantiated immediately when the module was imported, regardless of whether the application actually uses DataTables.

```typescript
// Old code - eager initialization
export const grpcTableService = new GrpcTableService('');
```

This caused the constructor's debug logging (`GrpcTableService constructor - Connecting to:`) to appear in the browser console for **every** application, even those without any DataTable widgets. The issue occurs because JavaScript executes the entire module upon first import — even when only importing types like `Filter` or `SortOrder`.

## Solution

Implemented lazy initialization using a JavaScript `Proxy`. The `GrpcTableService` instance is now created only when one of its methods is actually called:

```typescript
let _grpcTableServiceInstance: GrpcTableService | null = null;

export function getGrpcTableService(): GrpcTableService {
  if (!_grpcTableServiceInstance) {
    _grpcTableServiceInstance = new GrpcTableService('');
  }
  return _grpcTableServiceInstance;
}

export const grpcTableService = new Proxy({} as GrpcTableService, {
  get(_target, prop) {
    const instance = getGrpcTableService();
    const value = instance[prop as keyof GrpcTableService];
    if (typeof value === 'function') {
      return value.bind(instance);
    }
    return value;
  },
});
```

This approach:
- **Avoids initialization overhead** for apps that don't use DataTables
- **Maintains backward compatibility** — existing code using `grpcTableService.queryTable()` works without changes
- **Eliminates console noise** — the constructor log only appears when DataTables are actually used

### DIFF
--- a/frontend/src/services/grpcTableService.test.ts
+++ b/frontend/src/services/grpcTableService.test.ts
@@ -765,8 +765,13 @@ describe('GrpcTableService', () => {
   });
 
   describe('singleton instance', () => {
-    it('should export a singleton grpcTableService', () => {
-      expect(grpcTableService).toBeInstanceOf(GrpcTableService);
+    it('should export a lazy grpcTableService proxy that delegates to GrpcTableService', () => {
+      // grpcTableService is now a Proxy that lazily initializes GrpcTableService
+      // We test that it has the expected methods
+      expect(typeof grpcTableService.queryTable).toBe('function');
+      expect(typeof grpcTableService.parseFilter).toBe('function');
+      expect(typeof grpcTableService.isStreaming).toBe('function');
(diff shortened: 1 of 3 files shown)


defymecobra:
@claude review this

defymecobra:
- grpcTableService.test.ts: Update singleton test to verify proxy exports methods correctly
- tableDataFetcher.test.ts: Mock getGrpcTableService instead of grpcTableService proxy
- tableDataFetcher.ts: Use getGrpcTableService() for proper test mocking support

PR #1927, opened 2025-12-29
[Auth]: allow user to logout even without user info

defymecobra:
## Problem

The `DefaultSidebarChrome` component conditionally displays the user footer based on the result of `GetUserInfoAsync()`. When user info is available, a full footer with avatar, name, and a **Logout** button is shown. However, when `GetUserInfoAsync()` returns `null` (due to an expired token, network issues, or provider errors), the code falls back to showing a generic "Settings" menu **without** the Logout option.

This creates a deadlock situation: the user is authenticated (has a valid session), but cannot log out because the UI doesn't provide that option. The user is effectively stuck and has no way to sign out through the application interface.

## Solution

Instead of creating a separate footer variant for the "logged in but no user info" scenario, I chose a simpler approach: add the **Logout** menu item directly to the existing Settings footer when the user has an active auth session.

The fix:
1. Check `auth.GetAuthSession()` to determine if the user is authenticated
2. In the Settings footer (the `else` branch), conditionally include the Logout menu item based on `isLoggedIn`

This approach keeps the code minimal and avoids UI duplication, while ensuring users can always log out regardless of whether their profile information could be fetched.

How it looks:
image

### DIFF
--- a/Ivy/Chrome/DefaultSidebarChrome.cs
+++ b/Ivy/Chrome/DefaultSidebarChrome.cs
@@ -394,24 +394,6 @@ public class DefaultSidebarChrome(ChromeSettings settings) : ViewBase
                 ..commonMenuItems, MenuItem.Default("Logout").Tag("$logout").Icon(Icons.LogOut).HandleSelect(onLogout)
             ], navigator));
         }
-        else if (isLoggedIn)
-        {
-            var trigger = new Button()
-                .Content(
-                    Layout.Horizontal().Align(Align.Left)
-                        | Icons.User.ToIcon()
-                        | Text.Muted("User")
-                    )
-                    .Variant(ButtonVariant.Ghost).Width(Size.Full());
-
-            footer = new DropDownMenu(
-                    DropDownMenu.DefaultSelectHandler(),
-                    trigger)
-                .Top()
-                .Items(settings.FooterMenuItemsTransformer([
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

Ivy/Chrome/DefaultSidebarChrome.cs
```diff
@@ -403,12 +404,16 @@ void OnTabReorder(Event<TabsLayout, int[]> @event)
...
+                ? [.. commonMenuItems, MenuItem.Default("Logout").Tag("$logout").Icon(Icons.LogOut).HandleSelect(onLogout)]
+                : commonMenuItems;
+
             footer = new DropDownMenu(
                     DropDownMenu.DefaultSelectHandler(),
                     trigger)
                 .Top()
                 .Items(
-                    settings.FooterMenuItemsTransformer(commonMenuItems, navigator)
+                    settings.FooterMenuItemsTransformer(footerMenuItems, navigator)
```
rorychatt, line 416: are we getting rid of the default common menu items here?

So here we see the "LogOut" only, without the other elements that are present, for example, in the footer menu?

Otherwise, LGTM

zachwolfe: Based on my testing, the common menu items are still present, whether we are logged in with user info, logged in without user info, or not logged in (auth not configured).

zachwolfe (approved):

PR #1931, opened 2025-12-30
fix(core): full replacement of root node

zachwolfe:
This PR fixes a bug where I couldn't get past the first screen of `ivy db generate`, "Generate Database Schema" would spin forever. Meanwhile, the frontend had already received a SignalR message from the backend with the next screen's diff.


rorychatt:
@claude review this

PR #1932, opened 2025-12-30
(fix): Removed cropping from code snippets and edited copy button

### LINKED ISSUE #1859: (docs): code examples cropped on the right
### Description

Image

https://docs.ivy.app/widgets/advanced/chat

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Code Samples

```csharp

```

### Screenshots

_No response_

### Additional Context

_No response_

KaiserReich95:
Closes #1859 

Code shown in code snippets in Ivy.Docs is no longer cutoff by the padding the copy button had before.

https://github.com/user-attachments/assets/5f389bf7-40e0-49d6-9e2d-b11ececdea05

### DIFF
--- a/frontend/src/components/CopyToClipboardButton.tsx
+++ b/frontend/src/components/CopyToClipboardButton.tsx
@@ -18,13 +18,13 @@ const copyIconVariants = cva('', {
 });
 
 const copyButtonSizeVariants = cva(
-  'p-1 rounded hover:bg-accent focus:outline-none cursor-pointer flex items-center',
+  'p-2 rounded hover:bg-accent focus:outline-none cursor-pointer flex items-center',
   {
     variants: {
       scale: {
-        Small: 'h-5',
-        Medium: 'h-6',
-        Large: 'h-7',
+        Small: 'h-6',
+        Medium: 'h-8',
+        Large: 'h-9',
       },
     },
     defaultVariants: {


rorychatt:
image

It overlaps with scrollbar so this needs fixing.
Maybe we can also experiment with setting the same spacing so it's harmonic with the rest of the UI

KaiserReich95:
Seems like the padding was different between main and the branch. Reverted back to the padding from main. Now it looks the same as before.

Option 1 (Current commit):
ivy-copy-button-different-padding

I dont know if this was exactly what you suggested when you wrote "Maybe we can also experiment with setting the same spacing so it's harmonic with the rest of the UI". So I tested with the same spacing as the code has.

Option 2:
ivy-copy-button-same-padding

Maybe it also looks better with a bigger button:

Option 3:
ivy-copy-button-different-padding-large
Option 4:
ivy-copy-button-same-padding-large

Let me now which option you prefer or if it should be different than these options. @rorychatt

rorychatt:
choose between 3 and 4.
We also need to make sure this is harmonic across different scales. Maybe scale is not implemented for this widget yet.

rorychatt:
(code): implement scale API

KaiserReich95:
I chose option 3, I think option 4 has to much spacing for the button. As you said we have not implemented scale for code snippets yet so it always defaults to medium scale.

PR #1936, opened 2025-12-31
[Docs]: add Layout page

defymecobra:
image
image
image
image
image
image
image
image
image
image

## Why no API section?

The `Layout` class is a **static factory class** with no instantiable properties or configurable state — it only provides factory methods (`Vertical()`, `Horizontal()`, `Center()`, etc.) that return `LayoutView` instances. Since the `<WidgetDocs>` component is designed to document widget properties, constructors, and setters, it doesn't provide meaningful documentation for a static class. Adding a `WidgetDocs` tag for `LayoutView` would be misleading, as the page documents the `Layout` entry point, not the `LayoutView` implementation details. The comprehensive method tables in the "Available Methods" and "Extension Methods" sections serve as the API reference for this page.

## Location

Placed in `01_Onboarding/02_Concepts/` rather than `02_Widgets/04_Layouts/` because `Layout` is a **concept/pattern** for building UIs, not a specific widget. This aligns with other conceptual pages like `Views.md` and `Widgets.md`.

### DIFF
--- a/Ivy.Docs.Shared/Docs/01_Onboarding/02_Concepts/Layout.md
+++ b/Ivy.Docs.Shared/Docs/01_Onboarding/02_Concepts/Layout.md
@@ -14,10 +14,10 @@ searchHints:
 # Layout
 
 <Ingress>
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

PR #1939, opened 2025-12-31
(fix): unblock OAuth popups on Safari

### LINKED ISSUE #744: [Feature]: Avoid blocked OAuth login popup on Safari
### Description

Currently: to login with an OAuth provider (i.e. Microsoft Entra, Supabase), the user clicks on the appropriate button, which leads to this message on Safari:

Image

To open the login page, they must then hit this button:

Image

I think the reason for this is because the `onclick` handler leads to a roundtrip between the frontend and backend, which I guess takes too long (from Safari's POV) to redirect to the login page. So Safari sees it as a malicious popup instead of a direct result of the user's action.

Note: I tested Entra and Supabase auth on Safari, Chrome and Firefox, and it seems like only Safari has this issue.

### Use Case

_No response_

### Proposed Solution

Pre-compute OAuth login URLs, and pass them directly to the button so we can navigate to them immediately after the user clicks.

Complication: I think the URLs expire after some time for some providers? So we'll probably need to have a timer that periodically asks the provider for new URLs as well.

### Alternatives Considered

_No response_

### Impact

_No response_

### Code Examples

```csharp

```

### Additional Context

_No response_

zachwolfe:
Fixes #744

### DIFF
--- a/Ivy/Auth/AuthController.cs
+++ b/Ivy/Auth/AuthController.cs
@@ -20,22 +20,26 @@ public class AuthController() : Controller
         [FromQuery] string optionId,
         [FromQuery] string callbackId,
         [FromQuery] string connectionId,
-        [FromServices] AppSessionStore sessionStore)
+        [FromServices] AppSessionStore sessionStore,
+        [FromServices] ILogger<AuthController> logger)
     {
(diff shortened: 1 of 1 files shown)


zachwolfe:
@claude review please

Issue #1796, opened 2025-12-10
(docs): line chart page example refactor


defymecobra:
### Description

When applying too much width, this visual bug appears. I think we just need to shorten the max value user can apply.
<img width="669" height="552" alt="Image" src="https://github.com/user-attachments/assets/ddd07103-b4b5-4017-95d8-19c356623446" />

### Code Samples

```csharp

```

Issue #1795, opened 2025-12-10
(docs): tabs layout page example refactor


defymecobra:
### Description

When clicking on "Security", tab strangly hides. Also Im not sure that other examples on the page work as expected
<img width="707" height="325" alt="Image" src="https://github.com/user-attachments/assets/efc0a97f-b773-4f03-adce-8305c00943fd" />

<img width="673" height="238" alt="Image" src="https://github.com/user-attachments/assets/285ec713-df95-4be2-af66-6a67bf54c3a3" />

### Code Samples

```csharp

```

Issue #1789, opened 2025-12-10
(docs): skeleton page example refactor


defymecobra:
### Description

<img width="914" height="526" alt="Image" src="https://github.com/user-attachments/assets/3233318d-d24b-4ce3-ad05-d72467d464ba" />

### Code Samples

```csharp

```

Issue #1788, opened 2025-12-10
(docs): separator page example refactor


defymecobra:
### Description

<img width="795" height="502" alt="Image" src="https://github.com/user-attachments/assets/ad49c5db-3388-4c26-b98b-da023059955f" />

### Code Samples

```csharp

```

PR #1943, opened 2026-01-02
[Docs]: fix bunch of issues

defymecobra:
This PR fixes bunch of issues that were created by me before. Each commit resolves each issue
---
## Issue #1796 
Before: 
image
image
After:
image
image
Solution: refatored the example. Problem was in max value for width and useless layouts that were breaking number select inputs

## Issue #1795
Before:
image
After:
image
image
Solution: Added w-full class to the root div and container ref in `ContentVariant` component to ensure proper width inheritance, and added a new "Responsive Overflow" section to the TabsLayout documentation with a 13-tab example demonstrating the overflow behavior. Problem was that the `ContentVariant` component's containers were not explicitly set to full width, causing nested TabsLayout widgets to inherit a narrow container width from their parent Box element, which triggered the responsive overflow logic to incorrectly hide tabs in a dropdown menu even when there was sufficient space available.

## Issue #1789 
Before:
image
After:
image
Solution: replaced not working link in example for working one.

## Issue #1788
Before:
image
After: 
image
Solution: removed .Width() for layout in example, that was breaking that example

### DIFF
--- a/Ivy.Docs.Shared/Docs/02_Widgets/03_Primitives/Separator.md
+++ b/Ivy.Docs.Shared/Docs/02_Widgets/03_Primitives/Separator.md
@@ -38,7 +38,7 @@ public class ProfileDetailView : ViewBase
             }
         };
         
-        return Layout.Vertical().Gap(4).Width(Size.Units(600))
+        return Layout.Vertical().Gap(4)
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

PR #1961, opened 2026-01-05
(docs): update and refactor typography

dcrjodle:
There was a lot of tech debt inside the /lib folder so I refactored that, including changing of file and folder names. 
Most of the changes are either name changes, or tailwind changes

I removed the gap in the text containers so we can have more consistent spacing and so it doesn't interfere with the margins, but that meant I had to add margin to the code widgets and the callout widget

- ingresses are bigger (18px)
- paragraphs are bigger (16px)
- < H2s are thinner (font-medium)
- articles are more narrow
- now using system where 
- margins above headings are line height * 2 and margins below headings are line height / 2
- margins below paragraphs are line height * 1 

### Large screen

Screenshot 2026-01-05 at 15 45 38 (2)

### Before

Screenshot 2026-01-05 at 15 32 20

### After

Screenshot 2026-01-05 at 15 32 06


frontend/src/widgets/article/ArticleWidget.tsx
```diff
@@ -31,10 +31,10 @@ export const ArticleWidget: React.FC<ArticleWidgetProps> = ({
   const articleRef = useRef<HTMLElement>(null);
 
   return (
-    <div className="flex flex-col gap-2 max-w-7xl mx-auto px-4 sm:px-6 lg:px-8 relative mt-8">
-      <div className="flex gap-8 flex-grow">
+    <div className="flex flex-col gap-2 max-w-7xl mx-auto px-4 sm:px-6 lg:px-16 relative mt-8">
```
dcrjodle: gap doesn't make the article narrower, need to adjust the max width for that to work

rorychatt:
image
too big font here

rorychatt:
image
The text app now just confuses me. We need to reorganize this

rorychatt:
I wrote a test app and commited it to this repo so we can compare

http://localhost:5173/tests/typography-comparison



the html and md render slightly differently

image


(I think column in the middle is okay don't look at it)

rorychatt:
Ingress IMO should not be a larger font and should match regular P 
image

rorychatt:
UPD: text also needs to look similar, I pushed most recent code to the example test app
image

rorychatt:
Also here we should see all font sizes to be equal
image

http://localhost:5173/tests/different-font-cases

dcrjodle:
@rorychatt Did what we discussed. Tested Samples and Docs a lot to verify everything looks good. 
If we want to organise TexpApp.cs maybe we can do it in another PR, since it feels kind of unrelated and requires a lot of extra work

PR #1963, opened 2026-01-05
(fix): Fixed sidebar arrow for nested grouped items

### LINKED ISSUE #1949: (sidebar): arrow is not updating
### Description

Image

KaiserReich95:
Closes #1949 

https://github.com/user-attachments/assets/3cf09c5c-2d25-4e1c-8e12-b19a731cdbe8

### DIFF
--- a/frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
+++ b/frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
@@ -203,7 +203,7 @@ const CollapsibleMenuItem: React.FC<{
   widgetId: string;
   level: number;
 }> = ({ item, eventHandler, widgetId, level }) => {
-  const [isOpen, setIsOpen] = useState(item.expanded);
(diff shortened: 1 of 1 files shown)


frontend/src/widgets/layouts/sidebar/SidebarLayoutWidget.tsx
```diff
@@ -203,7 +203,7 @@ const CollapsibleMenuItem: React.FC<{
   widgetId: string;
   level: number;
 }> = ({ item, eventHandler, widgetId, level }) => {
-  const [isOpen, setIsOpen] = useState(item.expanded);
+  const [isOpen, setIsOpen] = useState(item.expanded ?? false);
```
dcrjodle, line 206: Is this needed?

rorychatt: my guess is not really, but in case if this was null, sets it to false. But I think its okay here

PR #1971, opened 2026-01-08
[Serialization]: add AlwaysSerialize for input Value properties

defymecobra:
## Problem

When using `UseState<T>()` with an enum type without specifying an initial value, the `SelectInput` widget appears empty even though the state correctly holds the first enum value.

### Root Cause

The `WidgetSerializer` optimizes JSON payload size by skipping properties that equal their default values. For `SelectInput<TValue>`:

1. `UseState<Priority>()` initializes state with `default(Priority)` = `Priority.Low` (index 0)
2. Serializer creates a default instance via `internal SelectInput() { }` where `Value = default!` = also 0
3. Serializer compares: `Priority.Low == Priority.Low` → skips serialization
4. Frontend receives JSON without `value` property → displays empty select

| User Value | Default Instance | Comparison | Result |
|------------|------------------|------------|--------|
| `Priority.Low` (0) | `default!` = 0 | 0 == 0 | ❌ SKIPPED |
| `Priority.Medium` (1) | `default!` = 0 | 1 != 0 | ✅ Serialized |
| `Priority.High` (2) | `default!` = 0 | 2 != 0 | ✅ Serialized |

---

## Solution

Add `AlwaysSerialize` property to `PropAttribute` to bypass default value optimization for specific properties.

#### `Ivy/Core/PropAttribute.cs`
```diff
+ public bool AlwaysSerialize { get; set; } = false;
```

#### `Ivy/Core/WidgetSerializer.cs`
```diff
- if (metadata.DefaultInstance != null)
+ if (!propInfo.Attribute.AlwaysSerialize && metadata.DefaultInstance != null)
```

#### `Ivy/Widgets/Inputs/SelectInput.cs`
```diff
- [Prop] public TValue Value { get; } = default!;
+ [Prop(AlwaysSerialize = true)] public TValue Value { get; } = default!;
```

---

## Impact

- ✅ First enum value now displays correctly
- ✅ Other input properties still use optimization (Disabled, Placeholder, etc.)
- ✅ Non-input widgets unchanged
- ⚠️ Minimal JSON size increase for Value property only

### DIFF
--- a/Ivy/Core/WidgetSerializer.cs
+++ b/Ivy/Core/WidgetSerializer.cs
@@ -5,7 +5,6 @@ using System.Text.Json.Nodes;
 using System.Text.Json.Serialization;
 using System.Text.Json.Serialization.Metadata;
 using Ivy.Core.Helpers;
-using Ivy.Widgets.Inputs;
 
 namespace Ivy.Core;
 
@@ -123,7 +122,7 @@ public static class WidgetSerializer
                 Type.EmptyTypes,
                 null);
 
-            if (defaultCtor != null && !typeof(IAnyInput).IsAssignableFrom(t))
+            if (defaultCtor != null)
             {
                 try
                 {
@@ -179,8 +178,8 @@ public static class WidgetSerializer
         {
             var value = GetPropertyValue(widget, propInfo.Property, propInfo.Attribute);
 
-            // Skip properties that match their default values
-            if (metadata.DefaultInstance != null)
+            // Skip properties that match their default values (unless AlwaysSerialize is set)
+            if (!propInfo.Attribute.AlwaysSerialize && metadata.DefaultInstance != null)
             {
                 var defaultValue = GetPropertyValue(metadata.DefaultInstance, propInfo.Property, propInfo.Attribute);
                 if (ValuesAreEqual(value, defaultValue))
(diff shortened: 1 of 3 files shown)


defymecobra:
@claude review this

Ivy/Core/WidgetSerializer.cs
```diff
@@ -122,7 +123,7 @@ private static SerializationTypeMetadata GetMetadata(Type type)
                 Type.EmptyTypes,
                 null);
 
-            if (defaultCtor != null)
+            if (defaultCtor != null && !typeof(IAnyInput).IsAssignableFrom(t))
```
rorychatt: I think we need another solution. Here we basically tell that "dont optimize if you are an input" -> but we still want to optimize.

defymecobra: yes, agree

Ivy/Widgets/Inputs/SelectInput.cs
```diff
@@ -81,7 +81,7 @@ public SelectInput(IEnumerable<IAnyOption> options, string? placeholder = null,
 
     internal SelectInput() { }
 
-    [Prop] public TValue Value { get; } = default!;
+    [Prop(AlwaysSerialize = true)] public TValue Value { get; } = default!;
```
rorychatt, line 84: we keep this for now, but if more cases like this happen we might need to opt in for the check by `value` name

PR #1972, opened 2026-01-09
[Charts]: add SortBy API for X-axis data sorting

defymecobra:
## Problem

Chart data on the X-axis was displayed in the exact order it was provided in the data array, without any sorting. This caused unexpected behavior in common scenarios:

**Numeric strings:** Data like `"1", "10", "2", "3"` would display as `1, 10, 2, 3` instead of the expected numeric order `1, 2, 3, 10`.

**Unsorted data sources:** When data came from databases or APIs without guaranteed ordering, charts would show data in arbitrary order.

**Date strings:** Dates like `"30 Dec", "17 Dec", "28 Dec"` would display in definition order rather than chronological order.

The root cause is that the `generateDataProps` function on the frontend simply extracts categories via `data.map()` without any sorting, and the backend (PivotTable) returns grouped data without ordering guarantees.

## Solution

Added a fluent `.SortBy()` API to `LineChartBuilder`, `BarChartBuilder`, and `AreaChartBuilder` that allows users to control X-axis data ordering on the backend before sending to the frontend.

**Two overloads provided:**

1. **Simple sort by dimension value:**
   ```csharp
   .SortBy(SortOrder.Ascending)
   ```

2. **Custom key selector for type conversion:**
   ```csharp
   .SortBy(row => int.Parse(row["Label"].ToString()!), SortOrder.Ascending)
   .SortBy(row => DateTime.Parse(row["Date"].ToString()!), SortOrder.Ascending)
   ```

The custom selector allows users to explicitly define how string values should be interpreted for sorting (as numbers, dates, etc.), solving the problem that strings cannot be automatically sorted in a semantically correct way.

## Changes

- Added `SortOrder` enum (`None`, `Ascending`, `Descending`) to `Ivy.Charts`
- Added `SortBy()` and `SortBy<TKey>()` methods to `LineChartBuilder`, `BarChartBuilder`, `AreaChartBuilder`
- Sorting is applied in `Build()` after data aggregation

image
image
image
image
image
image

### DIFF
--- a/Ivy/Views/Charts/PivotTable.cs
+++ b/Ivy/Views/Charts/PivotTable.cs
@@ -220,37 +220,24 @@ public class PivotTableBuilder<TSource>(IQueryable<TSource> data)
 
     private Dictionary<string, object>[] ApplySorting(Dictionary<string, object>[] data)
     {
+        var dimension = _dimensions.FirstOrDefault();
+        if (dimension == null)
+        {
+            return data;
+        }
+        var dimensionName = dimension.Name;
+
         if (_sortSelector != null)
         {
-            // Compile the selector and create a key extractor that works with Dictionary rows
-            var compiledSelector = _sortSelector.Compile();
-            var dimensionName = _dimensions.FirstOrDefault()?.Name ?? string.Empty;
-
-            // We need to extract the original value and apply the selector
-            // Since data is already aggregated, we use the dimension value
-            Func<Dictionary<string, object>, object> keyExtractor = row =>
+            if (TryExtractParseMethod(_sortSelector.Body, out var parseType))
             {
-                if (!row.TryGetValue(dimensionName, out var dimValue))
-                    return null!;
-                // Apply the original selector logic to convert the value
-                // We create a mock object with just the dimension value
-                return dimValue;
-            };
-
-            // For complex selectors like int.Parse(e.Label), we need to apply the transformation
-            // Get the expression body to understand what transformation is needed
-            var body = _sortSelector.Body;
-            if (body is UnaryExpression { Operand: MethodCallExpression methodCall })
-            {
-                // Handle int.Parse, DateTime.Parse, etc.
-                var method = methodCall.Method;
-                if (method.DeclaringType == typeof(int) && method.Name == "Parse")
+                if (parseType == typeof(int))
                 {
                     return _sortOrder == SortOrder.Ascending
                         ? data.OrderBy(row => int.TryParse(row.TryGetValue(dimensionName, out var v) ? v?.ToString() : null, out var num) ? num : 0).ToArray()
                         : data.OrderByDescending(row => int.TryParse(row.TryGetValue(dimensionName, out var v) ? v?.ToString() : null, out var num) ? num : 0).ToArray();
                 }
-                if (method.DeclaringType == typeof(DateTime) && method.Name == "Parse")
+                if (parseType == typeof(DateTime))
                 {
                     return _sortOrder == SortOrder.Ascending
                         ? data.OrderBy(row => DateTime.TryParse(row.TryGetValue(dimensionName, out var v) ? v?.ToString() : null, out var dt) ? dt : DateTime.MinValue).ToArray()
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

Ivy/Views/Charts/AreaChartView.cs
```diff
@@ -165,6 +173,36 @@ public AreaChartBuilder<TSource> Toolbox()
...
     }
+
+    public AreaChartBuilder<TSource> SortBy(SortOrder order)
+    {
+        _sortOrder = order;
+        _sortKeySelector = null;
+        return this;
+    }
+
+    public AreaChartBuilder<TSource> SortBy<TKey>(
```
rorychatt: I am thinking we should have this as common functionality for all the charts

rorychatt:
image
This can be end up in a runtime error

Ivy/Views/Charts/AreaChartView.cs
```diff
@@ -165,6 +173,36 @@ public AreaChartBuilder<TSource> Toolbox()
...
+        Func<Dictionary<string, object>, TKey> keySelector,
```
rorychatt: Expression<Func<TSource, object>>? dimension = null,
    Expression<Func<IQueryable<TSource>, object>>[]? measures = null,

Maybe something like this here?

rorychatt:
-> sortBy needs to be done in the .ToPivotTable()

defymecobra:
# SortBy API — Architecture Decision

## Problem

The original approach implemented `SortBy` separately in each `ChartBuilder`, leading to:
1. Code duplication across `LineChartBuilder`, `BarChartBuilder`, `AreaChartBuilder`
2. Runtime errors when using string parsing (e.g., `int.Parse(row["Label"].ToString())`)

## Solution: Centralized Sorting in PivotTableBuilder

### Architecture

```
User API:
data.ToLineChart()
    .Dimension("Label", e => e.Label)
    .SortBy(e => int.Parse(e.Label), SortOrder.Ascending)
    .Measure(...)

Flow:
ChartBuilder.SortBy() → stores parameters
    ↓
ChartBuilder.Build() → passes to PivotTableBuilder.SortBy()
    ↓
PivotTableBuilder.ExecuteAsync() → calls ApplySorting()
    ↓
ApplySorting() → analyzes Expression, applies correct sorting
```

### Key Changes

| File | Change |
|------|--------|
| `Enums.cs` | Added `SortOrder` enum |
| `PivotTable.cs` | Added `SortBy()` methods + `ApplySorting()` with int/DateTime parsing support |
| `LineChartView.cs` | Proxy fields + methods delegating to PivotTable |
| `BarChartView.cs` | Proxy fields + methods delegating to PivotTable |
| `AreaChartView.cs` | Proxy fields + methods delegating to PivotTable |

## Why Proxy Methods in ChartBuilders?

ChartBuilders **cannot be removed** because:

1. **Internal ToPivotTable call**: `LineChartBuilder.Build()` internally calls `data.ToPivotTable()`. Users don't have access to this PivotTableBuilder instance.

2. **Fluent API expectation**: Users expect to chain methods on ChartBuilder:
   ```csharp
   data.ToLineChart()
       .Dimension(...)
       .SortBy(...)  // Must be available here
       .Measure(...)
   ```

3. **Without proxy methods**, users would need to:
   ```csharp
   // Awkward alternative:
   var pivotData = await data.ToPivotTable()
       .Dimension(...)
       .SortBy(...)
       .Measure(...)
       .ExecuteAsync();
   
   pivotData.ToLineChart();  // Would need new method
   ```

### What Proxy Methods Do

```csharp
// ChartBuilder stores parameters:
private Expression<Func<TSource, object>>? _sortSelector;
private SortOrder _sortOrder = SortOrder.None;

// ChartBuilder.Build() delegates to PivotTable:
var pivotBuilder = data.ToPivotTable().Dimension(...).Measures(...);

if (_sortOrder != SortOrder.None)
{
    pivotBuilder = _sortSelector != null
        ? pivotBuilder.SortBy(_sortSelector, _sortOrder)
        : pivotBuilder.SortBy(_sortOrder);
}
```

## Sorting Logic (Centralized)

All sorting logic is in `PivotTableBuilder.ApplySorting()`:

- Detects `int.Parse()` in Expression → uses `int.TryParse` for sorting
- Detects `DateTime.Parse()` in Expression → uses `DateTime.TryParse` for sorting
- Default: sorts by raw dimension value

image
image

defymecobra:
@claude review this

defymecobra:
@claude review this

nielsbosma (approved):

rorychatt:
(charts): refactor docs

PR #2000, opened 2026-01-11
(epic) External Widgets

nielsbosma:


Ivy/Core/ExternalWidgets/ExternalWidgetAttribute.cs
```diff
@@ -0,0 +1,35 @@
+namespace Ivy.Core.ExternalWidgets;
+
+/// <summary>
```
rorychatt, line 3: let's leave the xml docs until we implement docs for this

Ivy/Core/ExternalWidgets/ExternalWidgetRegistry.cs
```diff
@@ -0,0 +1,365 @@
...
+        catch (Exception ex)
+        {
+            Console.WriteLine($"[ExternalWidgetRegistry] Failed to scan assembly {fullName}: {ex.Message}");
+        }
+    }
+
+    private static bool IsSystemAssemblyName(string? name)
+    {
+        if (string.IsNullOrEmpty(name)) return true;
+        return name.StartsWith("System") ||
```
rorychatt, line 215: maybe there is a better way to do this

frontend/src/widgets/externalWidgetLoader.ts
dcrjodle, line 1: This file has commented-out code and a lot of unnecessary function comments

dcrjodle: It's also the only file starting with a small letter

rorychatt: not a tsx file

rorychatt (approved):

PR #2002, opened 2026-01-11
[TOC]: add API section

defymecobra:
## 🎯 Problem

After merging [PR #1883](https://github.com/Ivy-Interactive/Ivy-Framework/pull/1883) "(toc): generate compile time instead of frontend runtime", the API section stopped appearing in the Table of Contents (TOC).

### Root Cause

PR #1883 changed how TOC headings are generated:

| Before PR #1883 | After PR #1883 |
|-----------------|----------------|
| TOC scanned DOM at runtime and found ALL `<h1>`, `<h2>`, `<h3>` elements including API section | TOC receives headings as props from backend (compile-time generation in `MarkdownConverter.cs`) |

**The issue:** When switching to compile-time generation, headings for `<WidgetDocs>` block were not added. The `MarkdownConverter` only generates headings from Markdown headings (`## Title`), but `WidgetDocsView` uses `Text.H2("API")` directly which doesn't participate in headings generation.

### Part 1: Add API heading to TOC (MarkdownConverter.cs)

Added heading generation for `<WidgetDocs>` block:

```csharp
private static void HandleWidgetDocsBlock(StringBuilder codeBuilder, XElement xml, List<(string Id, string Text, int Level)>? headings)
{
    // ... existing code ...
    headings?.Add(("api", "API", 2));
}
```

### Part 2: Enable scroll navigation (WidgetDocsView.cs)

Replaced `Text.H2("API")` with `new Markdown("## API")`:

```csharp
// Before
| Text.H2("API")

// After  
| new Markdown("## API")
```

**Why this approach?**

The `Markdown` widget uses `rehype-slug` which automatically generates `id="api"` on the heading element. This allows TOC to scroll to the API section when clicked.

### Alternative Solutions Considered

| Solution | Pros | Cons | Rating |
|----------|------|------|--------|
| **A: HTML Anchor** `<span id='api'>` | Quick fix | Workaround, not clean | 2.5/5 |
| **B: Auto-generate id in TextBlockWidget** | Consistent | Unnecessary overhead for all framework users | 2/5 |
| **C: Add ElementId prop to TextBlock** | Explicit control | Requires framework changes | 3.75/5 |
| **D: Use Markdown (chosen)** | Simple, uses existing mechanism | Slight overhead for parsing | 4/5 |

We chose **Option D** because:
- Minimal code changes
- No framework core modifications
- Uses existing `rehype-slug` mechanism
- Only affects documentation, not framework users

### DIFF
--- a/Ivy.Docs.Shared/Helpers/WidgetDocsView.cs
+++ b/Ivy.Docs.Shared/Helpers/WidgetDocsView.cs
@@ -120,7 +120,7 @@ public class WidgetDocsView(string typeName, string? extensionsTypeName, string?
         string? fileName = sourceUrl != null ? System.IO.Path.GetFileName(sourceUrl) : null;
 
         return Layout.Vertical().Gap(2)
-               | Text.H2("API")
+               | new Markdown("## API")
                | (fileName != null
                    ? (Layout.Horizontal().Align(Align.Left).Gap(0) | Icons.Github.ToIcon() |
                       new Button(fileName).Link().Url(sourceUrl))


defymecobra:
@claude review this

PR #2003, opened 2026-01-12
[DocumentTools]: refactor to use backend markdown API

defymecobra:
## Problem Description
Previously, the "Copy Page" and "Download Markdown" functionalities in the `DocumentTools` component relied on client-side **DOM scraping**.

This approach had significant drawbacks:
1.  **Unreliability:** Any changes in HTML structure or styles could break the content extraction logic.
2.  **Complexity:** It required complex logic to "expand" hidden elements (`ExpandDetails`), load tabs, and reverse-engineer HTML back to Markdown.
3.  **Inaccuracy:** The generated markdown often differed from the original source (loss of formatting, extra whitespace).

## Selected Solution
I completely removed client-side scraping in favor of **fetching the raw content directly from the backend**.

The Ivy Docs server is already capable of serving raw Markdown files if the corresponding resource is requested. We updated the frontend to utilize this capability.

### Key Changes:
*   **Legacy Code Removal:** Completely removed `extractApiSections`, `extractTableWithAssociatedContent`, `loadAllTabs`, `expandDetailsElements`, and other DOM parsing utilities.
*   **New Fetching Method:** Implemented `fetchMarkdown` function which performs a `fetch` request to the current path with an appended `.md` extension (e.g., `/docs/widgets/button` -> `/docs/widgets/button.md`).
*   **Relative Path Handling:** Added a dummy base host (`http://dummy.com`) to the `URL` constructor when parsing `documentSource`. This is necessary to correctly extract the filename even if `documentSource` is a relative path (which caused errors in `new URL()` without a base).

## Benefits
*   **Stability:** We retrieve the *exact* file residing in the repository, without distortion.
*   **Performance:** No need to render all tabs and hidden sections before copying.
*   **Code Cleanliness:** Significantly simplified the `DocumentTools` component (removed over 100 lines of complex logic).

### DIFF
--- a/frontend/src/widgets/article/DocumentTools.tsx
+++ b/frontend/src/widgets/article/DocumentTools.tsx
@@ -23,573 +23,6 @@ export const DocumentTools: React.FC<DocumentToolsProps> = ({
 }) => {
   const { toast } = useToast();
 
-  // Function to load all tabs by clicking through them
-  const loadAllTabs = async (): Promise<void> => {
-    if (!articleRef.current) return;
-
-    // Find all tab elements with aria-role="tab"
-    const tabs = articleRef.current.querySelectorAll('[role="tab"]');
-
-    if (tabs.length === 0) return;
-
-    // Click through each tab and wait for content to load
-    for (const tab of Array.from(tabs)) {
-      try {
-        // Check if tab is not already selected
-        const isSelected = tab.getAttribute('aria-selected') === 'true';
-
-        if (!isSelected && tab instanceof HTMLElement) {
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

defymecobra:
Also I want to add, that this functionality doesn't work on FE (5173), but works on BE (5010) and will work on hosting. That because vite can't find endpoints for .md pages (check, you can't go to the .md page of any page too (on 5173 you can't, on 5010 you can), it doesn't redirect and thats why).

PR #2012, opened 2026-01-12
[OAuth]: Fix redirect to use same tab instead of new tab

ArtemLazarchuk:
## Changes
- Changed OAuth authentication flow to open in the same window instead of a new tab
- Replaced `client.OpenUrl()` with `client.Redirect()` for OAuth URL navigation
- Extract relative path from OAuth URL for redirect validation

## Problem
Previously, clicking OAuth login buttons opened the authentication flow in a new tab/window, which could be blocked by pop-up blockers and provides a less seamless user experience.

## Solution
- Use `client.Redirect()` instead of `client.OpenUrl()` to navigate in the same window
- Extract relative path (`PathAndQuery`) from the OAuth URL to comply with redirect validation (which only accepts relative paths or same-origin URLs)

### DIFF
--- a/Ivy/Auth/DefaultAuthApp.cs
+++ b/Ivy/Auth/DefaultAuthApp.cs
@@ -143,6 +143,7 @@ public class OAuthFlowView(AuthOption option) : ViewBase
     {
         var args = this.UseService<AppArgs>();
         var auth = this.UseService<IAuthService>();
+        var client = this.UseService<IClientProvider>();
 
         var callback = this.UseWebhook(async (request) =>
         {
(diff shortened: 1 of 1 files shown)


rorychatt (approved):

zachwolfe:
@ArtemLazarchuk This is a good idea in general, but IIRC webhook handlers are unregistered when the view context goes away (such as when redirecting to another page in the same tab). I don’t think this should work?

ArtemLazarchuk:
@zachwolfe Good point — you’re right. The handler is tied to the view context, so it will be unregistered on redirect. This approach won’t work in that case.

rorychatt:
@ArtemLazarchuk start a new branch from main as you have old main. Propose another solution

PR #2021, opened 2026-01-13
(forms): avoid infinite recursion in validation

ArtemKhvorostianyi:
Got a problem with examples with validations:

Screenshot 2026-01-13 at 14 30 43

Fixed.
Tested out other forms, nothing seems to be broken


ArtemKhvorostianyi:
TO test out: Docs -> Forms -> Validation Examples -> Collections and Arrays

PR #2025, opened 2026-01-13
(text): refactor margins logic

ArtemKhvorostianyi:


src/Ivy/Widgets/Primitives/Html.cs
```diff
@@ -13,4 +13,11 @@ public Html(string content)
     internal Html() { }
 
     [Prop] public string Content { get; set; } = string.Empty;
+
+    [Prop] public int Gap { get; set; } = 4;
```
rorychatt: have this FE default

PR #2047, opened 2026-01-15
[MetricView]: add text-overflow ellipsis to header titles

defymecobra:
## Issue Description

When the MetricView widget is narrowed, the header title text wraps to multiple lines instead of being truncated with an ellipsis (`...`). This creates an inconsistent and cluttered appearance in dashboard layouts.

### Before
image

### After
image

---

#### 1. MetricView.cs
Added `.NoWrap()` and `.Overflow(Overflow.Ellipsis)` to the header title:

```csharp
// Before
Text.H4(title).WithLayout().Grow()

// After
Text.H4(title).NoWrap().Overflow(Overflow.Ellipsis).WithLayout().Grow()
```

#### 2. lib/styles.ts
Added `minWidth: 0` to the `grow` case in `_getWantedWidth`:

```typescript
case 'grow':
  return { flexGrow: parseFloat(value) || 1, minWidth: 0 };
```

#### 3. card/styles.ts
Modified `baseHeaderClasses` to enable text truncation:

- **Removed:** `!flex` and `!flex-1` (blocked text-overflow)
- **Added:** `[&>*]:min-w-0`, `!leading-9`, `!min-w-0`, `!truncate`

---

### The Problem
CSS `text-overflow: ellipsis` requires specific conditions to work:
1. `overflow: hidden`
2. `white-space: nowrap`
3. `text-overflow: ellipsis`
4. **A constrained width** (the missing piece)

In flexbox, elements don't shrink below their content size by default. Adding `min-width: 0` allows flex items to shrink, enabling text truncation.

### Why Remove `!flex` from Headings?
The `!flex` utility made headings flex containers. `text-overflow: ellipsis` **does not work** on flex containers with direct text content — it only works on block-level elements.

### Why Add `!leading-9`?
After removing `!flex`, vertical centering was lost. Using `line-height` matching the `min-height` (both `9` = 2.25rem) provides vertical centering for single-line text.

---

## Testing Checklist

- [x] MetricView headers truncate correctly with ellipsis
- [x] Trend indicator (percentage) is vertically aligned with title
- [x] Card documentation pages display correctly
- [x] Sidebar and other layouts using `Grow()` work properly
- [x] Dashboard demo page works as expected

### DIFF
--- a/src/Ivy/Views/Dashboards/MetricView.cs
+++ b/src/Ivy/Views/Dashboards/MetricView.cs
@@ -56,7 +56,7 @@ public class MetricView(
         return new Card(
                 content: Text.P(x.MetricFormatted).Large().NoWrap().Overflow(Overflow.Clip),
                 header: Layout.Horizontal().Align(Align.Center)
-                    | Text.H4(title).NoWrap().Overflow(Overflow.Ellipsis).WithLayout().Grow()
+                    | Text.H4(title).NoWrap().Overflow(Overflow.Ellipsis).Width(Size.Grow())
                     | (Layout.Horizontal().Align(Align.Right).Gap(1).Width(Size.Fit())
                         | (x.TrendComparedToPreviousPeriod != null ? x.TrendComparedToPreviousPeriod >= 0
                                 ? Icons.TrendingUp.ToIcon().Color(Colors.Success).Small()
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
@defymecobra I think there is supposed to be an icon in those headers on the right.

We need to make it so the icon remains displayed, and title text to the left of it overflows (with NoWrap?)

defymecobra:
Changes:
- **MetricView.cs**:
  - Replaced `.WithLayout().Grow()` with `.Width(Size.Grow())` on the header title. This removes the unnecessary wrapper div that was causing layout conflicts and preventing correct text truncation on the left side.
  - **Restored the missing metric icon** in the success state header, positioning it to the right of the trend indicator.
  
 ## Screenshot (high-scale)
 ### Before:
image

### After:
image

defymecobra:
@claude review this

PR #2055, opened 2026-01-16
(expandable): disable cursor for disabled variant

ArtemKhvorostianyi:
Had a problem:
Screenshot 2026-01-16 at 11 54 32
On disabled state cursor still was active.
Fixed:
https://github.com/user-attachments/assets/2d10d528-819a-4d11-a843-8aad212b7c66


rorychatt:
Put a switch inside of it (like in ivy db generate)
Is the switch active? It should not get the cursor disabled and be clickable

rorychatt:
https://github.com/user-attachments/assets/175846ad-ef51-4b9f-94cf-33222e19e4c5

can not toggle selects inside expandable

PR #2068, opened 2026-01-18
[OAuth]: implement same-tab OAuth flow

defymecobra:
## 🐛 The Issue
Previously, initiating an OAuth login (e.g., GitHub) forced the authorization page to open in a new browser tab.
This resulted in:
- **Fragmented UX:** Users lost context of their original session.
- **Connection Loss:** The original application tab often disconnected or became stale.
- **Mobile Issues:** Poor experience on mobile devices where tab switching is cumbersome.

## 🔍 Root Cause Analysis
My investigation revealed two primary architectural causes:

1.  **Frontend Constraint (`ButtonWidget`):**
    The `ButtonWidget.tsx` component was hardcoded to explicitly set `target="_blank"` for any URL-based button, preventing same-tab navigation even if desired.

2.  **Lifecycle Dependency (`UseWebhook`):**
    The backend relied on `UseWebhook` for handling the OAuth callback. Webhooks in Ivy are bound to the specific **View ID** and its lifecycle.
    - When navigating away in the *same tab* (to GitHub), the current View (e.g., login page) is unmatched/unmounted by the router.
    - This destroys the View instance and unregisters the Webhook handler.
    - When GitHub redirects back, the Webhook no longer exists, resulting in a 404 error or lost callback handling.

## 🛠 Implemented Solution
I implemented a **Stateless / Global Callback Registry** pattern to decouple authentication from the View lifecycle.

### Key Components:

1.  **`OAuthCallbackRegistry` (New Service):**
    A singleton service that temporarily stores pending OAuth requests, mapping a unique `state` parameter to the user's `connectionId`. It handles the correlation between the outgoing request and the incoming callback without relying on a specific View instance.

2.  **`AuthController` Updates:**
    - Added a new endpoint `/ivy/auth/oauth-callback` to handle the return from the provider directly.
    - It retrieves the pending state from the Registry.
    - Performs the token exchange using `IAuthProvider` directly.
    - Sets authentication cookies (JSON serialized `AuthToken`) in the HTTP response.
    - Redirects the user seamlessly back to the app root (`/`).

3.  **Widget Enhancements:**
    Updated `Button.cs` (backend) and `ButtonWidget.tsx` (frontend) to support a new `Target` property.
    - `LinkTarget.Self`: Renders `target="_self"`.
    - `LinkTarget.Blank`: Renders `target="_blank"` (default behavior maintained).

4.  **Refactoring:**
    Refactored cookie handling logic in `CookieRegistryExtensions` to be reusable, ensuring consistent security settings (HttpOnly, Secure, SameSite) across the codebase.

## 💡 Why this Approach?
We chose this "Global Registry" approach over other options (like persisting Views or using distributed caching immediately) because:

- **Simplicity:** It requires minimal changes to the core architecture.
- **Robustness:** It completely eliminates the "vanishing webhook" race condition.
- **Performance:** Fast, in-memory validaton of callbacks.
- **Future-Proof:** The `IOAuthCallbackRegistry` interface can be easily backed by Redis later if we move to a distributed server farm.

## ✅ Verification
- Verified manually using the `OAuthTestApp`.
- Confirmed GitHub login now initiates and completes entirely within a single browser tab.
- Confirmed correct session restoration and cookie setting upon return.

https://github.com/user-attachments/assets/274d1b45-fcb8-4584-bd0d-9b658b84dee4

### DIFF
--- a/src/Ivy/Auth/AuthController.cs
+++ b/src/Ivy/Auth/AuthController.cs
@@ -1,6 +1,7 @@
 using Ivy.Apps;
 using Ivy.Client;
 using Ivy.Core;
+using Ivy.Helpers;
 using Ivy.Hooks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -85,6 +86,7 @@ public class AuthController() : Controller
         [FromQuery(Name = "error_description")] string? errorDescription,
         [FromServices] IOAuthCallbackRegistry registry,
         [FromServices] IAuthProvider authProvider,
+        [FromServices] AppSessionStore sessionStore,
         [FromServices] ILogger<AuthController> logger)
     {
         var effectiveId = callbackId ?? state;
@@ -110,7 +112,15 @@ public class AuthController() : Controller
 
         try
         {
-            var tempSession = new AuthSession(new HttpClientHandler());
+            // Get the session and its HttpMessageHandler using the connectionId from the pending callback
+            if (!sessionStore.Sessions.TryGetValue(pending.ConnectionId, out var appSession))
+            {
+                logger.LogWarning("OAuth callback failed: Session not found for connection {ConnectionId}", pending.ConnectionId);
+                return BadRequest("Authentication error: app session expired");
+            }
+
+            var httpMessageHandler = appSession.AppServices.GetRequiredService<HttpMessageHandler>();
+            var tempSession = AuthHelper.GetAuthSession(HttpContext, httpMessageHandler);
 
             var token = await authProvider.HandleOAuthCallbackAsync(tempSession, HttpContext.Request);
 
--- a/src/Ivy/Auth/DefaultAuthApp.cs
+++ b/src/Ivy/Auth/DefaultAuthApp.cs
@@ -144,6 +144,7 @@ public class OAuthFlowView(AuthOption option) : ViewBase
     {
         var args = this.UseService<AppContext>();
         var registry = this.UseService<IOAuthCallbackRegistry>();
+        var auth = this.UseService<IAuthProvider>();
 
         var state = this.UseState(() => registry.RegisterPending(args.ConnectionId, option.Id ?? ""));
 
@@ -156,6 +157,7 @@ public class OAuthFlowView(AuthOption option) : ViewBase
             .Icon(option.Icon)
             .Width(Size.Full())
             .Url(oauthUriBuilder.ToString())
-            .Target(LinkTarget.Self);
+            .Target(LinkTarget.Self)
+            .OpenInNewTab(auth.OpenOAuthLoginInNewTab);
     }
 }
--- a/src/Ivy/Auth/IAuthProvider.cs
+++ b/src/Ivy/Auth/IAuthProvider.cs
@@ -27,4 +27,6 @@ public interface IAuthProvider
     Task<AuthToken?> HandleOAuthCallbackAsync(IAuthSession authSession, HttpRequest request, CancellationToken cancellationToken = default);
 
     Task<TokenLifetime?> GetAccessTokenLifetimeAsync(IAuthSession authSession, CancellationToken cancellationToken = default);
+
+    bool OpenOAuthLoginInNewTab => false;
 }
--- a/src/Ivy/Widgets/Button.cs
+++ b/src/Ivy/Widgets/Button.cs
@@ -205,5 +205,5 @@ public static class ButtonExtensions
     public static Button Target(this Button button, LinkTarget target) => button with { Target = target };
 
     [RelatedTo(nameof(Button.Target))]
-    public static Button OpenInNewTab(this Button button) => button with { Target = LinkTarget.Blank };
+    public static Button OpenInNewTab(this Button button, bool openInNewTab = true) => button with { Target = openInNewTab ? LinkTarget.Blank : LinkTarget.Self };
 }
--- a/src/auth/Ivy.Auth.Clerk/ClerkAuthProvider.cs
+++ b/src/auth/Ivy.Auth.Clerk/ClerkAuthProvider.cs
@@ -31,6 +31,8 @@ public class ClerkAuthProvider : IAuthProvider
     private readonly bool _isProduction;
     private string? _origin = null;
 
+    public static bool OpenOAuthLoginInNewTab => true;
+
     private static (bool IsProduction, string Key) ParseKey(string name, string type, string key)
     {
         var tokens = key.Split('_', 3);


defymecobra:
@claude review this

rorychatt:
@defymecobra I am not sure if it makes sense to add another property to Button API due to Auth stuff. Feels like it is at least the wrong place.

Leave this issue with me. I will try and solve it myself. Maybe I will just edit your branch a bit

src/Ivy/Auth/AuthController.cs
```diff
@@ -73,6 +74,70 @@ public async Task<IActionResult> OAuthLogin(
...
+    [Route("ivy/auth/oauth-callback")]
+    [HttpGet]
+    public async Task<IActionResult> OAuthCallback(
+        [FromQuery] string? code,
+        [FromQuery] string? state,
+        [FromQuery] string? error,
+        [FromQuery(Name = "error_description")] string? errorDescription,
+        [FromServices] IOAuthCallbackRegistry registry,
+        [FromServices] IAuthProvider authProvider,
+        [FromServices] ILogger<AuthController> logger)
```
zachwolfe: This callback must support paths like `ivy/auth/oauth-callback/{callback-id}` as well (see `WebhookController`).

src/Ivy/Auth/AuthController.cs
```diff
@@ -51,14 +51,15 @@ public async Task<IActionResult> OAuthLogin(
...
+        // Construct the OAuth callback endpoint
         var scheme = HttpContext.Request.Scheme;
         if (HttpContext.Request.Headers.TryGetValue("X-Forwarded-Proto", out var forwardedProto))
         {
             scheme = forwardedProto.ToString();
         }
         var host = HttpContext.Request.Host.Value ?? throw new InvalidOperationException("Host not found in request");
-        var callback = new WebhookEndpoint(callbackId, scheme, host);
+        var callbackBaseUrl = $"{scheme}://{host}/ivy/auth/oauth-callback";
+        var callback = new WebhookEndpoint(callbackId, callbackBaseUrl);
```
zachwolfe: The `WebhookEndpoint` type is now being used for non-webhook callbacks. Maybe we should rename it and move it out of `UseWebhook.cs`?

src/Ivy/Auth/AuthController.cs
```diff
@@ -73,6 +74,70 @@ public async Task<IActionResult> OAuthLogin(
         }
     }
 
+    [Route("ivy/auth/oauth-callback")]
+    [HttpGet]
+    public async Task<IActionResult> OAuthCallback(
+        [FromQuery] string? code,
+        [FromQuery] string? state,
+        [FromQuery] string? error,
+        [FromQuery(Name = "error_description")] string? errorDescription,
```
zachwolfe, line 86: The Clerk provider does not receive `code`, `error` or `error_description`* in its callback, so these should not be relied upon in general.

*Or `state`, but that is covered by my comment about accepting callback ids passed as part of the path.

defymecobra:
# OAuth Flow Refactoring & Improvements

Description of the latest changes regarding the OAuth Same-Tab flow refactoring.

## 1. [UseWebhook]: rename WebhookEndpoint to CallbackEndpoint

**What was done:**
- Renamed `WebhookEndpoint` class to `CallbackEndpoint`.
- Added `BuildBaseUrl` method for convenient URL construction.

**Why:**
The class is used not only for webhooks but also for handling OAuth callbacks. The name `CallbackEndpoint` more accurately reflects its universal purpose and eliminates terminology confusion.

## 2. [AuthController]: refactor OAuth callback handling

**What was done:**
- Added a new route: `/ivy/auth/oauth-callback/{callbackId}`.
- Implemented ID resolution logic: `effectiveId = callbackId ?? state`.
- Made `state`, `code`, and `error` parameters optional.

**Why:**
It was necessary to separate the logic for regular webhooks (default `/ivy/webhook`) and OAuth callbacks. The new route provides explicit separation. The `effectiveId` logic supports providers that pass IDs via the URL path and those that use the `state` parameter.

## 3. [Docs]: update auth callback urls and rename WebhookEndpoint

**What was done:**
- Updated `UseWebhook.md` to reflect the renaming to `CallbackEndpoint`.
- Updated documentation for all authentication providers (Google, GitHub, Microsoft Entra, Auth0, Supabase, etc.).
- Changed callback URLs in examples to `/ivy/auth/oauth-callback`.

**Why:**
Documentation must match the code changes (item 2). Specifying correct URLs is critical for developers configuring their applications.

## 4. [example]: add OAuthTestApp & simplified UI

**What was done:**
- Implemented `OAuthTestApp` testing application.
- Added logic to fetch user data via `GetUserInfoAsync`.
- Simplified UI to an "Authentication Successful!" status displaying avatar, name, and email.

**Why:**
The application is needed to verify the new OAuth flow functionality. The simplified UI allows for instant visual confirmation of successful authentication and data retrieval without unnecessary steps (Login clicked -> Profile received).

defymecobra:
@claude review this

src/Ivy/Hooks/UseWebhook.cs
```diff
@@ -12,45 +12,47 @@ public static class UseWebhookExtensions
...
+public record CallbackEndpoint(string Id, string BaseUrl)
 {
-    public WebhookEndpoint(string id, string scheme, string host) : this(id, BuildBaseUrl(scheme, host))
+    public CallbackEndpoint(string id, string scheme, string host) : this(id, BuildBaseUrl(scheme, host))
     {
     }
 
     public static string BuildBaseUrl(string scheme, string host) => $"{scheme}://{host}/ivy/webhook";
 
+    public static string BuildBaseUrl(string scheme, string host, string path) => $"{scheme}://{host}{path}";
```
zachwolfe: Looks like this new API is not called anywhere. Did you mean to use it?

src/Ivy/Hooks/UseWebhook.cs
```diff
@@ -12,45 +12,47 @@ public static class UseWebhookExtensions
...
 
         context.UseEffect(() => webhookController.Register(webhookId.Value, handler), [EffectTrigger.OnMount()]);
 
-        return new WebhookEndpoint(webhookId.Value, args.Scheme, args.Host);
+        return new CallbackEndpoint(webhookId.Value, args.Scheme, args.Host);
     }
 }
 
-public record WebhookEndpoint(string Id, string BaseUrl)
+public record CallbackEndpoint(string Id, string BaseUrl)
```
zachwolfe: This should arguably be moved out of UseWebhook.cs

src/Ivy/Auth/DefaultAuthApp.cs
```diff
@@ -143,20 +143,21 @@ public class OAuthFlowView(AuthOption option) : ViewBase
...
+        var registry = this.UseService<IOAuthCallbackRegistry>();
 
-        var callback = this.UseWebhook(async (request) =>
-        {
-            var token = await auth.HandleOAuthCallbackAsync(request);
-            return new RedirectResult("/");
-        });
+        var state = this.UseState(() => registry.RegisterPending(args.ConnectionId, option.Id ?? ""));
+
+        var callbackUrl = $"{args.Scheme}://{args.Host}/ivy/auth/oauth-callback";
```
zachwolfe: I think this variable can be removed.

src/examples/OAuthTestApp/Program.cs
```diff
@@ -0,0 +1,34 @@
+using Ivy;
+using Ivy.Auth.GitHub;
+
+// Создаём сервер
```
zachwolfe: These comments should probably be in English for consistency

src/Ivy.Docs.Shared/Docs/01_Onboarding/03_CLI/04_Authentication/02_Supabase.md
```diff
@@ -60,7 +60,7 @@ To customize your authentication settings:
 2. **Click "URL Configuration"**
 3. **Configure**:
    - **Site URL**: Your application's URL (e.g., `http://localhost:5010`)
-   - **Redirect URLs**: Add your callback URLs (e.g., `http://localhost:5010/ivy/webhook/*`)
+   - **Redirect URLs**: Add your callback URLs (e.g., `http://localhost:5010/ivy/auth/oauth-callback/*`)
```
zachwolfe: After seeing it in the docs like this, `auth/oauth-callback` feels redundant (i.e. all auth callbacks are likely to be OAuth callbacks). Perhaps it should be renamed to `auth/callback`?

zachwolfe (commented):
Thanks for the changes @defymecobra. I have a few more comments below.

zachwolfe:
@rorychatt
> I am not sure if it makes sense to add another property to Button API due to Auth stuff. Feels like it is at least the wrong place.

In fairness, I think links that open in the current tab could be useful in non-auth cases as well.

rorychatt:
@zachwolfe as always, good point. I created a separate issue for this as it is not auth related, I guess.

src/examples/OAuthTestApp/Program.cs
rorychatt: @defymecobra 
_and remember, no Russian_

defymecobra: yes, sorry. I didn't see that when was going to pull

rorychatt:
@defymecobra fix merge conflicts.
Also redirect in same tab should be the default behavior. Users should however be able to override it, make sure in docs we have a sentence about how to do this.
And make sure that it works

defymecobra:
## Code Cleanup (zachwolfe feedback)

### Changes Made

#### 1. Removed Unused `BuildBaseUrl` Overload
- **File:** `src/Ivy/Hooks/UseWebhook.cs`
- Removed `BuildBaseUrl(scheme, host, path)` method that was never called

#### 2. Moved `CallbackEndpoint` to `IAuthProvider.cs`
- **From:** `src/Ivy/Hooks/UseWebhook.cs`
- **To:** `src/Ivy/Auth/IAuthProvider.cs`
- More logical location since it's primarily used for OAuth callbacks
- Updated `using` statements in affected files

#### 3. Removed Unused Variable
- **File:** `src/Ivy/Auth/DefaultAuthApp.cs`
- Removed unused `callbackUrl` variable

#### 4. Renamed OAuth Callback Route
- **From:** `/ivy/auth/oauth-callback`
- **To:** `/ivy/auth/callback`
- Updated in `AuthController.cs` and all auth provider documentation

---

## Same-Tab Default Behavior (rorychatt feedback)

### Changes Made

#### 1. Changed Button URL Default to Same-Tab
- **Files:** 
  - `src/Ivy/Widgets/Button.cs` - Changed `LinkTarget` default from `Blank` to `Self`
  - `src/frontend/src/widgets/button/ButtonWidget.tsx` - Changed `target` default from `'Blank'` to `'Self'`

#### 2. Added Override Method
- Added `.OpenInNewTab()` extension method to `Button.cs`
- Allows explicit new-tab behavior when needed

#### 3. Updated Documentation
- **File:** `src/Ivy.Docs.Shared/Docs/02_Widgets/03_Common/01_Button.md`
- Updated "Buttons with URLs" section to reflect same-tab default
- Added example using `.OpenInNewTab()`

Before: 
image
After:
image

### Why Both Files?
The frontend (`ButtonWidget.tsx`) has its own default value. When the backend doesn't serialize a property (because it matches the default), the frontend uses its own default. Both needed to be synchronized.

defymecobra:
Created CallbackEndpoint.cs in src/Ivy/Core

rorychatt:
oh fuck I did not mean to merge this yet

PR #2072, opened 2026-01-19
(Input): Add Surface color to Inputs

dcrjodle:
Dark Theme
Screenshot 2026-01-20 at 16 30 42
Screenshot 2026-01-20 at 16 30 19

Light Theme
Screenshot 2026-01-20 at 16 30 48
Screenshot 2026-01-20 at 16 30 11


nielsbosma (changes requested):
Should we really remove shade every where?

When we make this theme editor we will make it configurable. Then it will be pain to readd all shows.

Maybe better to keep the classes but set shadow to 0

PR #2103, opened 2026-01-22
[Kanban]: prevent premature selector compilation

defymecobra:
## Problem
A "premature compilation" issue was identified in the `Kanban` component.
Calling the `ToKanban(...)` extension method caused immediate compilation of the passed expressions (selectors) because the `KanbanBuilder` constructor expected parameters of type `Func<...>`, not `Expression<Func<...>>`.

### Root Cause
The code looked like this:
```csharp
public static KanbanBuilder<...> ToKanban<...>(..., Expression<Func<...>> groupBySelector, ...)
{
    // Compilation happened here, immediately
    return new KanbanBuilder<...>(..., groupBySelector.Compile(), ...);
}
```
Calling `.Compile()` is an expensive operation. Since it was in the static factory method, compilation happened **always** when creating the builder object, even if:
1.  The builder is never used (e.g., created inside a conditional block that isn't rendered).
2.  Expressions haven't changed and could theoretically be cached.
3.  The application is just initializing the UI structure but hasn't proceeded to rendering yet.

This created unnecessary CPU load (JIT/IL generation) in hot UI creation paths.

## Solution
An architectural decision was made to **defer compilation** until the moment the component is actually built (`Build()` method).

### Changes Made
1.  **Refactoring `KanbanBuilder`:**
    -   The primary constructor now accepts `Expression<Func<...>>` instead of `Func<...>`.
    -   Class fields changed to store `Expression`.
    -   Configuration methods (e.g., `.CardOrder()`) now store expressions without compiling.

2.  **Deferred Compilation:**
    -   Calls to `.Compile()` for necessary selectors were added to the `Build()` method, directly before executing data queries (LINQ).

3.  **Updating `KanbanExtensions`:**
    -   Calls to `.Compile()` were removed from the `ToKanban` method. Expressions are passed to the constructor as is.

### Conclusion
Now compilation happens lazily and only when truly necessary to display the Kanban board. The external API for developers remains unchanged (source compatible for most scenarios), but the internal implementation has become more performant and logically correct. Kanban works as before unchaged.

### DIFF
--- a/src/Ivy/Views/Kanban/KanbanBuilder.cs
+++ b/src/Ivy/Views/Kanban/KanbanBuilder.cs
@@ -50,7 +50,7 @@ public class KanbanBuilder<TModel, TGroupKey>(
         _columnOrderBySelector = Expression.Lambda<Func<TModel, object?>>(
             Expression.Convert(orderBySelector.Body, typeof(object)),
             orderBySelector.Parameters);
-        
+
         _columnOrderDescending = descending;
         return this;
     }
@@ -60,7 +60,7 @@ public class KanbanBuilder<TModel, TGroupKey>(
         _cardOrderBySelector = Expression.Lambda<Func<TModel, object?>>(
             Expression.Convert(orderBySelector.Body, typeof(object)),
             orderBySelector.Parameters);
-        
+
         _cardOrderDescending = descending;
         return this;
     }


rorychatt:
@defymecobra test this with all the kanban examples you have. Let me know, and I will then test this in one of the apps we have.

rorychatt:
also fix red

defymecobra:
> also fix red

image
image

It seems that these reds are not connected to my changes. I've tried to re-run jobs - nothing changed

defymecobra:
> @defymecobra test this with all the kanban examples you have. Let me know, and I will then test this in one of the apps we have.

I've tested Kanban functionality in Samples. Everything works correctly

PR #2113, opened 2026-01-24
[Docs]: preserve links in markdown headings during conversion

defymecobra:
## Problem

Links inside documentation headings were not displayed on the page. For example:

```markdown
## Overview [hook](../02_RulesOfHooks.md)
```

Should have rendered as "Overview hook" where "hook" is a clickable link, but instead the word "hook" either disappeared or displayed as plain text without the link functionality.

## Root Cause

The issue was located in the backend, in `src/Ivy.Docs.Tools/MarkdownConverter.cs`.

When processing headings (`HeadingBlock`), the code extracted **only** `LiteralInline` nodes (plain text) and completely ignored all other inline elements, including **links (`LinkInline`)**:

```csharp
// OLD CODE (lines 278-285)
var headingText = new StringBuilder();
foreach (var inline in hBlock.Inline.Descendants())
{
    if (inline is LiteralInline literal)  // <-- Only LiteralInline!
    {
        headingText.Append(literal.Content.ToString());
    }
}
// LinkInline and other types were COMPLETELY IGNORED

sectionBuilder.AppendLine($"{new string('#', hBlock.Level)} {headingText.ToString().Trim()}");
```

As a result, the heading `## Overview [hook](../02_RulesOfHooks.md)` was converted to `## Overview hook` (plain text without the link).

## Solution

Instead of "reconstructing" the heading from individual `LiteralInline` elements, we now use the **original raw markdown** of the heading directly from the source content:

```csharp
// NEW CODE
string rawHeadingMarkdown = markdownContent.Substring(hBlock.Span.Start, hBlock.Span.Length).Trim();

sectionBuilder.AppendLine();
sectionBuilder.AppendLine(rawHeadingMarkdown);

// Text-only version is still extracted for TOC ID generation
if (headings != null)
{
    var headingText = new StringBuilder();
    foreach (var inline in hBlock.Inline.Descendants())
    {
        if (inline is LiteralInline literal)
        {
            headingText.Append(literal.Content.ToString());
        }
    }
    var text = headingText.ToString().Trim();
    var id = GenerateHeadingId(text);
    headings.Add((id, text, hBlock.Level));
}
```

### Why This Solution

1. **Preserves all formatting** — links, bold, italic and other inline elements are now correctly preserved
2. **Minimal changes** — doesn't require rewriting the inline element processing logic
3. **Backward compatible** — TOC IDs are still generated from the text-only version of the heading

## Side Effects

After applying the fix, a previously hidden bug was discovered: a broken link in `UseRefreshToken.md` pointing to a non-existent file `./07_EventHandlers.md`. Previously, this link was silently "swallowed" during processing; now it's correctly processed and causes a build error if the target file doesn't exist.

**Fix:** Updated the path to the correct location `../../01_Onboarding/02_Concepts/05_EventHandlers.md`.

## Changed Files

| File | Change |
|------|--------|
| `src/Ivy.Docs.Tools/MarkdownConverter.cs` | Fixed heading processing logic |
| `src/Ivy.Docs.Shared/Docs/03_Hooks/02_Core/16_UseRefreshToken.md` | Fixed broken path to EventHandlers.md |

## Before:
image

## After: 
image

### DIFF
--- a/src/Ivy.Docs.Tools/MarkdownConverter.cs
+++ b/src/Ivy.Docs.Tools/MarkdownConverter.cs
@@ -287,6 +287,17 @@ public static partial class MarkdownConverter
                     sectionBuilder.AppendLine();
                     sectionBuilder.AppendLine($"{new string('#', hBlock.Level)} {headingText.ToString().Trim()}");
 
+                    // Extract heading for TOC if not inside details/demo block (which are handled recursively or separately and shouldn't appear in main TOC usually unless properly implemented)
+                    // Note: We are currently inside HandleBlocks which iterates over top-level blocks or nested blocks.
+                    // If isNestedContent is true, it means we are inside a custom block (like Details), so we might NOT want to add to main headings list?
+                    // The requirement is to avoid "leaked headings from demo-below elements".
+                    // Demo-below elements are handled in HandleDemoCodeBlock which generates a Box with content.
+                    // Regular headings are HeadingBlock.
+
+                    // IF we are processing the main document (headings != null), we add the heading.
+                    // If we are inside a nested block (like Details body), headings might be passed or null depending on intent.
+                    // For now, let's assume we only collect top level headings or headings where 'headings' list is provided.
+
                     if (headings != null)
                     {
                         var text = headingText.ToString().Trim();


defymecobra:
@claude review this

src/Ivy.Docs.Shared/Docs/03_Hooks/02_Core/16_UseRefreshToken.md
```diff
@@ -93,7 +93,7 @@ Return values can be any type, including complex objects like records or classes
 | `IsRefreshed` | `bool` | `true` if the token has been refreshed at least once |
 | `ReturnValue` | `object?` | The value passed to the last `Refresh()` call |
 
-## Refresh Tokens vs [Event Handlers](./07_EventHandlers.md)
+## Refresh Tokens vs [Event Handlers](../../01_Onboarding/02_Concepts/05_EventHandlers.md)
```
rorychatt: I in general think that we should not support links in headers.

meaning we should not parse this.

Fix is to just remove link and have this

```text
## Refresh Tokens vs Event Handlers

defymecobra:
Done. Reverted unnecessary cnanges and fixed header in UseRefreshToken

PR #2117, opened 2026-01-25
[Calendar]: replace dropdowns with editable month/year input

defymecobra:
## Issue Description
The standard navigation controls (`captionLayout="dropdown"`) provided by `react-day-picker` were not aligning with the desired UX for date selection.
- **Problem 1**: The native dropdowns for Month and Year were bulky and didn't fit the compact, modern design requirements.
- **Problem 2**: In the Range selection mode, custom headers were sometimes overridden or not receiving the correct date props, leading to rendering inconsistencies (e.g., duplicated months or empty titles).
- **Goal**: Implement a unified, editable `[ Month / Year ]` input directly in the calendar header that supports both navigation (arrows) and direct manual entry.

## Solution

I implemented a custom `MonthYearInput` component integrated directly into the core `Calendar` UI.

### 1. Core Logic (`calendar.tsx`)
- **Custom Component**: Created `MonthYearInput` to replace the default `CaptionLabel`.
- **Display Logic**: Implemented a robust parsing strategy: since the `displayMonth` prop was not consistently available in the used library context, the component extracts the initial date by parsing the localized title string (`props.children`).
- **Interaction**: The input allows users to type a month (1-12) and year. On `blur` or `Enter`, it triggers the `useDayPicker().goToMonth()` method to update the view.
- **Style**: Adjusted `pointer-events` in the navigation container to ensure the input is clickable while maintaining the functionality of the standard "Previous/Next" arrow buttons.

### 2. Global Integration
- **Refactoring**: Removed `captionLayout="dropdown"` and ad-hoc header customizations from all consumer widgets:
  - `DateRangeInputWidget`
  - `DateVariant` (Single Date Input)
  - `DateTimeVariant` (Date + Time Input)
- **Result**: By removing local overrides, all date widgets now default to the standard layout which uses our new custom `MonthYearInput`, ensuring a consistent experience across the entire application.

## Why this approach?
- **Better UX**: Combines the precision of manual input (jumping to "10/2030") with the ease of visual navigation.
- **Maintainability**: The logic is centralized in one file (`calendar.tsx`). If the calendar library is updated or styles change, we only need to modify one component instead of updating every widget.
- **Reliability**: The parsing fallback ensures the header works correctly even if the underlying library's prop passing behavior varies between view modes (Single vs Range).

## Demo

https://github.com/user-attachments/assets/04603c82-e92c-4635-aa86-4e8887160a95

### DIFF
--- a/src/frontend/src/widgets/inputs/DateRangeInputWidget.tsx
+++ b/src/frontend/src/widgets/inputs/DateRangeInputWidget.tsx
@@ -10,9 +10,12 @@ import {
 import { cn } from '@/lib/utils';
 import { CalendarIcon, X } from 'lucide-react';
 import {
+  addMonths,
   endOfMonth,
   endOfYear,
   format,
+  isBefore,
+  isSameMonth,
   startOfMonth,
   startOfYear,
   subDays,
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
image


In this mode, we need to be able to separately move selection on the left, and selection on the right.

curretnly by using arrows you move both selections. What if user wants to pick a gap for like 3 months?


```text
----

<---- June 06/2024 ---->  | <---- May 05/2025 ---->


you are missing this ^  and  ^ this arrow
```

defymecobra:
Previously, both calendars in DateRangeInput shared a single navigation state, making it impossible to view months far apart.
Now each calendar has its own navigation arrows. When one calendar would overlap the other, both move together

https://github.com/user-attachments/assets/0088227e-07ee-4eb7-8471-31cab124fce6

rorychatt:
Screenshot 2026-01-29 at 11 26 23
we also need to fix this. But let's do this in a separate PR

PR #2122, opened 2026-01-26
feat: add internal ChatBox.cs

dcrjodle:
## Summary
- Add `BorderColor` property to `Box` widget 
- Add `BorderOpacity` property to `Box` widget 
- Adjust small button height from `h-8` to `h-6`
- Remove background color from outline button variant
Screenshot 2026-01-26 at 18 07 06
Screenshot 2026-01-26 at 18 07 03

### DIFF
--- a/src/Ivy/Widgets/Internal/ChatBox.cs
+++ b/src/Ivy/Widgets/Internal/ChatBox.cs
@@ -4,7 +4,7 @@ using Ivy.Views;
 // ReSharper disable once CheckNamespace
 namespace Ivy;
 
-public class QuestionBox(string? title = null, object? content = null) : ViewBase
+public class ChatBox(string? title = null, object? content = null) : ViewBase
 {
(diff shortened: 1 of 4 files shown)


src/Ivy/Widgets/Primitives/Box.cs
```diff
@@ -25,6 +22,8 @@ internal Box()
 
     [Prop] public BorderStyle BorderStyle { get; set; } = BorderStyle.Solid;
 
+    [Prop] public string? BorderColor { get; set; }
```
nielsbosma: This should be the enum Color

src/Ivy/Widgets/Internal/ChatBox.cs
rorychatt, line 1: Maybe this should have been moved to Ivy/Ivy.Console.UI or something (this had been moved to a new place)

PR #2124, opened 2026-01-26
feat: theme-aware input background colors

dcrjodle:
## Summary
- Remove `bg-transparent` from input/textarea components to allow CSS background styling
- Add subtle shadows for light theme with reduced opacity values
- Disable shadows entirely in dark theme
- Add 10% black background tint to inputs in dark theme for better visibility
Screenshot 2026-01-27 at 10 54 49
Screenshot 2026-01-27 at 10 54 41
Screenshot 2026-01-27 at 10 54 36
Screenshot 2026-01-27 at 10 54 23

### DIFF
--- a/src/frontend/src/index.css
+++ b/src/frontend/src/index.css
@@ -444,50 +444,6 @@ ivy-widget {
   }
 }
 
-/* Input background styling - dark theme only */
-@layer base {
-  .dark input,
-  .dark textarea,
-  .dark select {
-    background-color: color-mix(
-      in srgb,
-      var(--black) 15%,
-      var(--background) 10%
-    );
-  }
-
-  /* SelectTrigger (dropdown), DateRange button - same bg color-mix */
-  .dark button[role='combobox'],
-  .dark button[data-slot='calendar'] {
-    background-color: color-mix(
-      in srgb,
-      var(--black) 15%,
-      var(--background) 10%
-    ) !important;
-  }
-
-  .dark button[role='combobox']:hover,
-  .dark button[data-slot='calendar']:hover {
-    background-color: color-mix(
-      in srgb,
-      var(--black) 8%,
-      var(--background) 10%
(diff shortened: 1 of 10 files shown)


src/frontend/src/index.css
```diff
@@ -24,24 +24,27 @@ ivy-widget {
...
+    --shadow-md: 0 2px 4px 0 hsl(0 0% 0% / 0.05);
+    --shadow-lg: 0 4px 6px 0 hsl(0 0% 0% / 0.05);
+    --shadow-xl: 0 6px 10px 0 hsl(0 0% 0% / 0.05);
+    --shadow-2xl: 0 8px 14px 0 hsl(0 0% 0% / 0.08);
   }
 
   .dark {
     /* Additional dark theme overrides not in design system yet */
     --toolbox: #edeaf1;
+    --shadow-2xs: none;
```
rorychatt, line 40: so no shadows in dark theme only?

dcrjodle: Yes following shadcn examples, and applying bg color to dark theme inputs only. In light theme these bg colors look too dark, and this allows us to be more flexible with ivy-studio design

rorychatt:
https://github.com/user-attachments/assets/0197404a-76c2-437e-8556-6d0087323223


We need to sync all the inputs to have the same on-hover behavior, same bg fill and so on. Right now, all of them are different. And then also the code inputs and stuff like that now look completely odd to me.  See this app

http://localhost:5173/concepts/form
https://samples.ivy.app/concepts/form

rorychatt:
Screenshot 2026-01-26 at 19 26 56

it might have been your other PR but now we also have this

dcrjodle:
@rorychatt there is no change to the hover effects compared to production. But I can look at adding the bg color to select inputs as well

I also don't see a difference between code inputs in this branch and in prod

PR #2127, opened 2026-01-27
(upload): remove file size limit for uploads

ArtemKhvorostianyi:
Tested file uploads up to 350 MB

https://github.com/user-attachments/assets/60d71e21-ba85-4121-8bc6-114d2aae40c5

### DIFF
--- a/src/Ivy/Server.cs
+++ b/src/Ivy/Server.cs
@@ -17,7 +17,6 @@ using Ivy.Views.DataTables;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http; //do not remove - used in RELEASE
-using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;


rorychatt:
@ArtemKhvorostianyi 
So we set up max file upload limit here, in server.cs?

Figure out how users should configure this.
In this case, maybe it makes sense you can't upload extra large files?
Maybe solution is to add a ref to docs of file input/upload, that in order to support big file upload you need to do x. ?

ArtemKhvorostianyi:
> maybe solution is not to fix it like this(we don't want people uploading mega files everywhere) but explaining how others can configure this in docs

Sure. I removed the changes from the server configuration, created a test app, and verified that removing the max file size limit for uploads works as expected. I also added a new documentation section explaining how users can enable large file uploads. This includes a link to the Program.cs setup tutorial and a small code block with a warning callout, highlighting that users must explicitly configure their own maximum file size afterward.

I can provide the source code for the test app if needed.

Test app:
https://github.com/user-attachments/assets/b5e6c22b-e81f-4f8a-937a-1871b7c3f125

Implemented Doc section:
Screenshot 2026-01-28 at 01 10 09

src/Ivy/Server.cs
```diff
@@ -17,6 +17,7 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http; //do not remove - used in RELEASE
+using Microsoft.AspNetCore.Http.Features;
```
rorychatt: not needed?

PR #2128, opened 2026-01-27
(feat): improve expandable disabled state

ArtemKhvorostianyi:
https://github.com/user-attachments/assets/5af191e9-c8ea-438c-92ca-19940ee136c6

Referenced to:
https://github.com/Ivy-Interactive/Ivy-Framework/issues/1820
image

https://github.com/user-attachments/assets/8aa0ce68-02f8-4485-9401-b407adf13eeb

### DIFF
--- a/src/frontend/src/components/ui/expandable/expandable-variants.ts
+++ b/src/frontend/src/components/ui/expandable/expandable-variants.ts
@@ -17,7 +17,7 @@ export const expandableTriggerVariants = cva(
 );
 
 export const expandableHeaderVariants = cva(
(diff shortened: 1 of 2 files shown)


dcrjodle:
Should disabled be hoverable?

dcrjodle:
ChatGPT says 
The general rule

If an element is disabled (can’t be interacted with), it shouldn’t look or behave like it’s interactive. That typically means:

❌ No hover styles that suggest “click me”

❌ No pointer cursor

❌ No focus / active states

Why? Because hover feedback is a promise. If nothing can happen, the UI shouldn’t tease the user.

dcrjodle:
Now I saw the other video example. I would still only have a hover effect on the toggle switch and not in the exapandable container. it's ok to merge

Issue #743, opened 2025-09-05
[Feature]: Submit auth form on return/enter


zachwolfe:
### Description

When logging in with a username/password (`PasswordEmailFlowView`) it would be nice to be able to hit enter/return to submit, as is the norm for most websites.

### Code Examples

```csharp

```

zachwolfe:
This still doesn't work for me

rorychatt:
I solved this recently

zachwolfe:
@rorychatt how recently? I tested on latest main before reopening

zachwolfe:
I'm going to reopen this again because it still isn't working for me. I will investigate shortly. Here is a video (latest main, rebuilt frontend):

https://github.com/user-attachments/assets/e0f1ef7c-1ba7-49a0-b1e6-0bfb6f13a2db

rorychatt:
hmm...

Works for me here

<img width="3839" height="2159" alt="Image" src="https://github.com/user-attachments/assets/be900658-1458-4824-994a-5fd4227ee595" />

https://samples.ivy.app/tests/form-submit-test-app

Maybe issue is within the specific implementation.
I can verify that the feature should work for forms otherwise.

I will investigate.

zachwolfe:
@rorychatt That looks like a bespoke form rather than `PasswordEmailFlowView`; maybe that is the difference?

zachwolfe:
@rorychatt That sample you linked works for me as well BTW

zachwolfe:
Reopening this again because this behaviour is still not working in `DefaultAuthApp`. After some investigation it looks like `FormBuilder`'s [validation/submission logic](https://github.com/Ivy-Interactive/Ivy-Framework/blob/f769269e799442ce4a6e328a066c5471c1ecbdcb/Ivy/Views/Forms/FormBuilder.cs#L323-L334) does run when I hit enter, but `DefaultAuthApp`'s logic for [actually logging in](https://github.com/Ivy-Interactive/Ivy-Framework/blob/f769269e799442ce4a6e328a066c5471c1ecbdcb/Ivy/Auth/DefaultAuthApp.cs#L91-L134) does not, because it is never passed to the `FormBuilder`.

zachwolfe:
Fixed in #2678, thanks @ArtemKhvorostianyi!

PR #2137, opened 2026-01-28
(feat): enable submitting auth form on enter

ArtemKhvorostianyi:
Reference: https://github.com/Ivy-Interactive/Ivy-Framework/issues/743
The authentication form in Ivy samples had several issues:

1. Form submission on Enter key was working, but the login logic wasn't executing
2. Password verification was failing due to Secret mismatch between hash generation and verification
3. Array index errors were occurring during form validation and auth flows
4. Page reload loops were happening on logout in optional auth mode

https://github.com/user-attachments/assets/c5a1845d-be2a-4247-a80c-a40298b294f0

- Fixed form submission callback connection
- Fixed password verification Secret alignment
- Added defensive checks for array access
- Prevented reload loops in optional auth mode
- Made DefaultAuthApp auto-discoverable

P.S. I edited some code that I’m not directly responsible for, so I’m not 100% sure whether I was allowed to change it. I’m open to feedback or criticism on that part. That said, we clearly have issues with authentication in the Ivy samples, and this is something that needs attention regardless. The PR can be closed if needed — my goal was mainly to highlight the problematic areas. That said, the implementation does work as it is now.

### DIFF
--- a/src/Ivy/AppHub.cs
+++ b/src/Ivy/AppHub.cs
@@ -79,9 +79,9 @@ public class AppHub(
                 requestScheme = forwardedProto.ToString();
             }
 
-            var authProvider = server.ServiceProvider!.GetService<IAuthProvider>();
-            if (authProvider != null)
+            if (server.AuthProviderType != null)
             {
+                var authProvider = server.ServiceProvider!.GetService<IAuthProvider>() ?? throw new Exception("IAuthProvider not found");
 #if DEBUG
                 authProvider = new CheckedAuthProvider(authProvider);
 #endif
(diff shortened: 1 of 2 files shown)


src/Ivy/Auth/BasicAuthProvider.cs
```diff
@@ -71,10 +71,11 @@ public BasicAuthProvider()
 
     private bool PasswordMatches(string username, string password, string hash)
     {
+        // Note: Hash is generated without Secret in SamplesServer, so verify without Secret to match
         return Argon2.Verify(hash, new Argon2Config
         {
             Password = Encoding.UTF8.GetBytes(password),
-            Secret = _hashSecret,
```
rorychatt, line 77: yes, but SamplesServer is not the only one using this.

So I think this could be a bad fix

src/Ivy.Samples.Shared/SamplesServer.cs
```diff
@@ -10,9 +13,35 @@ public static class SamplesServer
     public static async Task RunAsync(ServerArgs? args = null)
     {
         CultureInfo.DefaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US");
+
```
rorychatt, line 16: I don't think we need this.

We should instead focus on creating a new Ivy app, adding different auth providers (including basic) and testing out enter behavior there.

This issue is not about Ivy-Framework, although it might not work there due to some other issue like, maybe, bad config of SamplesServer.

src/Ivy/Views/Forms/FormBuilder.cs
```diff
@@ -309,7 +309,11 @@ public FormBuilder<TModel> Scale(Scale scale)
     private FormBuilderField<TModel> GetField<TU>(Expression<Func<TModel, TU>> field)
     {
         var name = Utils.GetNameFromMemberExpression(field.Body);
-        return _fields[name];
+        if (!_fields.TryGetValue(name, out var fieldValue))
+        {
```
rorychatt, line 313: this can be very handy check

src/Ivy/AppHub.cs
```diff
@@ -79,9 +79,9 @@ public override async Task OnConnectedAsync()
                 requestScheme = forwardedProto.ToString();
             }
 
-            if (server.AuthProviderType != null)
+            var authProvider = server.ServiceProvider!.GetService<IAuthProvider>();
```
rorychatt: any change you do to an important file like here needs to have a very strong justification.

Considering my comments above, I am not reviewing this file.

rorychatt (changes requested):
->  solve this by seeing what happens after you:

ivy init 
ivy auth add

-> in the new app. not in samples

PR #2138, opened 2026-01-28
(tables): refactors across frontend code

ArtemKhvorostianyi:
- We intentionally use wrap-break-word because break-words gets highlighted as a mistake.

### 2. TableApp.cs

> Code Duplication (lines 89-93, 102-106, 115-119): The column width configuration is repeated identically across all three table size examples. Consider extracting this to a helper method:

```
private static Table ConfigureColumnWidths(Table table) =>
    table.ColumnWidth(e => e.Sku, Size.Fraction(0.15f))
         .ColumnWidth(e => e.Foo, Size.Fraction(0.1f))
         .ColumnWidth(e => e.Name, Size.Fraction(0.3f))
         .ColumnWidth(e => e.Price, Size.Fraction(0.15f))
         .ColumnWidth(e => e.Url, Size.Fraction(0.3f));
```

> Magic Numbers: Fraction values (0.15f, 0.1f, 0.3f) are repeated without explanation. Consider using named constants or comments explaining the rationale.

- Since this is a samples app, I don’t think it needs refactoring. Also, the helper method doesn’t really make it look cleaner to me. I can refactor it if needed, though.

image

### 3. TableWidget.tsx

> Width Detection Logic (lines 23-28): The string-based width type detection is fragile. Consider using proper type discrimination

- Refactored. Tested. Works the same. Looks cleaner and formatted in consistence with other widgets. 

### 4. TableCellWidget.tsx:

> Type Safety (line 66): The typeof children === 'string' check is good, but consider using a more robust type guard

- Have a look at my conversation with @dcrjodle lower in comments. 

### - All changes were tested out.

### DIFF
--- a/src/frontend/src/widgets/tables/TableWidget.tsx
+++ b/src/frontend/src/widgets/tables/TableWidget.tsx
@@ -17,7 +17,6 @@ export const TableWidget: React.FC<TableWidgetProps> = ({
   scale = Scales.Medium,
 }) => {
   const widthStyles = getWidth(width || 'Full');
-  const isFull = widthStyles.width === '100%';
 
   return (
     <Table
@@ -25,8 +24,9 @@ export const TableWidget: React.FC<TableWidgetProps> = ({
       className={cn('w-full caption-bottom')}
       style={{
         ...widthStyles,
-        tableLayout: isFull ? 'fixed' : 'auto',
-        ...(isFull && { maxWidth: '100%' }),
+        ...(widthStyles.width === '100%'
+          ? { maxWidth: '100%', tableLayout: 'fixed' as const }
+          : { tableLayout: 'auto' as const }),
       }}
     >
       <TableBody>{children}</TableBody>


src/frontend/src/widgets/tables/TableCellWidget.tsx
```diff
@@ -83,20 +89,17 @@ export const TableCellWidget: React.FC<TableCellWidgetProps> = ({
   // Don't apply to data cells without widths - they need to size naturally
   const shouldTruncate = width || isHeader;
 
-  // Only show tooltip for string children to avoid "[object Object]" issues
-  const shouldShowTooltip = !multiLine && typeof children === 'string';
+  // Only show tooltip for string/number children to avoid "[object Object]" issues
+  const shouldShowTooltip = !multiLine && isStringContent(children);
```
dcrjodle:
```suggestion
  const shouldShowTooltip = !multiLine && ['string', 'number'].contains(children)
```

Why do we need a function for this? Let's write less code

ArtemKhvorostianyi: Thanks for feedback, @dcrjodle 
Thanks for the feedback, @dcrjodle — I mostly agree with you. If you check the linked issue, you’ll see that I’m intentionally trying to discard most of the refactors in the comments.

This particular point is a bit tricky, and I wasn’t fully confident about it. On one hand, we have the note:

> The typeof children === 'string' check is good, but consider using a more robust type guard

—which refers exactly to this part of the code. On the other hand, there’s the question of code clarity and readability.

That’s why the PR is still in draft: I wanted to get advice from @rorychatt before finalizing the approach.

For now, I’ll keep the inline check, but we should still improve it by handling numbers as well — that part was missed.

> The old version only checked typeof children === 'string'. Table cells often render numbers (42, 3.14, etc.). If you only allow strings, those cells never get a tooltip when truncated.

rorychatt:
I don't care if we break out something that is used just in 1 place or not, as long as it is easier to read the function call vs function implementation.

src/frontend/src/widgets/tables/TableWidget.tsx
```diff
@@ -13,39 +13,26 @@ interface TableWidgetProps {
 
 export const TableWidget: React.FC<TableWidgetProps> = ({
   children,
-  width,
+  width = 'Full',
```
rorychatt: does this kind of contradict with check on line 21 "isFullWidth"?

I am afraid we might have broken some of the behaviors for tables in different containers.

Why was this default missing before?
Do tables look as they should in blades, sheets, cards, with set or not set custom widths, etc?

Inside of container that can grow or not?

rorychatt (changes requested):
.

ArtemKhvorostianyi:
Refactored and restructured the code to align with the logic of other widgets — Size.Full() is now set by default.

Tables in cards:
image

Tables in sheets:
https://github.com/user-attachments/assets/70027734-7d06-478d-8b13-7eb5cf64f410

Tables in blades:
https://github.com/user-attachments/assets/065c59df-aea3-448d-8c2e-f00ff135da2c

Issue #2059, opened 2026-01-16
(labels): overflow over other elements in grid


rorychatt:
### Description

<img width="1040" height="886" alt="Image" src="https://github.com/user-attachments/assets/61f96ee9-07f5-4166-8ec9-f9757294c9b4" />

I expected to see ellipsis overflow, where you could hover and get full reading (if it got ellipsis)

PR #2149, opened 2026-01-29
[GridLayout]: fix cell overflow behavior on hover

defymecobra:
**Description:**
This PR fixes a critical UX issue where interactive widgets (like `ColorInput`) inside a `Layout.Grid` would unexpectedly "pop out" (switching to absolute positioning with high z-index) when hovered, causing layout shifting and preventing proper interaction.

**Context:**
This behavior was introduced in #2095 to handle text overflow in grid cells. However, enforcing this style (`gridCellOverflow.ellipsis`) globally on `GridLayoutWidget` is too aggressive for a generic layout container. It indiscriminately treats all cell content—including complex forms and interactive components—as truncatable text.

**Changes:**
- Reverts the forced `gridCellOverflow.ellipsis` usage in `GridLayoutWidget`.
- Restores stable layout behavior for components inside Grids.

### DIFF
--- a/src/frontend/src/lib/styles.ts
+++ b/src/frontend/src/lib/styles.ts
@@ -382,11 +382,6 @@ export const getOverflow = (overflow?: Overflow): React.CSSProperties => {
   };
 };
 
-export const gridCellOverflow = {
-  ellipsis:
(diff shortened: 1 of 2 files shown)


defymecobra:
As long as old implementation caused a bug, we need to reopen this issue https://github.com/Ivy-Interactive/Ivy-Framework/issues/2059 and find other better solution

PR #2169, opened 2026-02-02
(theme): add border radius support to themes

dcrjodle:


src/frontend/src/index.css
```diff
@@ -132,17 +131,13 @@
     --text-small-label: 12px;
 
     /* Spacing */
-    --radius-sm: calc(var(--radius) - 4px);
-    --radius-md: calc(var(--radius) - 2px);
-    --radius-lg: var(--radius);
-    --radius-xl: calc(var(--radius) + 4px);
+    --radius-sm: var(--radius-sm);
```
rorychatt: as we discussed, we should have configurable radiuses not for sizes, but for component types

rorychatt (approved):

rorychatt:
(theme customizer): add configurator for border radiuses

We now support theme defined border radiuses. We need to update existing FE components and theme editor to support behavior like it is done in daisy ui

https://github.com/user-attachments/assets/cc5c0c0f-dc82-4104-b19a-497671eef284

PR #2172, opened 2026-02-03
[Tabs]: Fix dropdown overflow and improve width calculations

ArtemLazarchuk:
## Problem
- Dropdown button was overflowing container boundaries when width decreased
- Add button width wasn't accounted for in tab visibility calculations
- Incorrect width reference used in TabsVariant (container vs tabsList)

https://github.com/user-attachments/assets/a7641d43-cb70-4a67-8844-72e85b3fe409

## Solution
- Moved dropdown button inside flex container (after add button) to prevent overflow
- Removed unnecessary wrapper div around add button for cleaner structure
- Fixed width calculation to use correct reference element (tabsList for TabsVariant, container for ContentVariant)
- Added automatic measurement of non-tab elements (add button, dropdown) to accurately calculate available space

## Changes
- **DropdownMenu.tsx**: Changed dropdown positioning from absolute to inline flex item
- **Variants.tsx**: Simplified add button structure and moved dropdown inside flex container
- **useTabCalculation.ts**: Improved width calculations with proper reference element and automatic non-tab element measurement

### DIFF
--- a/src/frontend/src/widgets/layouts/tabs/components/DropdownMenu.tsx
+++ b/src/frontend/src/widgets/layouts/tabs/components/DropdownMenu.tsx
@@ -32,7 +32,7 @@ interface TabsDropdownMenuProps {
 /**
  * Dropdown menu component for displaying hidden tabs when they overflow.
  * Supports drag-and-drop reordering within the dropdown.
- * 
+ *
  * Returns an object with trigger button and menu content to allow flexible placement.
  */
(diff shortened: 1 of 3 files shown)


ArtemLazarchuk:
@claude review this

rorychatt:
I don't really like that we ended up calculating width in ts. But I guess it is a hell of an issue to do this in css.

defymecobra (approved):

PR #2183, opened 2026-02-05
[Forms]: fix nullable inputs in scaffolded forms

defymecobra:
## Problem Description

In scaffolded forms, the "X" button for clearing nullable fields was not appearing. For example, a `string? NullableString` field was displayed without the ability to clear the value to `null`.

## Root Cause

The problem consisted of two parts:

### 1. `TextInput<TString>` uses `new` keyword

```csharp
// TextInputBase
[Prop] public bool Nullable { get; set; }

// TextInput<TString>
[Prop] public new bool Nullable { get; set; } = typeof(TString).IsNullableType();
```

The `new` keyword creates a **new** property that hides the base one. When code in `ToTextInput()` set `input.Nullable = true` through a variable of type `TextInputBase`, it was setting the **base** class property. However, the serializer was reading the **derived** class property, which remained `false`.

### 2. `Utils.IsNullableType()` does not support nullable reference types

```csharp
public static bool IsNullableType(this Type type)
{
    return Nullable.GetUnderlyingType(type) != null; // Only works for int?, DateTime?
}
```

This method correctly identifies `int?`, `DateTime?`, but **not** `string?`, since `Nullable.GetUnderlyingType(typeof(string?))` returns `null`.

### 1. Reflection for setting Nullable on derived class

The `ToTextInput()` method now uses reflection with `BindingFlags.DeclaredOnly`:

```csharp
var nullableProperty = genericType.GetProperty("Nullable",
    BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
nullableProperty?.SetValue(input, type.IsNullableType());
```

### 2. Support for nullable reference types via NullabilityInfoContext

Added `IsNullable` property to `FieldPropertyInfo` that uses `NullabilityInfoContext` (.NET 6+ API):

```csharp
public bool IsNullable
{
    get
    {
        // Nullable value types (int?, DateTime?)
        if (Nullable.GetUnderlyingType(Type) != null)
            return true;

        // Nullable reference types (string?, MyClass?)
        if (Type.IsClass)
        {
            var nullabilityContext = new NullabilityInfoContext();
            var nullabilityInfo = nullabilityContext.Create(PropertyInfo);
            return nullabilityInfo.ReadState == NullabilityState.Nullable;
        }
        return false;
    }
}
```

### 3. Requirements from @nielsbosma

As per the comment in the issue:

> Make Nullable a part of IAnyInput (move from bool etc.)
> Use X in all input types.
> If Required => then never show the X

All requirements fulfilled:

| Requirement | Status |
|-------------|--------|
| Nullable as part of IAnyInput | ✅ Already implemented in `IAnyInput.Nullable` |
| X in all input types | ✅ All widgets support `nullable` prop |
| Required => don't show X | ✅ `if (field.IsNullable && !field.Required)` |

## Changed Files

- `src/Ivy/Widgets/Inputs/TextInput.cs` — reflection in `ToTextInput()`
- `src/Ivy/Views/Forms/FormScaffolder.cs` — `IsNullable` property + `!field.Required` check

image

### DIFF
--- a/src/Ivy/Widgets/Inputs/TextInput.cs
+++ b/src/Ivy/Widgets/Inputs/TextInput.cs
@@ -138,10 +138,8 @@ public static class TextInputExtensions
         var type = state.GetStateType();
         Type genericType = typeof(TextInput<>).MakeGenericType(type);
         TextInputBase input = (TextInputBase)Activator.CreateInstance(genericType, state, placeholder, disabled, variant)!;
-        
         var nullableProperty = genericType.GetProperty("Nullable", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
         nullableProperty?.SetValue(input, type.IsNullableType());
-        
         return input;
     }
 


defymecobra:
@claude review this

rorychatt (approved):

nielsbosma (approved):

PR #2188, opened 2026-02-06
(search): show document path in search results

ArtemKhvorostianyi:
Reference:
image

Result:
image

How it was:
image


rorychatt:
navigation with arrow keys is now broken

PR #2206, opened 2026-02-08
[DataTable]: implement automatic data refresh

defymecobra:
## Problem

DataTable currently loads data only once during initialization and does not update automatically when the underlying data changes. Users expect real-time updates without manual page refresh.

**Current behavior:**
- User adds/removes/modifies data
- DataTable shows stale data
- User must manually refresh the page to see changes

**Expected behavior:**
- Data changes are immediately reflected in the DataTable

## Solution

Implemented a SignalR-based notification system that allows the backend to signal the frontend when data changes, triggering an automatic refresh.

### Why this approach?

1. **SignalR already exists** — The framework uses SignalR for real-time communication, so no new dependencies needed
2. **Minimal API surface** — Single method `NotifyChange()` for developers to call
3. **Targeted updates** — Uses `sourceId` to refresh only the specific DataTable that changed, not all tables on the page
4. **No polling** — Push-based updates are more efficient than polling

### Architecture

```
Backend                              Frontend
────────                             ────────
connection.NotifyChange()
    ↓
IClientNotifier.NotifyClientAsync()
    ↓
SignalR "DataTableRefresh" ──────→  use-backend.tsx
                                        ↓
                                    CustomEvent dispatch
                                        ↓
                                    tableProvider.tsx
                                        ↓
                                    refreshTrigger++
                                        ↓
                                    useDataLoading.ts re-fetches data
```

### Backend

| File | Change |
|------|--------|
| `DataTableConnection.cs` | Added `NotifyChange()` method |
| `UseDataTable.cs` | Integrated `IClientNotifier` for SignalR messaging |

### Frontend

| File | Change |
|------|--------|
| `use-backend.tsx` | Added `DataTableRefresh` SignalR handler |
| `tableProvider.tsx` | Subscribe to `CustomEvent`, manage `refreshTrigger` state |
| `useDataLoading.ts` | Added `refreshTrigger` to effect dependencies |

## Usage

```csharp
var connection = UseDataTable(data.AsQueryable());
var table = new DataTable(connection, ...);

async void OnAddItem(Event<Button> _)
{
    data.Add(new Item(...));
    await connection.NotifyChange(); // ← Table refreshes automatically
}
```

https://github.com/user-attachments/assets/0ca347f5-23ba-42a9-8901-74654ecd411c

### DIFF
--- a/src/frontend/src/widgets/dataTables/types/types.ts
+++ b/src/frontend/src/widgets/dataTables/types/types.ts
@@ -45,10 +45,10 @@ export interface DataColumn {
 }
 
 export interface DataTableConnection {
-  connectionId: string;
-  sourceId: string;
   port: number;
   path: string;
+  connectionId: string;
+  sourceId: string;
 }
 
 export interface DataTableConfig {


defymecobra:
@claude review this

rorychatt (changes requested):
@defymecobra what about a pattern using a refresh token?

I am just worried that with this solution, we introduce yet another refresh mechanism, and this might be super confusing.

Can't this be just part of the internal hook `UseDataTable`?

defymecobra:
# Auto-refresh DataTable on view rebuild

`UseDataTable` hook now automatically sends a refresh signal to the frontend on every rebuild after the initial setup. This eliminates the need for users to manually call `NotifyChange()`.

**How it works:**
- User modifies data and calls any `UseState.Set()` → view rebuilds
- `UseDataTable` detects it's not the first run and calls `NotifyChange()` internally
- Frontend receives the SignalR signal and re-fetches fresh data

**Changes:**
- `UseDataTable.cs` — auto-refresh on subsequent rebuilds
- `DataTableConnection.cs` — `NotifyChange()` changed from `public` to `internal`
- `DataTableRefreshTestApp.cs` — updated to use the new pattern

Visual work lasts the same

rorychatt:
@defymecobra what if you have scrolled a bit more. lets say you have 1000 rows, and scroll to row number 500.

You add a row to the end. Does the datatable get refreshed?

defymecobra:
> @defymecobra what if you have scrolled a bit more. lets say you have 1000 rows, and scroll to row number 500.
> 
> You add a row to the end. Does the datatable get refreshed?

No

defymecobra:
- Added `Version` field to `DataTableConnection` record
- Updated `UseDataTable` to increment version on rebuilds to force reference updates
- Updated frontend `useDataLoading` to depend on `connection.version` for auto-refresh
- Updated types and tests

src/Ivy/Views/DataTables/UseDataTable.cs
```diff
@@ -27,7 +29,12 @@ public static class UseDataTableExtensions
             // Store cleanup for later
             context.UseEffect(() => cleanup, []);
         }
+        else if (hasRun.Value && connection.Value != null)
+        {
+            version.Set(v => v + 1);
```
rorychatt: pretty sure this won't work too.

How would a similar problem be solved in pure react?

defymecobra:
- Update [UseDataTable](cci:1://file:///c:/Users/User/Desktop/Ivy-Framework/src/Ivy/Views/DataTables/UseDataTable.cs:13:4-37:5) to re-register queryable when instance reference changes, triggering a `SourceId` update
- Remove explicit `Version` field from [DataTableConnection](cci:2://file:///c:/Users/User/Desktop/Ivy-Framework/src/Ivy/Widgets/DataTables/DataTableConnection.cs:3:0-3:95) and related logic, relying on natural object identity
- Implement Stale-While-Revalidate pattern in [useDataLoading](cci:1://file:///c:/Users/User/Desktop/Ivy-Framework/src/frontend/src/widgets/dataTables/dataTableContext/hooks/useDataLoading.ts:27:0-216:2) to keep data visible during refreshes
- Persist column widths during data updates in `TableProvider` to prevent layout shifts
- Clean up unused hooks and update test mocks

rorychatt:
@claude review this

rorychatt:
@claude review this

PR #2207, opened 2026-02-08
(text area): redesign drag button

ArtemKhvorostianyi:
Screenshot 2026-02-08 at 19 25 00

How it was:
Screenshot 2026-02-08 at 19 26 01

### DIFF
--- a/src/frontend/src/index.css
+++ b/src/frontend/src/index.css
@@ -165,39 +165,6 @@
   [data-sidebar='sidebar'] svg[fill='currentColor'] {
     color: var(--primary);
   }
-
-  /* Textarea resizer: custom grip same as native (two diagonal lines), only color is theme var(--border). */
-  textarea.ivy-textarea::-webkit-resizer {
-    appearance: none;
-    background: transparent;
-  }
-  .ivy-textarea-wrapper {
-    position: relative;
-    overflow: visible;
-  }
-  .ivy-textarea-wrapper::before,
-  .ivy-textarea-wrapper::after {
-    content: '';
-    position: absolute;
-    bottom: 4px;
-    right: 2px;
-    width: 2px;
-    height: 1px;
-    background: var(--muted-foreground);
-    transform: rotate(-45deg);
-    transform-origin: bottom right;
-    pointer-events: none;
-    z-index: 10;
-  }
-  .ivy-textarea-wrapper::after {
-    width: 8px;
-    bottom: 8px;
-    right: 2px;
-  }
-  .ivy-textarea-wrapper:has(.resize-none)::before,
-  .ivy-textarea-wrapper:has(.resize-none)::after {
-    display: none;
-  }
 }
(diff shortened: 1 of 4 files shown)


rorychatt:
The shorter and longer line as it was, is better.

We just need to make it response to theme. Can you give it a fill color same like for borders maybe? Should solve the issue as borders are theme responsive.

ArtemKhvorostianyi:
@rorychatt 
We don’t render this element ourselves. It is an auto-generated ::-webkit-resizer that appears on <textarea> elements. Since it is a WebKit-specific resizer, the color of the three corner lines cannot be changed. We can modify the background color or set a background image, but not the line color itself.

This behavior is admittedly confusing, but it’s a known limitation—I’ve reviewed multiple articles discussing this restriction.

We already have a few issues related to WebKit behavior across the framework, and this is another example of that. As a potential solution, I would suggest rendering a custom SVG or Lucide icon with our own styling instead of relying on the native WebKit resizer.

rorychatt:
Screenshot 2026-02-10 at 10 43 46
it is possible and github has it for example

Screenshot 2026-02-10 at 10 44 30

rorychatt:
see how GitHub solved it and copy with pride 
@ArtemKhvorostianyi

ArtemKhvorostianyi:
> is possible and github has it for example

I did some research on this and found that GitHub also uses the standard ::-webkit-resizer, not a custom implementation, and it works well in most cases.

I’ve provided a solution with a custom style for the WebKit resizer. It now respects theme colors, and the issue with theme awareness is resolved. Please let me know before merging if the code needs any structural refactoring — for example, whether it would be better to move this logic into a separate CSS file for the textarea or handle it directly in textarea.tsx. I’m open to adjusting the implementation.

Example from release:
image

New style:
image

src/frontend/src/index.css
```diff
@@ -165,6 +165,39 @@
   [data-sidebar='sidebar'] svg[fill='currentColor'] {
     color: var(--primary);
   }
+
+  /* Textarea resizer: custom grip same as native (two diagonal lines), only color is theme var(--border). */
+  textarea.ivy-textarea::-webkit-resizer {
```
rorychatt: yes, this should be moved to components/ui/textarea/textarea.css

it also needs an index ts and proper named exports and module declaration (index.ts file)

rorychatt (changes requested):
.

src/frontend/src/index.css
ArtemKhvorostianyi: > yes, this should be moved to components/ui/textarea/textarea.css
> 
> it also needs an index ts and proper named exports and module declaration (index.ts file)

image

Sure, I moved it

rorychatt:
Screenshot 2026-02-12 at 17 01 06
next time if you want me to review something, ask for a review at top right in the page

PR #2228, opened 2026-02-11
(docs): use small scale for tables in api section

ArtemKhvorostianyi:
As an example of this change, here is the updated API section of the Grid Layout documentation page:
How it was:
image

How it's now:
image


rorychatt:
@ArtemKhvorostianyi do you think this looks better?

ArtemKhvorostianyi:
> @ArtemKhvorostianyi do you think this looks better?

It seems a bit senseless to make this change — it’s too minor, and the PR could probably be closed (it took around 10 minutes to implement).

In some cases, long strings appear on a single line when using a small scale. On the other hand, the gaps look too small, and tables become slightly harder to read.

Here’s a comparison:

Screenshot 2026-02-11 at 12 35 54

PR #2229, opened 2026-02-11
(docs): implement external widgets doc page

ArtemKhvorostianyi:
I built a simple external widget using this documentation, so it has been tested in practice (except for the troubleshooting section, which was copied from the source md file in the issue).

Screenshot 2026-02-11 at 05 26 12
Screenshot 2026-02-11 at 05 26 26
Screenshot 2026-02-11 at 05 26 38
Screenshot 2026-02-11 at 05 26 50
Screenshot 2026-02-11 at 05 27 05
Screenshot 2026-02-11 at 05 27 17
Screenshot 2026-02-11 at 05 27 27

### DIFF
--- a/src/Ivy.Docs.Shared/Docs/02_Widgets/07_Advanced/05_ExternalWidgets.md
+++ b/src/Ivy.Docs.Shared/Docs/02_Widgets/07_Advanced/05_ExternalWidgets.md
@@ -14,8 +14,6 @@ searchHints:
 External [widgets](../../01_Onboarding/02_Concepts/03_Widgets.md) let you extend the Ivy Framework with custom React components built and bundled separately from the core framework. Use them for domain-specific UI (e.g. diagrams, charts, or rich editors) without coupling that code to the framework backend.
 </Ingress>
 
-The pattern has three parts: a **C# proxy** (a widget record with `[ExternalWidget]`), a **React component** (your UI), and a **build pipeline** that compiles the frontend and embeds the assets into the widget assembly.
-
 ## Architecture Overview
 
 1. **C# proxy** — A record inheriting from `WidgetBase<T>` with `[ExternalWidget]`, defining [props](../../01_Onboarding/02_Concepts/03_Widgets.md) and [events](../../01_Onboarding/02_Concepts/05_EventHandlers.md).


rorychatt:
@ArtemKhvorostianyi troubleshooting needs validation too

ArtemKhvorostianyi:
> need to also test everything in troubleshooting if it makes sense or not

Okay, I tested this out and there is my feedback after testing:
### Script resource not found
Is this a real issue?
Yes. If the framework looks up the script by a resource name derived from the path in [ExternalWidget], and that path does not match what is actually embedded (or the DLL is stale), the lookup fails and you get a "Script resource not found" style error.
So: the issue is real, the resolution is right

### Global not found
Yes. The host loads the widget script and then looks up a global (e.g. window[globalName]) to get the component. If that global is missing or has the wrong name, you get "Global not found".
Yes. Each part was checked against this project and the built bundle.

### Duplicate type / CS0579
Observed errors:
CS0436 (first run): type DuplicateWidget in the host’s source conflicts with the type imported from the referenced DuplicateWidget assembly.
CS0579 (second run): duplicate attributes (e.g. TargetFrameworkAttribute, AssemblyCompanyAttribute, …) because the host was also compiling the widget project’s generated files under Widgets/DuplicateWidget/obj/.
So both “duplicate type” and “duplicate attribute” (CS0579) can occur when the widget lives under the host and the host does not exclude the widget directory.

After adding "what to check" to the host’s .csproj:
Build succeeded (no CS0436, no CS0579).
dotnet run worked and the host used the type from the referenced widget DLL.
So the troubleshooting advice is correct.

### Invalid hook call / multiple React
Yes. React’s docs and behavior are clear: hooks must run in the same React instance that is rendering the tree. If the widget bundle includes its own copy of React and the host app also has React, there are two instances. The widget’s hooks then run in the widget’s React, while the host renders with its React, so React reports “Invalid hook call” (or similar) because the hook and the render tree use different copies.
The resolution is correct: keep react and react-dom in external and globals in Vite, and use jsxRuntime: 'classic'.

### Wrong filename (.iife.js)
Yes. With a lib build and formats: ['iife'], Vite’s default output name is not ExternalWidget.js.
The resolution is correct: set fileName: () => 'ExternalWidget.js' in vite.config.ts so the output name has no extra suffix and matches the path in [ExternalWidget]

### Excluding the widget folder (integrated pattern)
The guidance is correct. Both excluding via DefaultItemExcludes and via explicit Compile/None Remove work; the path pattern and ProjectReference in your doc are right. 



And also I verified Multiple widgets in one bundle, built second widget to verify this section. Here is my feedback:
image

So the doc is tested out

PR #2235, opened 2026-02-12
[DataTable]: bind HandleCellAction to OnCellClick and fix link cell event propagation

defymecobra:
## Issue
Users reported that `.HandleCellAction()` was not triggering when clicking on cells rendered with `LinkDisplayRenderer`.
Upon analysis, the root causes were:
1.  **Backend Binding**: `HandleCellAction` was bound to `OnCellActivated` (Activated), which typically corresponds to double-clicks for text cells. Link interactions are naturally single-click.
2.  **Frontend Event Propagation**: The frontend logic (`useCellInteractions.ts`) handled link navigation (e.g., `window.open`) but stopped event propagation for link cells, preventing the backend from receiving the signal.
3.  **Data Format**: Link cells were sending a raw JSON object (e.g., `{"kind":"link-cell", ...}`) to the backend, making it difficult to use the value in C# actions without parsing.

## Solution
This PR unifies cell interactions to use **Single Click** semantics for actions and ensures proper event propagation for links.

### Backend Changes (`DataTableBuilder.cs`)
-   **Changed Binding**: Rebound `HandleCellAction` from `_onCellActivated` (Double Click) to `_onCellClick` (Single Click).
    -   *Why*: Actions in a web context (buttons, links, row selection) typically expect a single click. Requiring double-click for text actions but single-click for links created an inconsistent UX. Now all configured cell actions trigger on a single click.

### Frontend Changes (`useCellInteractions.ts`)
-   **Event Propagation**: Modified `handleCellClicked` to dispatch the `OnCellClick` event **before** executing link navigation logic.
    -   *Why*: Ensures the backend reliably receives the event even if the browser initiates navigation (especially critically for same-tab links where the page might unload).
-   **Value Unwrapping**: Added logic to extract the clean `url` string from `link-cell` data objects.
    -   *Why*: Improves Backend DX. The action handler now receives a simple string (e.g., `"https://example.com"`) instead of a complex JSON object.

## Verification
-   Verified with a reproduction app (`DataTableCellActionApp`).
-   **Link Cells**: Clicking a link now correctly opens the URL **AND** triggers the `HandleCellAction` (sending the clean URL string).
-   **Text Cells**: Clicking a text cell now triggers the `HandleCellAction` on a single click (consistent behavior).

image

### DIFF
--- a/src/frontend/src/widgets/dataTables/hooks/useCellInteractions.ts
+++ b/src/frontend/src/widgets/dataTables/hooks/useCellInteractions.ts
@@ -47,14 +47,16 @@ export const useCellInteractions = ({
           ) {
             return content.data;
           } else if ('data' in content) {
-            const cellData = (content as unknown as { data: any }).data;
+            const cellData = (content as unknown as { data: unknown }).data;
 
             if (
               cellData &&
               typeof cellData === 'object' &&
-              cellData.kind === 'link-cell'
+              'kind' in cellData &&
+              (cellData as { kind: string }).kind === 'link-cell' &&
+              'url' in cellData
             ) {
-              return cellData.url;
+              return (cellData as unknown as { url: string }).url;
             } else {
               return cellData;
             }


defymecobra:
@claude review this

PR #2237, opened 2026-02-13
(bool input): fix incorrect vertical alignment and missing top spacing

ArtemKhvorostianyi:
There was an issue where boolean input checkboxes and their labels had inconsistent vertical alignment.
image

The problem was mainly caused by raw heights not being properly handled across different scales, which resulted in the checkbox being rendered incorrectly. 
Added boolInputRowMinHeightVariants
Added shrink-0 to the control wrapper to keep its size and alignment stable.
Fixed:

image
image

### DIFF
--- a/src/frontend/src/widgets/inputs/BoolInputWidget.tsx
+++ b/src/frontend/src/widgets/inputs/BoolInputWidget.tsx
@@ -184,12 +184,7 @@ const VariantComponents = {
           )}
           onClick={e => e.stopPropagation()}
         >
-          <div
-            className={cn(
-              description && 'mt-1.5',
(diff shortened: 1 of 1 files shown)


ArtemLazarchuk:
@rorychatt That change actually comes from another PR. I updated the styling there to make the border-radius configurable, so it can adapt to different style variants.
Previously it was using `border-sm`, and now it’s using `border-selector` instead

ArtemKhvorostianyi:
@rorychatt please review feedback from @ArtemLazarchuk

rorychatt:
@ArtemKhvorostianyi can you review this? By default from Ivy Design system (external repo) borders for this should look as they did before. Create an issue and fix this if needed

PR #2238, opened 2026-02-13
(validation): improve validation strategy

ArtemKhvorostianyi:
Since we have three input widgets where users can type from the keyboard, we need proper validation. These widgets are TextInput, Field, and Form.

The TextInput widget includes variants such as password, email, URL, and telephone.
The Field widget also supports password, email, URL, and telephone variants.
The Form widget provides additional functionality: it can detect field types based on labels, and users can also supply custom validation.

The main logic is centered around text input: since every input — whether it is a Field or a Form field — is still based on the TextInput widget, the validation logic is built around the TextInput widget itself.

The main goal of this PR is not to change any API usage, but to implement validation logic for password, email, URL, and telephone inputs.

So I built a validation module with the main validation logic:
/Ivy/Validation/Validators.cs
image

Then I wired it up with the TextInput backend part so we can call these methods from textInput.cs.

image

When a TextInput is wrapped with WithField, or used inside Forms, we validate the text input first and then apply the invalid state if necessary.

There are no API changes, and I tried to minimize the amount of added code.

That is the main goal of the PR, but I also had to do some additional work on Forms validation, because it supports custom validation logic, which was often broken. I updated its code as well.

Note 1: TextInputBuildContext is necessary to get the current state of the text input so we can validate input in real time. This caused a small change in the widget tree to wire it properly.

Note 2: Some files shown in Files changed were not actually modified in the final result — GitHub highlighted whitespace changes because those files were edited during development but later reverted.

Note 3: All tests passed; I just removed them from the PR description.

It works, and it works smooth now

Text validation:

image

image

Field validation:

image

image

Form validation:
image

image

### DIFF
--- a/src/Ivy/Widgets/Inputs/TextInput.cs
+++ b/src/Ivy/Widgets/Inputs/TextInput.cs
@@ -1,12 +1,22 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using Ivy.Core;
 using Ivy.Core.Helpers;
 using Ivy.Core.Hooks;
+using Ivy.Validation;
 
 // ReSharper disable once CheckNamespace
 namespace Ivy;
+
+/// <summary>Internal: set by the framework during view.Build() so state.ToEmailInput() etc. can validate on blur without this. or duplicate methods.</summary>
+internal static class TextInputBuildContext
+{
+    private static readonly AsyncLocal<IViewContext?> Current = new();
+    public static void SetCurrent(IViewContext? context) => Current.Value = context;
+    internal static IViewContext? GetCurrent() => Current.Value;
+}
 
 public record Affix
 {
@@ -151,13 +161,15 @@ public static class TextInputExtensions
 
     public static TextInputBase ToSearchInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputVariants.Search);
 
-    public static TextInputBase ToPasswordInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputVariants.Password);
-
-    public static TextInputBase ToEmailInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputVariants.Email);
-
-    public static TextInputBase ToUrlInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputVariants.Url);
-
-    public static TextInputBase ToTelInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputVariants.Tel);
+    /// <summary>Email/Password/Url/Tel input. Validates on blur when built inside a view; in forms FormView also runs validation.</summary>
+    public static TextInputBase ToEmailInput(this IAnyState state, string? placeholder = null, bool disabled = false) =>
+        TextInputBuildContext.GetCurrent() is { } ctx ? BuildValidatedInput(ctx, state, TextInputVariants.Email, placeholder, disabled) : state.ToTextInput(placeholder, disabled, TextInputVariants.Email);
+    public static TextInputBase ToPasswordInput(this IAnyState state, string? placeholder = null, bool disabled = false) =>
+        TextInputBuildContext.GetCurrent() is { } ctxP ? BuildValidatedInput(ctxP, state, TextInputVariants.Password, placeholder, disabled) : state.ToTextInput(placeholder, disabled, TextInputVariants.Password);
+    public static TextInputBase ToUrlInput(this IAnyState state, string? placeholder = null, bool disabled = false) =>
+        TextInputBuildContext.GetCurrent() is { } ctxU ? BuildValidatedInput(ctxU, state, TextInputVariants.Url, placeholder, disabled) : state.ToTextInput(placeholder, disabled, TextInputVariants.Url);
+    public static TextInputBase ToTelInput(this IAnyState state, string? placeholder = null, bool disabled = false) =>
+        TextInputBuildContext.GetCurrent() is { } ctxT ? BuildValidatedInput(ctxT, state, TextInputVariants.Tel, placeholder, disabled) : state.ToTextInput(placeholder, disabled, TextInputVariants.Tel);
 
     public static TextInputBase Placeholder(this TextInputBase widget, string placeholder) => widget with { Placeholder = placeholder };
 
@@ -217,6 +229,21 @@ public static class TextInputExtensions
         return widget with { OnBlur = new(_ => { onBlur(); return ValueTask.CompletedTask; }) };
     }
 
+    private static TextInputBase BuildValidatedInput(IViewContext context, IAnyState state, TextInputVariants variant, string? placeholder, bool disabled)
+    {
+        var invalidState = context.UseState(default(string?), true);
+        var blurOnceState = context.UseState(false, true);
+        context.UseEffect(() =>
+        {
+            if (blurOnceState.Value)
+            {
+                var (_, err) = Validators.ValidateForVariant(state.As<object>().Value, variant);
+                invalidState.Set(err ?? "");
+            }
+        }, state, blurOnceState);
+        void OnBlur(Event<IAnyInput> _) => blurOnceState.Set(true);
+        return state.ToTextInput(placeholder, disabled, variant).Invalid(invalidState.Value ?? "").OnBlur(OnBlur);
+    }
     public static TextInputBase Value<T>(this TextInputBase widget, T value)
     {
         if (widget is TextInput<T> typedWidget)
--- a/src/Ivy/Validation/Validators.cs
+++ b/src/Ivy/Validation/Validators.cs
@@ -2,9 +2,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Mail;
 using System.Text.RegularExpressions;
-using Ivy;
-using Ivy.Widgets.Inputs;
-
 namespace Ivy.Validation;
 
 /// <summary>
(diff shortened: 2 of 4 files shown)


ArtemKhvorostianyi:
I removed the frontend part of the solution and implemented an approach aligned with how validation is handled in forms. Validation is now managed entirely on the backend, and everything works correctly — I’ve retested it to confirm.

Please take a look at the changes in TextInput.cs and Field.cs. I kept the implementation minimal, but the added classes and interfaces are necessary to properly connect the widgets with each other.

rorychatt:
@claude review this

src/Ivy/Widgets/Inputs/Field.cs
```diff
@@ -1,11 +1,18 @@
...
+using Ivy.Widgets.Inputs.Validated;
 
 // ReSharper disable once CheckNamespace
 namespace Ivy;
 
-public record Field : WidgetBase<Field>
+/// <summary>
+/// Return type of WithField(). Implemented by Field and ValidatedFieldView so Label/Description/Required/Help (and Width when Field) apply without conflicting with Card or other types.
+/// </summary>
+public interface IFieldWithLabel { }
```
rorychatt: Field should have a label inside of it. We don't need this refactor IMO (Field -> IFieldWIthLabel is a breaking change)
And also file is still named Field.cs
This needs to be fixed

src/Ivy/Widgets/Inputs/Field.cs
rorychatt, line 13: label is here

src/Ivy/Widgets/Inputs/Field.cs
```diff
@@ -58,6 +65,56 @@ public static class FieldExtensions
...
+
+    public static IWidget WithTooltip(this IFieldWithLabel fieldOrView, string toolTip) =>
+        new Tooltip(fieldOrView, toolTip);
+
+    public static IFieldWithLabel WithField(this IAnyInput input)
+    {
+        if (input is TextInputBase tb && tb.BoundState != null &&
+            tb.Variant is TextInputs.Email or TextInputs.Password or TextInputs.Tel or TextInputs.Url)
+            return new ValidatedFieldView(tb, tb.BoundState, tb.Variant);
+        return new Field(input);
```
rorychatt: also pls update main PR description

src/Ivy/Widgets/Inputs/TextInput.cs
```diff
@@ -40,6 +41,12 @@ public interface IAnyTextInput : IAnyInput
 
 public abstract record TextInputBase : WidgetBase<TextInputBase>, IAnyTextInput
 {
+    /// <summary>When set, enables auto-validation when wrapped with WithField().</summary>
```
rorychatt: text input has no logic for invalid state, disabled, etc.

They support this if it is passed over by a parent. 

A parent can be a form, or a field.

Field then can dictate disabled/invalid/etc state of what it is wrapped around (an input)

src/Ivy/Widgets/Inputs/TextInput.cs
```diff
@@ -155,6 +163,13 @@ public static TextInputBase ToTextInput(this IAnyState state, string? placeholde
 
     public static TextInputBase ToTelInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputs.Tel);
 
+    /// <summary>
+    /// Validates value for the given text input variant. Returns (true, null) if valid, (false, errorMessage) if invalid.
+    /// Used by FormFieldView via Validators.
+    /// </summary>
```
rorychatt: sounds super suspicious that we need this

src/Ivy.Samples.Shared/Apps/Tests/ValidationTestApp.cs
```diff
@@ -0,0 +1,98 @@
...
+        var email = UseState("");
+        var tel = UseState("");
+        var url = UseState("");
+        var password = UseState("");
+
+        return new Card(
+            Layout.Vertical().Gap(4)
+                | Text.H3("TextInput variants (no Field)")
+                | Text.P("Email, Tel, Url, and Password variants.")
+                | email.ToEmailInput("e.g. user@example.com")
```
rorychatt: maybe solution is to just in this case return not EmailInput instance, but Field<TextInput> or whatever we already support

rorychatt: then we don't need this new custom logic for when input has field's logic

src/Ivy.Samples.Shared/Apps/Tests/ValidationTestApp.cs
```diff
@@ -0,0 +1,98 @@
...
+                | tel.ToTelInput("e.g. +1 234 567 8900")
+                | url.ToUrlInput("e.g. https://example.com")
```
rorychatt: same for all these.

src/Ivy.Samples.Shared/Apps/Tests/ValidationTestApp.cs
```diff
@@ -0,0 +1,98 @@
...
+        var email = UseState("");
+        var tel = UseState("");
+        var url = UseState("");
+        var password = UseState("");
+
+        return new Card(
+            Layout.Vertical().Gap(6)
+                | Text.H3("Field-wrapped inputs")
+                | Text.P("Each input is wrapped in a Field with label, description, and required.")
+                | email.ToEmailInput("user@example.com").WithField().Label("Email").Description("We use this for account recovery.").Required()
```
rorychatt: maybe correct solution is to rename this:

.ToEmailInput -> .ToEmailField

src/Ivy.Samples.Shared/Apps/Tests/ValidationTestApp.cs
```diff
@@ -0,0 +1,98 @@
...
+                | tel.ToTelInput("+1 234 567 8900").WithField().Label("Phone").Description("7–15 digits.").Required()
+                | url.ToUrlInput("https://example.com").WithField().Label("Website").Description("Must start with http or https.")
```
rorychatt: ToUrlField

src/Ivy.Samples.Shared/Apps/Tests/ValidationTestApp.cs
```diff
@@ -0,0 +1,98 @@
...
+                | password.ToPasswordInput("At least 8 characters").WithField().Label("Password").Description("Required for sign-up.").Required()
+        ).Width(Size.Full());
+    }
+}
+
+public record ValidationFormModel(string Email, string Password, string? PhoneNumber, string? Website);
```
rorychatt, line 68: this already exists in Fields

rorychatt (changes requested):
in general let's not create new instance.

Stuff like .ToTelInut, .ToEmailInput is old code and just needs to be updated to support new fields.

likely they just need a refactor like

.ToEmailInput -> .ToEmailField

then we can also just use existing field functionality, where field controls invalid/disabled state of whatever input it wraps around.

And then we just predefine validation requirements to these new .To...Field as it was done before.

rorychatt:
@ArtemKhvorostianyi don't work on this until further confirmation by me. I need to sync with my team a bit

rorychatt:
This is now on hold for a week

rorychatt:
^ try and implement my suggested changes above. If they make sense to you

ArtemKhvorostianyi:
> Stuff like .ToTelInut, .ToEmailInput is old code and just needs to be updated to support new fields.
> likely they just need a refactor like
> 
> .ToEmailInput -> .ToEmailField

I did these refactors and changed usage of inputs across samples and docs. Please let me know what do you think about this solution. If it's ok, I'll update field and text input docs and I'll remove validation test app. Changes are tested out, it works. Also tests are succesfull

rorychatt:
@ArtemKhvorostianyi I will sync with my team. Let's have this on hold for now.

rorychatt (changes requested):
on hold for some time

src/Ivy/Widgets/Inputs/TextInput.cs
```diff
@@ -149,13 +150,17 @@ public static TextInputBase ToTextInput(this IAnyState state, string? placeholde
...
+    public static ValidatedFieldView ToEmailField(this IAnyState state, string? placeholder = null, bool disabled = false) => new(state, TextInputs.Email, placeholder, disabled);
 
-    public static TextInputBase ToEmailInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputs.Email);
+    /// <summary>Returns a validated Field for password. Use .Label(), .Description(), .Required() etc. on the result.</summary>
+    public static ValidatedFieldView ToPasswordField(this IAnyState state, string? placeholder = null, bool disabled = false) => new(state, TextInputs.Password, placeholder, disabled);
 
-    public static TextInputBase ToUrlInput(this IAnyState state, string? placeholder = null, bool disabled = false) => state.ToTextInput(placeholder, disabled, TextInputs.Url);
+    /// <summary>Returns a validated Field for URL. Use .Label(), .Description(), .Required() etc. on the result.</summary>
+    public static ValidatedFieldView ToUrlField(this IAnyState state, string? placeholder = null, bool disabled = false) => new(state, TextInputs.Url, placeholder, disabled);
```
rorychatt: let's keep the old naming where we had ToUrlInput (same for others)

src/Ivy/Views/Forms/FormHelpers.cs
```diff
@@ -38,9 +38,10 @@ public static bool IsRequired(PropertyInfo propertyInfo)
                         DisplayName = propertyInfo.Name
                     };
                     var result = capturedAttr.GetValidationResult(value, validationContext);
-                    return result == ValidationResult.Success
+                    var isValid = result == null || result == ValidationResult.Success;
+                    return isValid
                         ? (true, "")
-                        : (false, result?.ErrorMessage ?? "Validation failed");
+                        : (false, result!.ErrorMessage ?? "Validation failed");
```
rorychatt: can this ! break runtime? make sure we properly handle nullable types

ArtemKhvorostianyi: Well, that solution didn't break anything. But yes, I refactored it a bit. 
The code was refactored to an if/return so the compiler can see that result is non-null in the failure path, and the ! was removed .
Screenshot 2026-03-07 at 04 50 36

PR #2249, opened 2026-02-15
[Iframe]: block javascript execution by default

defymecobra:
### Issue
Previously, the `Iframe` widget allowed arbitrary JavaScript execution within embedded content. This posed a potential security risk (XSS), as there were no restrictions on what scripts could run inside the frame.

### Root Cause
The underlying `<iframe>` HTML element was rendered without a `sandbox` attribute. In modern browsers, iframes typically have full script execution privileges unless explicitly restricted by the `sandbox` attribute or Content Security Policy (CSP).

### Solution
I implemented a "secure by default" strategy:

1.  **Frontend (`IframeWidget.tsx`)**:
    *   Added the `sandbox` attribute to the rendered `<iframe>`.
    *   Configured the default sandbox policy to allow forms, modals, popups, and same-origin requests, but **block scripts** (`allow-scripts` is omitted by default).

2.  **Backend (`Iframe.cs`)**:
    *   Added a new property `AllowJavaScript` (boolean, default: `false`).
    *   This allows developers to explicitly opt-in to script execution if their use case requires it.

### Reasoning
This changes the behavior from "permissive by default" to "restrictive by default". It ensures that embedding external content does not automatically introduce script execution vulnerabilities. The addition of the `AllowJavaScript` flag satisfies the requirement to allow scripts only when explicitly requested by the developer.

### DIFF
--- a/src/Ivy/Widgets/Primitives/Iframe.cs
+++ b/src/Ivy/Widgets/Primitives/Iframe.cs
@@ -24,4 +24,6 @@ public record Iframe : WidgetBase<Iframe>
     [Prop] public string Src { get; set; } = null!;
 
     [Prop] public long? RefreshToken { get; }
+
+    [Prop] public bool AllowJavaScript { get; set; }
 }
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

rorychatt:
We will have this on hold.

rorychatt:
YOLO

PR #2250, opened 2026-02-15
[Docs]: serve /agents.md and /llms.txt routes

defymecobra:
## Problem

The documentation site (`docs.ivy.app`) did not serve the raw contents of `AGENTS.md` and `llms.txt` at dedicated URLs. AI agents and LLMs that look for these standard files (similar to `robots.txt` for search engines) could not automatically discover context about the Ivy framework from the docs site.

## Solution

Both `/agents.md` and `/llms.txt` now serve the raw content of the root `AGENTS.md` file. A separate `llms.txt` file is not needed because its content is identical to `AGENTS.md` — both are introductory guides for LLMs about the Ivy framework.

### Why reuse `AGENTS.md` for both routes?

The repo already had a comprehensive `llms.txt` (commit `6e904af6`) which was essentially the same content as `AGENTS.md`. Maintaining two separate files with identical content would create unnecessary duplication. Instead, both routes serve the same embedded resource.

### Why `SitemapMiddleware`?

The serving logic was added to the existing `SitemapMiddleware` rather than creating a new middleware. This middleware already handles similar "meta" files for bots and crawlers (`/robots.txt`, `/sitemap.xml`), so `/agents.md` and `/llms.txt` fit naturally alongside them.

### Backend

- **`SitemapMiddleware.cs`** — Added handling for `/agents.md` and `/llms.txt` routes, serving the embedded `AGENTS.md` resource as `text/plain`
- **`Ivy.Docs.Shared.csproj`** — Added `AGENTS.md` from the repo root as an embedded resource

### Frontend

- **`vite.config.ts`** — Updated the dev server proxy regex from `^/.*\.md$` to `^/(.*\.md|llms\.txt)$` so that `/llms.txt` is also proxied to the backend during development

image
image

### DIFF
--- a/src/Ivy.Docs.Shared/Ivy.Docs.Shared.csproj
+++ b/src/Ivy.Docs.Shared/Ivy.Docs.Shared.csproj
@@ -22,6 +22,7 @@
   <ItemGroup>
     <EmbeddedResource Include="Assets/**/*" />
     <EmbeddedResource Include="Generated/**/*.md" />
+    <EmbeddedResource Include="..\..\AGENTS.md" Link="Assets\AGENTS.md" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


defymecobra:
@claude review this

rorychatt (approved):

PR #2254, opened 2026-02-15
[ThemeCustomizer]: use local ThemeService to prevent global theme leak

defymecobra:
## Issue
Changing the color theme in **ThemeCustomizer** was applying changes globally to all users of the application. If one user modified the theme in demo mode, these settings became the default for the entire server.

**Root Cause:**
`ThemeService` is registered in the DI container as a **Singleton** (`Services.AddSingleton<IThemeService, ThemeService>()`).
In `ThemeCustomizer.cs`, this service was injected via `UseService<IThemeService>()`.
Calling `themeService.SetTheme(...)` mutated the state of the single `ThemeService` instance on the server. As a result, `GenerateThemeCss()`, called during page rendering for any user, returned the modified styles.

## Solution
In `ThemeCustomizer.cs`, the usage of the global `IThemeService` has been replaced with a local instance of `ThemeService`.

Before:
```csharp
var themeService = UseService<IThemeService>();
```

After:
```csharp
// Use a local ThemeService to generate CSS without mutating the global singleton
var themeService = new ThemeService();
```

## Why this solution was chosen
`ThemeCustomizer` is a tool for previewing and customizing the theme within the current session (client-side). It does not require modifying the global server state.
Creating a local `ThemeService` instance (lightweight object) allows us to:
1.  Reuse the existing CSS generation logic (`GenerateThemeCss`).
2.  Ensure changes remain isolated within the local variable scope.
3.  Send the generated CSS only to the **current** client via SignalR (`client.ApplyTheme(css)`), without affecting other users.

This is the safest and simplest way to fix the global state leak without changing the service registration architecture.

Tested this solution with different browsers with their own sessions and all works correctly, theme doesn't change

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Demos/ThemeCustomizer.cs
+++ b/src/Ivy.Samples.Shared/Apps/Demos/ThemeCustomizer.cs
@@ -16,18 +16,19 @@ public class ThemeCustomizer : SampleBase
         var currentTheme = UseState(Theme.Default);
         var isExportOpen = UseState(false);
         var client = UseService<IClientProvider>();
+        var themeService = new ThemeService();
         var selectedMode = UseState("light"); // "light" or "dark"
 
         // Individual color states for live editing
         var editingTheme = UseState(CloneTheme(Theme.Default));
 
+        // UseQuery handles theme application reactively with built-in state management
         var themeQuery = UseQuery(
             key: editingTheme.Value,
             fetcher: async ct =>
             {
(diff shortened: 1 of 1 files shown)


defymecobra:
@claude review this

rorychatt:
@claude review this

rorychatt:
@defymecobra can you look up docs for the hook we used before?
we need to make sure we properly describe how that works. So next time it is not a surprise that things like this happen

PR #2259, opened 2026-02-16
(auth): send refresh tokens only to refresh requests

### LINKED ISSUE #998: (auth): Don't send refresh token to every Ivy request
### Description

The user's refresh token is rarely needed. As such, it should be stored in a separate cookie from the access token, and set/retrieved only when strictly necessary.

### Use Case

_No response_

### Proposed Solution

- Split `jwt` cookie into `jwt` and `jwt_refresh_token` (and stop using `jwt_ext_refresh_token`)
- Scope `jwt_refresh_token` to an auth-related path that is as specific as possible (e.g., `/auth` or `/auth/refresh`)
- Expose an API endpoint from the server that can use `IAuthService` to refresh the access token and save a newly-rotated refresh token (this part might be tricky)
- Call into the refresh endpoint from a frontend hook, which in turn will be triggered from the backend.

### Alternatives Considered

_No response_

### Impact

_No response_

### Code Examples

```csharp

```

### Additional Context

_No response_

zachwolfe:
Fixes #998


zachwolfe:
I will revisit this after #2296 as there is some overlap

PR #2262, opened 2026-02-17
(bool): improve visibility to respect theme customiser and ivy desighn system

ArtemKhvorostianyi:
Now:
image

Before (broken):
image
The issue was that the bool input in the checkbox variant used the same border-radius scaling as larger elements (such as buttons and badges), which caused it to appear almost circular.

### DIFF
--- a/src/Ivy/Themes/ThemeConfig.cs
+++ b/src/Ivy/Themes/ThemeConfig.cs
@@ -21,8 +21,6 @@ public class Theme
 
     public string? BorderRadiusSelectors { get; set; }
 
-    public string? BorderRadiusCheckbox { get; set; }
-
     public static Theme Default => new()
     {
         Name = "Default",
--- a/src/frontend/src/components/ui/checkbox.tsx
+++ b/src/frontend/src/components/ui/checkbox.tsx
@@ -69,7 +69,7 @@ const Checkbox = React.forwardRef<
       }
     };
 
-    const baseClass = `peer ${getSizeClasses(scale)} shrink-0 rounded-selector border border-border shadow focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring cursor-pointer disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-primary data-[state=checked]:text-primary-foreground data-[state=checked]:border-primary dark:border-white/10`;
+    const baseClass = `peer ${getSizeClasses(scale)} shrink-0 rounded-checkbox border border-border shadow focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring cursor-pointer disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-primary data-[state=checked]:text-primary-foreground data-[state=checked]:border-primary dark:border-white/10`;
     const finalClass = className?.includes('bg-red-50')
       ? baseClass.replace('data-[state=checked]:bg-primary', '')
       : baseClass;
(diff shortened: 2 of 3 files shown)


src/Ivy/Themes/ThemeConfig.cs
```diff
@@ -21,6 +21,8 @@ public class Theme
 
     public string? BorderRadiusSelectors { get; set; }
 
+    public string? BorderRadiusCheckbox { get; set; }
```
rorychatt: Image

we dont need to add a 4th border radius config.

rorychatt:
https://daisyui.com/theme-generator/#theme=eJxtlOtu4yAQhV8FWaq0KzUIhoth34bYuLHqmAgctbtV330xThMI_heY75y54Xw1sznb5k8zjW-npXltOjc5fwjdyRbXh8MWOJpgD5SQGHLvU3f6FX-_IILI72cIMkirfYbljNhnOjcvdl7uHNDIYUIkAiWwUiJXXPx4Nv7vHearKQaOoG0xAbbDVgk0SwnYKgLcKpWLgo14n6eQtxQUMcExI_t4nYWnLKAQ44BBFDLTdTnbtitLBSCqKNYUarbyZyr5SxY1CnMqc81sr4s302OJWzFEbDMt53SD6w5gE8U5KYlZUdQ4D-5RfnKncWEMsJT0GaxXrLfao4Kz2HdbjPQaGw7h4S6Te9siKhmGsvYbXE8nTZQkkcKa55oP4-dxfruzKvVJlUaKYw56h638eXqklALiAmtS-FvvnX-Uv5GaI8qiu6rIejrba4jLooDJT-3e9OM1xAc32W5J_gQLb895cBjt1KcIPIeO7rOUhPGfLd2gjO2ZHZ3v7UrTy2e66O1lOa3ndJrdGNY_FRJPvR3MdYpdDWYK9rW5eDtYH-K38n67-_4PjS9Oyg

we expect it to work like here

ArtemKhvorostianyi:
> don't add a 4th border radius property.

I see your point, but in Ivy the checkbox currently uses the same border radius as larger elements like buttons and badges, which typically have radii of 8px or more. Because of that, checkboxes end up looking almost circular.
I think we should separate checkbox styling from other components and assign it a dedicated border radius. I can implement this by reusing the existing props but scaling them down specifically for checkboxes. Would that approach be acceptable?

rorychatt:
@ArtemKhvorostianyi we likely just need to adjust the values then.

See daisy ui default values: look more square

image



Our standard looks like this: 

image

ArtemKhvorostianyi:
> @ArtemKhvorostianyi we likely just need to adjust the values then.
> 
> See daisy ui default values: look more square

I removed the fourth property for setting the border radius for checkboxes. I implemented a simple solution by applying the same border radius as badges and buttons, but scaling it down for checkboxes so they keep a square appearance.
Now:
image
From release version:
image

The issue is that Theme Customizer and DaisyUI use hardcoded radius values (in rem), so those “square” checkboxes you showed in screenshots aren’t actually square — they are just large so border radius 0.5 rem(8px) isn't enough to make it circle and it stays square. If we apply scale.large to our bool input, it will appear square as well. This shows a limitation in the current implementation and indicates it should be reworked to be responsive. We can see the same issue in the DaisyUI app’s component variants view, where different element sizes are displayed.

Screenshot 2026-02-17 at 23 36 34

We also can’t simply reduce the border radius for buttons, badges, and checkboxes simultaneously: buttons look correct with current values, badges become slightly too sharp, and checkboxes need smaller radii to preserve their shape. So there isn’t a universal radius value that works well for all three components at once.

My conclusion is that ThemeCustomizer should use responsive metrics so border radii scale correctly across different element sizes, because we will problem with elements visibility in different scales in the future.

rorychatt (approved):

PR #2265, opened 2026-02-17
Revert "Revert "(icons): loading icon always spin""

rorychatt:
Reverts Ivy-Interactive/Ivy-Framework#2264


rorychatt:
@dcrjodle we need some other solution, the previous (this one) is too "magic"

PR #2289, opened 2026-02-19
(Command Palette): add command palette widget

joshuauaua:
-This PR proposes a Command Palette Widget to provide an alternative to the current search functionality in the sidebar. 
-Keyboard shortcut: Shift + 7 to open. Arrow keys to nav and Esc to close.
-Worked with both Ivy.Docs and Ivy.Samples
-I have injected it into DefaultSidebarChrome
-Mapped in widgetMap, utilizes Command primitives (cmdk)

-For testing purposes I tried to inject it into root (App.tsx) to test without DefaultSidebarChrome.

Screenshot 2026-02-19 at 12 01 03

https://github.com/user-attachments/assets/eee1ecd9-d9ce-4158-b7b6-d60b6024acd4

### DIFF
--- a/src/Ivy/Chrome/DefaultSidebarChrome.cs
+++ b/src/Ivy/Chrome/DefaultSidebarChrome.cs
@@ -50,9 +50,9 @@ public class DefaultSidebarChrome(ChromeSettings settings) : ViewBase
         {
             return navigate.Receive(navigateArgs =>
             {
-                  OpenApp(navigateArgs);
-                  return default!;
-              });
+                OpenApp(navigateArgs);
+                return default!;
+            });
         });
 
         UseEffect(() =>
--- a/src/frontend/src/widgets/internal/CommandPaletteWidget.tsx
+++ b/src/frontend/src/widgets/internal/CommandPaletteWidget.tsx
@@ -47,12 +47,12 @@ export const CommandPaletteWidget = ({
     }
   }, [search]);
 
-  // Reset search when dialog closes
-  useEffect(() => {
-    if (!isOpen) {
+  const onOpenChange = (open: boolean) => {
+    setIsOpen(open);
+    if (!open) {
       setSearch('');
     }
-  }, [isOpen]);
+  };
 
   const flatItems = useMemo(() => {
     return items.filter(item => item.tag && item.label);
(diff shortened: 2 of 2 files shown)


src/Ivy/Chrome/DefaultSidebarChrome.cs
```diff
@@ -466,6 +466,14 @@ void OnTabReorder(Event<TabsLayout, int[]> @event)
             ,
             Layout.Vertical(
                 new SidebarNews("https://ivy.app/news.json"),
+                new CommandPalette(
+                    appRepository.GetMenuItems().FlattenWithPath().Select(x => x.Item with { Path = x.Path }).ToArray(),
```
rorychatt, line 470: should live outside of sidebar chrome

src/Ivy/Chrome/DefaultSidebarChrome.cs
```diff
@@ -466,6 +466,14 @@ void OnTabReorder(Event<TabsLayout, int[]> @event)
             ,
             Layout.Vertical(
                 new SidebarNews("https://ivy.app/news.json"),
+                new CommandPalette(
```
rorychatt, line 469: this should likely be injectable into main entry point just like defaustsidebar chrome is

src/frontend/src/widgets/internal/CommandPaletteWidget.tsx
rorychatt, line 1: this will not be internal if we do implement this into the framework.

rorychatt: not a 100% sure though

rorychatt:
don't ask anyone's review, but mine

rorychatt:
@joshuauaua we are coming up with a bit of different feature that would play the same role, so let's have this on hold

joshuauaua:
> @joshuauaua we are coming up with a bit of different feature that would play the same role, so let's have this on hold

Sounds good

dcrjodle:
I know this is on hold, but ftr I think the command should shift+P

rorychatt:
merging this into
https://github.com/Ivy-Interactive/Ivy-Framework/pull/2407

PR #2310, opened 2026-02-22
[ErrorSheet]: scope scrolling to error content area

defymecobra:
## Summary

When an error was displayed in the sidebar sheet, the **entire sheet container** gained a vertical scrollbar (`overflow-y-auto` on `SheetContent`). This made it easy to miss parts of the error content and looked visually inconsistent with the rest of the UI — the sheet header and close button would scroll off-screen along with the error details.

## What was done

Replaced the scroll strategy in `ErrorSheet.tsx` so that the sheet itself **never scrolls vertically** — only the error content area inside it does.

### Changes in `ErrorSheet.tsx`

| Property | Before | After |
|---|---|---|
| `SheetContent` className | `overflow-y-auto` | `flex flex-col h-full overflow-hidden` |
| Content wrapper className | `mt-6` | `mt-6 flex-1 min-h-0 overflow-y-auto` |

### Why this approach

- **`overflow-hidden`** on `SheetContent` prevents the entire sheet from scrolling
- **`flex flex-col h-full`** turns the sheet into a full-height flex container so children can distribute space
- **`flex-1 min-h-0`** on the content wrapper allows it to shrink below its content size (critical for flex overflow behavior) and fill remaining space
- **`overflow-y-auto`** on the content wrapper enables scrolling **only within** the error content area

This fix is scoped entirely to `ErrorSheet.tsx` and does **not** modify `ErrorDisplay.tsx`, so the 3 other consumers (`ErrorBoundary`, `ErrorWidget`, `DataTableWidget`) are unaffected.

## Result

- ✅ Opening an error **no longer introduces** a scrollbar on the sheet container
- ✅ The sheet header ("Oops! Something went wrong") and close button **remain fixed** at the top
- ✅ Error content (message + stack trace) scrolls **within its own area**
- ✅ "Copy Details" button remains accessible after scrolling
- ✅ UI behaves correctly on different viewport widths

## Before:
image
image

## After:
image
image

### DIFF
--- a/src/frontend/src/components/ErrorSheet.tsx
+++ b/src/frontend/src/components/ErrorSheet.tsx
@@ -27,12 +27,12 @@ export function ErrorSheet() {
         >
           <SheetContent
             side="right"
-            className="w-full sm:max-w-lg overflow-y-auto"
+            className="w-full sm:max-w-lg flex flex-col h-full overflow-hidden"
           >
             <SheetHeader>
               <SheetTitle>Oops! Something went wrong</SheetTitle>
             </SheetHeader>
-            <div className="mt-6">
+            <div className="flex-1 min-h-0 overflow-y-auto">
               <ErrorDisplay
                 title={title}
                 message={message}


defymecobra:
@claude review this

rorychatt (approved):

rorychatt:
something is still wrong

image


I was expecting to see that only this error "code block" would be scrollbable (it's contents)

Like it is done in other palces. But it seems the scrollbar is a bit too high up in the dom

rorychatt (changes requested):
.

defymecobra:
The scrollbar was positioned on the entire sheet content area,
making it "too high up in the DOM". Moved it down to the stack
trace code block level by replacing ScrollArea with a native
div (max-h-[50vh] overflow-auto) in ErrorDisplay.

- Reverted SheetContent to original overflow-y-auto
- Constrained stack trace block height to 50vh with its own scrollbar
- Removed unused ScrollArea/ScrollBar imports from ErrorDisplay

https://github.com/user-attachments/assets/b2cc8983-b8a5-4c5e-865d-6cc2611e28db

rorychatt:
@defymecobra now you can see that the X on the top (header) scrolls away and is not sticky.

defymecobra:
Made the sheet header ("Oops! Something went wrong") and close button (X) sticky by converting SheetContent to a flex column layout with overflow-hidden. Error content scrolls independently in a flex-1 wrapper below the fixed header. 

Replaced ScrollArea with a native div (max-h-[50vh] overflow-auto) for the stack trace code block, so the scrollbar appears directly on the code block rather than on the sheet container.

Removed mt-6 from content wrapper since SheetContent's base gap-4
already provides spacing between header and content in flex mode.

https://github.com/user-attachments/assets/9264826a-a7e7-479d-bede-915a7a5ffd9e

defymecobra:
@claude review this

defymecobra:
@rorychatt check pls

PR #2315, opened 2026-02-23
[DataTable]: show scrollbars only when needed

ArtemLazarchuk:
## Summary
Fixes scrollbar behavior and layout issues in the DataTable widget.

### Scrollbar behavior
- **Vertical scrollbar**: Shown only when content overflows; hidden when empty rows fill the container (`overflow-y: hidden` when `hasEmptyRows`).
- **Horizontal scrollbar**: Removed unnecessary horizontal scrollbar caused by not accounting for vertical scrollbar width.
- Uses `scrollContainerHeight` from `.dvn-scroller` for correct scroll area measurement.

### Empty rows
- Adds empty filler rows when data does not fill the container and there is no pagination.
- Uses `Math.floor` to avoid overflow and unwanted scrollbar.
- `data-has-empty-rows` attribute drives scrollbar visibility and layout.

### Last column layout
- Last column uses `grow: 1` instead of manual width calculation.
- Eliminates the gap between the last column and the vertical scrollbar.
- Grid handles scrollbar space internally.

### Other
- Replaced invalid `ring`/`ring-color` with `box-shadow` for checkbox focus state.
- Added `flex: 1` and `minHeight: 0` on the grid container for correct layout in tabs.

## Before (the scrollbar was always visible, even when it wasn’t needed)

https://github.com/user-attachments/assets/e33a7d23-7fbf-4e95-94c7-49d081d12128

## After ( the scrollbar appears only when it is actually needed)

https://github.com/user-attachments/assets/1cb01643-c576-40d2-943a-c90296a11f53

### DIFF
--- a/src/frontend/src/widgets/dataTables/components/GridContainer.tsx
+++ b/src/frontend/src/widgets/dataTables/components/GridContainer.tsx
@@ -119,7 +119,7 @@ export const GridContainer: React.FC<GridContainerProps> = ({
     <div
       ref={containerRef}
       style={{ ...containerStyle, position: 'relative' }}
-      data-has-empty-rows={hasEmptyRows ? true : undefined}
+      data-has-empty-rows={hasEmptyRows || undefined}
     >
       <DataEditor
         ref={gridRef}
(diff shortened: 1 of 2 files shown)


ArtemLazarchuk:
@claude review this

PR #2317, opened 2026-02-23
(sidebar): make sidebar take full width on mobile

dcrjodle:
Screenshot 2026-02-23 at 16 54 14
Screenshot 2026-02-23 at 16 54 38


rorychatt:
@ArtemKhvorostianyi please test this

for default sidebar chrome only + wide app

default sidebar chrome + sidebar in a nested app

any other test cases you can think of

rorychatt:
@dcrjodle I like the idea but I have to put this on hold as we do a lot of weird stuff with sidebars (like smart search) and this might just not work all together. 

But having looked at diff, my guess is that this is relatively easy to one-shot this feature in the future. If we don't merge this PR this week I'll close this PR and we can re-open it.

PR #2407, opened 2026-02-26
Implement smart search widget

ArtemKhvorostianyi:
Smart search widget is a top pinned search input that is based on Ivy MCP server. User can ask question and get answers from mcp server in a view.

P.S. The code is ready for review. We have some related issues and a few new questions. 

https://github.com/user-attachments/assets/df759f59-0bbc-47af-b45b-74c3d818d797

@dcrjodle I handled the most of your feedback:

- Input is now focused when opening the dialoague. Users shouldn't have to click the input twice
- removed ESC KBD from search input
- Aligned the dialog at the top of the screen so the input never changes position
- Removed a title for the button, added the button text that explains what the button does, "Get an answer from Ivy Agent"
- Placed the button at the right corner of the dialogue (like a chat input)
- "Ideally the Chat UI should live inside the search dialogue, but I understand if that's complicated to implement" - I tried to implement it, but it works really tricky and for better user experience we should or improve dialog, sheet and chat components, or it looks very messy on the screen, so I decided to split this views separatly
- After closing result sheet, user should click on search again, the last search input stores there (have a look at the video). I think it's better then auto opening of search view after closing result view because user won't always search something else. 

Also I refactored placement of files for smart search since it's not a widget 
- /src/Ivy.Docs.Shared/Internal/SmartSearch.cs 
- /src/frontend/src/docs-internal/SmartSearch.tsx

In summary,  iliminated visual jank, improved UX

### DIFF
--- a/src/frontend/src/widgets/widgetMap.ts
+++ b/src/frontend/src/widgets/widgetMap.ts
@@ -82,7 +82,7 @@ import {
 import { DataTable } from '@/widgets/dataTables';
 import { TableWidget, TableRowWidget, TableCellWidget } from '@/widgets/tables';
 import React from 'react';
-import { SmartSearchWidget } from '@/widgets/internal/SmartSearchWidget';
+import { SmartSearch } from '@/docs-internal/SmartSearch';
 
 export const widgetMap = {
   $loading: LoadingScreen,
@@ -230,7 +230,7 @@ export const widgetMap = {
   ),
 
   // Internal
-  'Ivy.Docs.Shared.Internal.SmartSearch': SmartSearchWidget,
+  'Ivy.Docs.Shared.Internal.SmartSearch': SmartSearch,
   'Ivy.Widgets.Internal.SidebarNews': React.lazy(
     () => import('@/widgets/internal/SidebarNewsWidget')
   ),
--- a/src/frontend/src/docs-internal/SmartSearch.tsx
+++ b/src/frontend/src/docs-internal/SmartSearch.tsx
@@ -1,3 +1,7 @@
+/**
+ * Renderer for Ivy.Docs.Shared.Internal.SmartSearch (C#). Docs-only overlay + sidebar
+ * hijack; not a reusable widget, so it lives under docs-internal instead of widgets/.
+ */
 import { Dialog, DialogContent } from '@/components/ui/dialog';
 import { getHeight, getWidth } from '@/lib/styles';
 import { cn } from '@/lib/utils';
@@ -24,7 +28,7 @@ interface SmartSearchSlots {
   FollowUpChat?: React.ReactNode[];
 }
 
-interface SmartSearchWidgetProps {
+export interface SmartSearchProps {
   id: string;
   slots?: SmartSearchSlots;
   width?: string;
@@ -41,7 +45,8 @@ function openSmartSearchOverlay(): void {
     ?.click();
 }
 
-export const SmartSearchWidget: React.FC<SmartSearchWidgetProps> = ({
+/** Registered in widgetMap as Ivy.Docs.Shared.Internal.SmartSearch */
+export const SmartSearch: React.FC<SmartSearchProps> = ({
   id,
   slots: slotsProp,
   width = 'Full',
@@ -249,7 +254,6 @@ export const SmartSearchWidget: React.FC<SmartSearchWidgetProps> = ({
               '!top-4 sm:!top-8 !translate-y-0'
             )}
           >
-            {/* Wrapper so SearchVariant can query list buttons for ↑↓ navigation (same idea as sidebar search). */}
             <div data-smart-search-list>{overlayPanel}</div>
           </DialogContent>
         </Dialog>


ArtemKhvorostianyi:
@rorychatt it works pretty well, but I think I need some feedback and suggestions about backend/frontend implementation. I think it should be polished and refactored a bit more, but I need feedback to know what problems I need to solve

rorychatt:
@dcrjodle do you have any suggestions on the UI/UX?

src/frontend/src/widgets/widgetMap.ts
```diff
@@ -124,6 +122,8 @@ export const widgetMap = {
 
   // Widgets
   'Ivy.Article': ArticleWidget,
+  'Ivy.SmartSearch': SmartSearchWidget,
```
rorychatt: looks like bad imports here.

src/Ivy.Docs.Shared/Internal/SmartSearch.cs
```diff
@@ -0,0 +1,167 @@
...
+public record SmartSearch(params object?[] children) : WidgetBase<SmartSearch>(children.Where(c => c != null).Cast<object>().ToArray())
+{
+    internal SmartSearch() : this([]) { }
+}
+
+/// <summary>
+/// View for smart search: ask a question and get an AI-generated answer inline.
+/// Results shown below the search bar (no sheet).
+/// </summary>
+public class SmartSearchView : ViewBase
```
rorychatt, line 18: this is internal. Remove all xml docs.

rorychatt (changes requested):
update description of PR too

src/frontend/src/widgets/widgetMap.ts
rorychatt: should be beneath internal stuff

rorychatt:
@ArtemKhvorostianyi can you resolve the visual bugs when the response loads? 

You can just open the sheet right away with a fixed width. Don't make it resizeable - problem solved

src/frontend/src/widgets/widgetMap.ts
```diff
@@ -229,7 +223,20 @@ export const widgetMap = {
   ),
 
   // Internal
-
+  'Ivy.SmartSearch': React.lazy(() =>
+    import('@/widgets/internal/SmartSearchWidget/SmartSearchWidget').then(
+      m => ({
+        default: m.SmartSearchWidget,
+      })
```
rorychatt: why we have both Ivy.SmartSearch and Ivy.Docs.Shared.Helpers.SmartSearch?

It's the same thing.

For now use the second namespace. Later on we will flatten this

src/Ivy.Docs.Shared/Internal/SmartSearch.cs
```diff
@@ -0,0 +1,153 @@
+using Ivy.Docs.Shared.Services;
+
+namespace Ivy.Docs.Shared.Helpers;
+
+public record SmartSearch(params object?[] children) : WidgetBase<SmartSearch>(children.Where(c => c != null).Cast<object>().ToArray())
+{
+    internal SmartSearch() : this([]) { }
+}
+
+public class SmartSearchView : ViewBase
```
rorychatt, line 18: is helpers the right place for a view like this?

Don't we have a folder in ivy docs for ivy-docs-internal views?

src/frontend/src/widgets/internal/SmartSearchWidget/SmartSearchWidget.tsx
```diff
@@ -0,0 +1,507 @@
...
+                    }}
+                    activeTag={sidebarSearchState.activeTag}
+                  />
+                )}
+              </div>
+              {showMcpOption && (
+                <>
+                  <div className="shrink-0 border-t border-border" />
+                  <p className="shrink-0 py-3 px-2 text-sm text-muted-foreground">
+                    Search with Ivy MCP
```
rorychatt: ASK AI

-> let's not expose here that we have a MCP.

Yes, we use URL-s from there, but this is not to be shared

src/frontend/src/widgets/layouts/sidebar/sidebarSearchFilter.ts
```diff
@@ -0,0 +1,145 @@
+import type { MenuItem } from '@/types/widgets';
+
+/** Normalize string for search: strip whitespace/dash/underscore, lowercase */
```
rorychatt: so you load the full result into the frontend then apply sorting-scoring etc?

Haven't we done this for original main sidebar in the backend?

I mean, yes this is better for the server to sort on the client, but you need to note stuff like this when you don't handle state in the BE.

src/frontend/src/widgets/layouts/sidebar/sidebarSearchStore.ts
rorychatt, line 1: purpose?

src/frontend/src/widgets/article/ArticleWidget.tsx
```diff
@@ -34,9 +36,25 @@ export const ArticleWidget: React.FC<ArticleWidgetProps> = ({
...
   return (
-    <div className="flex flex-col gap-2 max-w-6xl mx-auto px-4 sm:px-6 lg:px-8 relative mt-8">
+    <div
+      className="flex flex-col gap-2 max-w-6xl mx-auto px-4 sm:px-6 lg:px-8 relative"
+      data-docs-content-boundary
+      style={
+        isOpen && panelWidthFraction > 0
+          ? {
+              width: '100%',
+              maxWidth: `calc(100vw * (1 - ${panelWidthFraction}))`,
```
rorychatt: if you had used ivy sheets with the BE logic, you would have not needed this hack.

You can totally build the UI you want using Ivy Code.

We have floating panels for the popup, we have sheet for the MCP search result.

I think the only thing that we need custom or to extract into it's own widget is the sidebar search results (or to just reimplement it as internal custom widget/view in the ivy docs)

rorychatt:
you wrote a bunch of custom frontend code.

We are a BE-first framework.
Meaning state handling etc should be done in the backend. You can define everything except maybe the search results in the modal using ivy views.

Allthough I also think you can do the search results with existing ivy components, just migth look a bit ugly -> we can fix this later

-> rewrite this to be a 100% ivy code. See comments above, they should help you.

src/Ivy.Docs.Shared/Internal/SmartSearch.cs
ArtemKhvorostianyi: Smart search widget is more like docs helper widget. Yes, we have internal folder, but it's shared across samples/docs. Smart search is docs-only tool, so it should be in this directory

rorychatt:
@ArtemKhvorostianyi good job! Let me sync with my team

rorychatt:
-> Need to iliminate visual jank, pop up dialog should have fixed position height-wise

PR #2418, opened 2026-02-28
[Security]: fix warning

defymecobra:
## Description of the Task
The primary goal of this pull request is to address a critical Dependabot security alert regarding a Regular Expression Denial of Service (ReDoS) vulnerability in the `minimatch` package. The vulnerable version (`< 9.0.6`) was being introduced into the project transitively through third-party dependencies such as `@glideapps/glide-data-grid` and the `eslint`/`typescript-eslint` stack.

## What Was Done
1. **Identified the Constraint:** It wasn't possible to simply run `npm install minimatch@latest` because `minimatch` is deeply nested within other packages. 
2. **Applied Package.json Overrides:** I added an `overrides` section to `src/frontend/package.json` to explicitly enforce the resolution of `minimatch` to version `9.0.6` (the patched version) across the entire dependency tree.
3. **Updated the Lockfile:** I ran `npm install` within the `src/frontend` directory to regenerate `package-lock.json`, thereby cementing the overridden dependency versions.
4. **Verified the Fix:** Running `npm ls minimatch` confirmed that version `9.0.6` is now successfully being used by all sub-dependencies.

## Why It Was Done
A ReDoS vulnerability in `minimatch` occurs when the engine processes extremely complex glob patterns against specific inputs, causing excessive polynomial backtracking. This can potentially hang or significantly slow down the Node.js runtime, leading to a denial of service. 

By enforcing version `9.0.6` via npm overrides, we ensure the patched code is applied immediately and everywhere. This protects our application's build tools and runtime from slowdowns without the need to wait indefinitely for parent packages (such as alpha versions of libraries) to update their internal dependency chains.

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -1194,30 +1194,6 @@
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
-    "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-      "version": "1.1.12",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

PR #2421, opened 2026-02-28
(video) feature: unified video player interface with YouTube IFrame API integration

### LINKED ISSUE #2420: (video) proposed interface and abstraction tsx
### Description

Refactor the video player widget around a clear interface and player abstraction so native video, YouTube, and future embed types share one contract and the main widget stays a thin router. No UI volume control in scope—only the programmatic interface and structure.

### 1. Shared control props

All player implementations receive the same control props:

- `volume?: number | null` — Initial playback volume, 0.0–1.0; `null` = browser/default
- `muted`, `autoplay`, `loop`, `controls` — Existing semantics, applied per player type
- `poster` (where applicable, e.g. native `<video>`)
- `width`, `height`, `id`, `source`/`url`

Each player is responsible for applying these in its own way (DOM properties, YouTube API, etc.).

### 2. Player type → component map

- **Detection:** Single helper, e.g. `getVideoType(url)` → `'native' | 'youtube' | 'vimeo' | ...`
- **Rendering:** Map type to component; no branching in the main widget:

  ```ts
  const players = { native: NativeVideoPlayer, youtube: YouTubePlayer, ... };
  const Player = players[getVideoType(url)];
  return <Player {...commonProps} url={url} />;
  ```

- **Placement:** Player components live under `widgets/primitives/` (e.g. same file or `video/NativeVideoPlayer.tsx`, `video/YouTubePlayer.tsx`). Main widget stays in `VideoPlayerWidget.tsx` and delegates.

### 3. Volume control abstraction

- **Contract:** Each player exposes a single programmatic “set volume” behavior, e.g. `setVolume(value: number)` (0–1), implemented per type:
  - **Native:** `videoRef.current.volume = value`
  - **YouTube:** `player.setVolume(value * 100)` (after API ready)
  - **Others:** Their API’s equivalent
- **Usage:** Parent or effect calls this when the `volume` prop is set or changes (e.g. via ref from `useImperativeHandle` or an `onReady({ setVolume })` callback). No UI volume button in scope—only this programmatic API.

### 4. YouTube IFrame API loading

- **Lazy, global, once:** e.g. `loadYouTubeAPI(): Promise<void>` in a small util (e.g. `lib/youtube-api.ts`).
- **Behavior:** First YouTube player mount triggers script inject and sets `window.onYouTubeIframeAPIReady` to resolve a cached promise; subsequent callers reuse that promise. No second script tag.
- **Scope:** Only the YouTube player path calls this; native and other embeds do not. No explicit unload of the script when no players are on screen.

## Advantages

- **Single contract:** One set of control props and one volume API across all player types; easier to reason about and test.
- **Extensibility:** New embed types (e.g. Vimeo) = new map entry + new player component; no growing if/else in the main widget.
- **Clear responsibilities:** Main widget only detects type and delegates; each player owns its own DOM/API and volume implementation.
- **Lazy YouTube script:** YouTube API loads only when at least one YouTube player is used; no cost when the widget is used only for native (or other) video.
- **Consistent volume semantics:** Same `volume` prop and `setVolume` semantics everywhere; backend/Ivy only deals with one model (0–1, null = default).

nwizugbesamson:
### Summary

This PR refactors the `VideoPlayer` widget's frontend implementation to establish a **shared props interface** across all player backends and introduce the **YouTube IFrame API** as the rendering layer for YouTube videos — replacing the previous `<iframe>` approach and laying the foundation for imperative prop controls (e.g. volume, seek).

### Why

The old implementation had no shared contract between native and YouTube player components. Each player received ad-hoc props with no type enforcement, and YouTube videos were rendered via a static `<iframe>` with no access to the player's JS API. That meant properties like `muted`, `loop`, `controls`, and `autoplay` had to be baked into the embed URL — and future controls (volume, playback rate, etc.) would be impossible without a full rewrite.

### What changed

**Shared `VideoPlayerProps` interface** (`video/types.ts`)

All player backends now receive the same typed props:

```ts
export interface VideoPlayerProps {
  id: string;
  validatedSrc: string;
  source: string;
  width?: string;
  height?: string;
  autoplay?: boolean;
  loop?: boolean;
  muted?: boolean;
  controls?: boolean;
  poster?: string | null;
  onError?: () => void;
}
```

**Player registry** (`players.ts`)

New players are registered in a typed map. Adding a new backend is a one-liner:

```ts
export const PLAYERS: Record<VideoType, ComponentType<VideoPlayerProps>> = {
  native: NativeVideoPlayer,
  youtube: YouTubePlayer,
};
```

**YouTube IFrame API loader** (`lib/youtube-api.ts`)

Lazy singleton — injects the YouTube script once per page, resolves when `YT` is ready:

```ts
loadYouTubeAPI().then(() => {
  playerRef.current = new window.YT.Player(containerRef.current, {
    videoId,
    playerVars: { autoplay: 1, mute: 1, controls: 1, loop: 1 },
    events: { onReady: (e) => { playerRef.current = e.target; } },
  });
});
```

This gives direct JS access to the player instance, so future controls like `player.setVolume(n)` or `player.seekTo(t)` require no structural changes — just a new prop and a call into `playerRef.current`.

### Extending with a new player

To add e.g. a Vimeo player:

1. Add `'vimeo'` to `VideoType` in `types.ts`
2. Create `VimeoPlayer: React.FC<VideoPlayerProps>`
3. Register it: `vimeo: VimeoPlayer` in `players.ts`
4. Add the detect rule in `VideoPlayerWidget`'s `EMBED_TO_VIDEO_TYPE` map

No other files need to change.

### C# API

Unchanged. `Source`, `Autoplay`, `Controls`, `Muted`, `Loop`, and `Poster` all serialize exactly as before.

Closes #2420


rorychatt:
@claude review this.

Explain me, does this add value when we already have support for Embed where we can plug in youtube URL-s.
Can we trust the added package?

Maybe the implementation with the shared controls needs to be done to the Embed widget instead for Youtube client and others?

rorychatt:
@nwizugbesamson My main worry is that video player was meant to display videos hosted by the user, not for embedding (we have separate Embed)

I will give this another thought on Monday and discuss with the Ivy team

PR #2422, opened 2026-02-28
[ListWidget]: remove parent padding hack

defymecobra:
### The Issue
In the `List` widget documentation, when lists were placed alongside other components (such as `Text.P` elements) inside a single default container (`TabPanel`), neighboring text elements lost their side padding and "stuck" directly to the edges of the container. This behavior broke the visual structure and padding consistency across the documentation.

### The Cause
The issue originated from a CSS/DOM mutation hack within `ListWidget.tsx`.
The `List` component attempted to implement a "Full-Bleed" layout (stretching the list and its hover background to the full width of the parent container, ignoring the parent's padding). 
To achieve this, the component used an aggressive approach:
1. During render, a `useLayoutEffect` hook was executed to programmatically locate the immediate parent element in the DOM tree.
2. The `remove-parent-padding` class was forcibly added to this parent.
3. The global `index.css` contained a broad `:has(> DIV.remove-parent-padding)` selector, which cascaded upwards and stripped all padding from the entire Demobox container.

**Consequence:** Because the list silently "destroyed" the padding of the wrapper container, neighboring text elements that relied on the parent's padding lost their boundaries.

### The Solution
**Chosen Approach: Clean Component Architecture (Removing DOM Mutations)**

1. **Frontend (`ListWidget.tsx`):** We completely removed the `useLayoutEffect` hook and the `remove-parent-padding` class from the list generator component.
   **Reasoning:** UI components should never mutate the DOM or styles of their ancestors. This fundamentally violates component encapsulation, leads to unpredictable behavior, and breaks the layout of sibling elements. `List` is now a standard block-level element that takes up 100% of its available width, without attempting to "break out" via CSS hacks.

### Changes Included
- `src/frontend/src/widgets/lists/ListWidget.tsx`: Removed DOM mutating hook and specific class.

image

### DIFF
--- a/src/frontend/src/widgets/lists/ListWidget.tsx
+++ b/src/frontend/src/widgets/lists/ListWidget.tsx
@@ -19,10 +19,7 @@ export const ListWidget = ({ children }: ListWidgetProps) => {
   });
 
   return (
-    <div
-      ref={parentRef}
-      className="relative h-full w-full overflow-y-auto"
-    >
+    <div ref={parentRef} className="relative h-full w-full overflow-y-auto">
       <div
         style={{
           height: rowVirtualizer.getTotalSize(),


defymecobra:
@claude review this

PR #2443, opened 2026-03-03
(tree): item action menu

rorychatt:
https://github.com/user-attachments/assets/fa61a1ae-814e-4413-8988-95f12609492d


src/frontend/src/widgets/tree/TreeItem.tsx
```diff
@@ -138,7 +164,26 @@ export const TreeItem: React.FC<TreeItemWidgetProps> = ({
           name={item.icon}
         />
       )}
-      <span className="truncate">{item.label}</span>
+      <span className="truncate flex-1">{item.label}</span>
```
dcrjodle, line 167: this is completely repeated above (duplicated code)

PR #2488, opened 2026-03-06
feat(progress): add explicit Indeterminate property

### LINKED ISSUE #2356: (progress): add explicit indeterminate mode
### Description

# Add explicit Indeterminate mode to Progress widget

## Summary

The Progress widget should support an explicit `Indeterminate` property that displays an animated "unknown progress" state — a bar that continuously animates to indicate work is happening without a known completion percentage.

Retool's progress bar has an `indeterminate` boolean that toggles this behavior independently of the value.

## Motivation

Currently, passing `null` as the value to `Progress` is the only way to indicate unknown progress, and the behavior is implicit. An explicit `Indeterminate` property is more expressive and covers additional use cases:

- **Loading with a known value** — show the indeterminate animation while waiting for the next progress update, then switch to determinate once the new value arrives. With `null`-only indeterminate, you lose the last known value when entering the loading state.
- **Explicit intent** — `new Progress(0).Indeterminate(true)` clearly communicates "work is happening, we just don't know how far along" vs. `new Progress(null)` which could mean "no value provided" or "reset."
- **Conditional toggling** — in polling or streaming scenarios, toggle indeterminate on/off based on whether the backend is responsive, while preserving the last progress value for display when the bar returns to determinate mode.

### Example scenarios

- **File upload** — show indeterminate while the server processes the uploaded file (progress is unknown), then switch to determinate when chunk-level progress is available
- **Build pipeline** — show indeterminate for steps that don't report progress (e.g., "Installing dependencies..."), then determinate for steps that do
- **Search/query** — show indeterminate while the database query runs, then hide or complete the bar when results arrive

## Proposed API

```csharp
public record Progress : WidgetBase<Progress>
{
    // ... existing properties ...
    [Prop] public bool Indeterminate { get; set; }
}
```

### Extension method

```csharp
public static Progress Indeterminate(this Progress progress, bool indeterminate = true)
    => progress with { Indeterminate = indeterminate };
```

### Usage

```csharp
// Explicit indeterminate — animated bar, no percentage shown
new Progress().Indeterminate().Goal("Loading...")

// Indeterminate with a stale value preserved for when it resolves
new Progress(lastKnownValue).Indeterminate(isLoading).Goal("Syncing...")

// Conditional based on state
var progress = UseState(0);
var isPolling = UseState(true);

return new Progress(progress.Value)
    .Indeterminate(isPolling.Value)
    .Goal(isPolling.Value ? "Waiting for server..." : $"{progress.Value}% Complete");
```

## Behavior

| `Indeterminate` | `Value` | Result |
|-----------------|---------|--------|
| `false` | `75` | Standard determinate bar at 75% |
| `false` | `null` | Empty bar or theme-default indeterminate (backward compat) |
| `true` | any | Animated indeterminate bar (value ignored visually) |

- When `Indeterminate` is `true`, the bar shows a continuous animation (e.g., a shimmer or sliding highlight) regardless of `Value`.
- The `Goal` text is still displayed — it serves as a status message (e.g., "Loading...", "Processing...").
- When `Indeterminate` transitions from `true` to `false`, the bar should animate smoothly to the current `Value` (if a `Duration` is set, per the transition duration proposal).

## Key Considerations

- Default is `false` to preserve backward compatibility.
- The existing behavior where `Value = null` implies indeterminate should be preserved for backward compatibility. `Indeterminate(true)` is an explicit opt-in that works regardless of the value.
- The frontend animation should respect `prefers-reduced-motion` — when active, show a static striped pattern or pulsing opacity instead of a sliding animation.
- This pairs well with the `Duration` proposal — the transition from indeterminate to determinate can use the configured duration.

## References

- Retool: `indeterminate` (boolean) on progress bar component
- Radix UI: `<Progress>` with `value={null}` for indeterminate state
- Ivy: current `Progress` widget — `null` value is implicitly indeterminate, no explicit toggle

rorychatt:
## Summary

Add an explicit `Indeterminate` property to the Progress widget that displays an animated progress bar when the completion percentage is unknown.

Closes #2356

---
Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/frontend/src/components/ui/progress.tsx
+++ b/src/frontend/src/components/ui/progress.tsx
@@ -26,9 +26,7 @@ const Progress = React.forwardRef<
     <ProgressPrimitive.Indicator
       className={cn(
         'h-full flex-1 bg-primary transition-all',
-        indeterminate
-          ? 'w-1/3 animate-indeterminate'
-          : 'w-full'
+        indeterminate ? 'w-1/3 animate-indeterminate' : 'w-full'
       )}
       style={
         indeterminate
--- a/src/frontend/src/widgets/progress/ProgressWidget.tsx
+++ b/src/frontend/src/widgets/progress/ProgressWidget.tsx
@@ -49,7 +49,8 @@ export const ProgressWidget: React.FC<ProgressWidgetProps> = ({
   width = 'Full',
   indeterminate = false,
 }) => {
-  const isIndeterminate = indeterminate || value === null || value === undefined;
+  const isIndeterminate =
+    indeterminate || value === null || value === undefined;
   const isCompleted = !isIndeterminate && value && value >= 100;
 
   const containerStyles: React.CSSProperties = {
@@ -85,7 +86,11 @@ export const ProgressWidget: React.FC<ProgressWidgetProps> = ({
             )}
           </Badge>
         )}
-        <Progress value={isIndeterminate ? undefined : value} indeterminate={isIndeterminate} className="bg-neutral/10" />
+        <Progress
+          value={isIndeterminate ? undefined : value}
+          indeterminate={isIndeterminate}
+          className="bg-neutral/10"
+        />
       </div>
     </>
   );


rorychatt:
@claude fix frontend formatting issues

PR #2516, opened 2026-03-08
(callout): add closable support to callout

ArtemKhvorostianyi:
Added OnClose prop which allow user to close callout. 
image

Added section to docs:

https://github.com/user-attachments/assets/1be07eff-d0dc-4c27-954e-ffc230d0e71c

And same changes to callout samples app, also used tabs approach for samples app

P.S. Didn't commit, just to review result, some callout test variants:
image

### DIFF
--- a/src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/12_Callout.md
+++ b/src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/12_Callout.md
@@ -180,35 +180,22 @@ public class ComplexContentCalloutView : ViewBase
 
 ### Closable callouts
 
-When you set an `OnClose` handler, the callout shows a close (X) button in the top-right corner. Clicking it fires the handler so you can hide the callout (e.g. by updating state).
+When you set an `OnClose` handler, the callout shows a close (X) button in the top-right corner. Clicking it fires the handler so you can hide the callout. Use [UseTrigger](../../03_Hooks/02_Core/17_UseTrigger.md) to control visibility.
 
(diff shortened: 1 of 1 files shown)


src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/12_Callout.md
```diff
@@ -178,6 +178,41 @@ public class ComplexContentCalloutView : ViewBase
...
+### Closable callouts
+
+When you set an `OnClose` handler, the callout shows a close (X) button in the top-right corner. Clicking it fires the handler so you can hide the callout (e.g. by updating state).
+
+```csharp demo-tabs
+public class ClosableCalloutView : ViewBase
+{
+    public override object? Build()
+    {
+        var showUpdateBanner = UseState(true);
```
rorychatt: useTrigger ?

src/frontend/src/widgets/primitives/CalloutWidget.tsx
```diff
@@ -43,13 +47,18 @@ const defaultIcons = {
 };
 
 export const CalloutWidget: React.FC<CalloutWidgetProps> = ({
+  id,
```
rorychatt, line 50: why is this needed?

ArtemKhvorostianyi: The server uses that id to know which Callout instance sent the event and runs the correct OnClose handler for that widget. Without id, the backend couldn’t tell which callout was closed when there are several on the page.

PR #2522, opened 2026-03-09
(routing): Allow dots in app IDs

### LINKED ISSUE #2469: Fix PathToAppIdMiddleware Dot-Based Extension Check
## Problem

In `src/Ivy/Core/Server/Middleware/PathToAppIdMiddleware.cs` line 73, `!appId.Contains('.')` is used to skip file extensions. This breaks for app IDs that contain dots (e.g., "app.v2").

## Solution

1. Read `PathToAppIdMiddleware.cs` fully.
2. Replace the `Contains('.')` check with a more robust file extension check. Use `Path.HasExtension(appId)` or check against a known set of static file extensions (e.g., `.js`, `.css`, `.html`, `.ico`, `.map`, `.json`, `.woff`, `.woff2`, `.svg`, `.png`, `.jpg`).
3. `Path.HasExtension` is slightly better but still has edge cases. Best approach: check if the path matches a known static file pattern or has a known static file extension.

**File:** `src/Ivy/Core/Server/Middleware/PathToAppIdMiddleware.cs`

## Tests

```
cd D:\Repos\_Ivy\Ivy-Framework
dotnet test
```

## Finish

Commit!

zachwolfe:
### Changes
- '.’ is now allowed in app IDs (fixes https://github.com/Ivy-Interactive/Ivy-Framework/issues/2469).
- ‘:’, ‘%’ and ‘\’ are no longer allowed.
- App IDs that end with a known static file extension are now never allowed (previously they were sometimes allowed)
- Added error message for invalid app IDs that were added after a hot reload (instead of only on full restart). This error is fatal on start (like it was before). On hot-reload, execution will continue but the apps will not be added to AppRepository.
- PathToAppIdMiddleware now shares more of the same logic with the check during restart/reload.
- Routes such as those required for datatables still work

Fixes #2469

### DIFF
--- a/src/Ivy.Test/AppRoutingHelpersTests.cs
+++ b/src/Ivy.Test/AppRoutingHelpersTests.cs
@@ -1,0 +1,398 @@
+using Ivy.Core.Apps;
+
+namespace Ivy.Test;
+
+public class AppRoutingHelpersTests
+{
+    #region ValidateAppId - Valid Cases
+
+    [Theory]
+    [InlineData("dashboard")]
+    [InlineData("my-app")]
+    [InlineData("app_name")]
+    [InlineData("app123")]
+    [InlineData("MyApp")]
+    public void ValidateAppId_SimpleValidIds_ReturnsValid(string appId)
+    {
+        var reservedPaths = new HashSet<string>();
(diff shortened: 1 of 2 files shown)


zachwolfe:
@claude review please

PR #2534, opened 2026-03-10
(validation): improve validation strategy

ArtemKhvorostianyi:
Closed #2238 since it was crashed after resolving merge conflicts (pushed all changes to file changes)
Since we have three input widgets where users can type from the keyboard, we need proper validation. These widgets are TextInput, Field, and Form.

The TextInput widget includes variants such as password, email, URL, and telephone.
The Field widget also supports password, email, URL, and telephone variants.
The Form widget provides additional functionality: it can detect field types based on labels, and users can also supply custom validation.

The main logic is centered around text input: since every input — whether it is a Field or a Form field — is still based on the TextInput widget, the validation logic is built around the TextInput widget itself.

The main goal of this PR is not to change any API usage, but to implement validation logic for password, email, URL, and telephone inputs.

So I built a validation module with the main validation logic:
/Ivy/Validation/Validators.cs

Then I wired it up with the TextInput backend part so we can call these methods from textInput.cs.
When a TextInput is wrapped with WithField, or used inside Forms, we validate the text input first and then apply the invalid state if necessary.

There are no API changes, and I tried to minimize the amount of added code.

That is the main goal of the PR, but I also had to do some additional work on Forms validation, because it supports custom validation logic, which was often broken. I updated its code as well.

It works, and it works smooth now

**P.S. Validation of text inputs works in both ways:
email.ToEmailInput() and return new TextInput.Variant(TextInputVariants.Email).WithField()**

Text validation:  ( email.ToEmailInput(). for example)

image

image

Text validation:  ( return new TextInput.Variant(TextInputVariants.Email).WithField() for example)
image

image

Field validation:

image

image

Form validation:

image

image

### DIFF
--- a/src/Ivy/Views/Forms/FormBuilder.cs
+++ b/src/Ivy/Views/Forms/FormBuilder.cs
@@ -390,7 +390,7 @@ public class FormBuilder<TModel> : ViewBase
             {
                 return submitReceiver.Receive(unit =>
                 {
-                     _ = OnSubmit();
+                    _ = OnSubmit();
                 });
             }
             return null;


rorychatt:
Does form submission strategy (see latest changes): work with this too?

```
# Form Submission Strategy Breaking Changes - v1.2.18

## Summary

In previous versions, a `Form` would implicitly render a "Save" button and only validate and commit state changes when that button was clicked. With version 1.2.18, the framework introduces a `FormSubmitStrategy` enum to explicitly control when a form validates and submits changes to its bounded state. The default is still `OnSubmit`, but if you rely on immediate state updates, this logic has been formalized. Additionally, overriding the submit button and its click handler has changed due to the `EventHandler` wrapper Refactoring.

## What Changed

### Form Submit Strategy

You can now use the `.SubmitStrategy()` fluent method on a `FormBuilder` to dictate how state is updated:

- `FormSubmitStrategy.OnSubmit`: (Default) Renders the submit button. State is validated and updated only when the user clicks the button.
- `FormSubmitStrategy.OnBlur`: The submit button is hidden. State is validated and updated whenever the user tabs away or clicks out of a field.
- `FormSubmitStrategy.OnChange`: The submit button is hidden. State is validated and updated immediately upon every keystroke or selection change.

### The Submit Button and `OnSubmit` Callback

Previously, you might have used `.HandleSubmit()` on the button builder. Now, there is a dedicated `.OnSubmit()` callback directly on the `FormBuilder` to handle async operations, and the click handler for a custom button builder uses `.OnClick()` due to the `EventHandler` update.

#### Before

```csharp
model.ToForm()
    .SubmitBuilder(isLoading => new Button("Apply")
        .Loading(isLoading)
        .HandleClick(_ => { /* custom logic */ return default; }))
```

#### After

```csharp
model.ToForm()
    .SubmitBuilder(isLoading => new Button("Apply").Loading(isLoading))
    .OnSubmit(async (m) => {
        // Runs after successful validation, before state is committed
        await myService.SaveAsync(m);
    })
```

## How to Refactor

1. If you need a form to update state immediately without a "Save" button, chain `.SubmitStrategy(FormSubmitStrategy.OnChange)` or `.SubmitStrategy(FormSubmitStrategy.OnBlur)`.
2. Find uses of `.HandleSubmit()` on the `SubmitBuilder` button and move the logic to the `FormBuilder`'s `.OnSubmit()` method. Change the button's action handler to `.OnClick()` if you still need it for standard buttons, but for Forms, `.OnSubmit(Func<TModel, Task>)` is the correct hook for validation-aware submissions.

## Verification

After applying these changes, run:

```bash
dotnet build
```
```

File: src/Ivy/Core/WidgetTree.cs
Comment (rorychatt) [inline src/Ivy/Core/WidgetTree.cs:1]: why we need changes here?

File: src/Ivy/Views/Forms/FormBuilder.cs [hunk 1]
```diff
@@ -332,7 +332,9 @@ private Expression<Func<TModel, object>> CreateSelector(string name)
     {
         var currentModel = context.UseState(() => StateHelpers.DeepClone(_model.Value), buildOnChange: false);
 
-        var validationSignal = context.UseSignal<FormValidateSignal, Unit, bool>();
+        // Per-form signal instances so Submit validates only this form's fields (not every field on the page).
```
rorychatt: not sure about this

rorychatt: feel like we are changing behavior.

See my other comment and look up how we now handle form submission.

src/Ivy/Views/Forms/FormBuilder.cs
```diff
@@ -383,14 +387,13 @@ async Task<bool> OnSubmit()
 
         var fieldViews = bindings.Select(e => e.fieldView).ToArray();
 
-        var submitReceiver = context.UseSignal<FormSubmitSignal, Unit, Unit>();
         context.UseEffect(() =>
         {
             if (_submitStrategy is FormSubmitStrategy.OnBlur or FormSubmitStrategy.OnChange)
             {
-                return submitReceiver.Receive(unit =>
+                return submitSignal.ReceiveWithId(Guid.NewGuid(), _ =>
```
rorychatt: don't understand why we need this

ArtemKhvorostianyi: It was hot fix after merged Ivy Studio. Forms contain submitting issue: when user tries to submit  form, all the forms from page validates, not only the current one. 

Screenshot from branch main:

image

ArtemKhvorostianyi:
@rorychatt I reverted changes about logic of form submitting. Validation still works, have a look at the screenshot to understand issue:

image

I suggested solution, have a look at the last commit diff.

PR #2556, opened 2026-03-10
enums: use singular vs plural

rorychatt:


rorychatt:
@claude review this. Have we missed anything?

PR #2606, opened 2026-03-14
(datatable): update docs clarify usage of rowactions and onrowaction

ArtemKhvorostianyi:
"instead of strings we could have enum values!! (As much as possible code should be failing in compile time)

Make it so we use fluent API in all of the code examples. not MenuITem(..., "edit") but MenuItem().Tag()"

Screenshot 2026-03-14 at 05 55 32

Works the same as before


rorychatt:
@claude any other spots in docs that could use enums instead of strings for menuoptions?

create an issue if you find any and assign @ArtemKhvorostianyi

PR #2615, opened 2026-03-15
feat: make server listening interface configurable

rorychatt:
This PR makes the server listening interface configurable via the \HOST\ environment variable or \--host\ command-line argument. It also automatically defaults to wildcard binding (\*\) when the \PORT\ environment variable is set, fixing deployment issues on platforms like Sliplane.


rorychatt:
@claude review this

rorychatt:
@claude do the suggested changes

PR #2639, opened 2026-03-16
fix(widgets): cache React.lazy wrappers for external widgets to preve…

rorychatt:
…nt full rerender


rorychatt:
@claude review this

PR #2677, opened 2026-03-17
fix(chat): enable Enter key to submit messages in Chat widget

rorychatt:
Pressed Enter in the chat input now correctly submits the prompt, aligning with recent form submission strategies that move away from native form wrappers. This was tested and verified on the documentation site's live demo.


rorychatt:
@ArtemKhvorostianyi test before and after. I think this PR is unneeded

ArtemKhvorostianyi:
> @ArtemKhvorostianyi test before and after. I think this PR is unneeded

Yes, it's unneeded
Tested from branch main and this branch, in both branches we are able to send message to chat with Enter key. So it works fine without this PR and PR can be closed

PR #2680, opened 2026-03-18
Fix: restored borders for ghost variant in Select components

rorychatt:
This PR fixes a bug where the ghost variant for the Select inputs accidentally stripped the default borders due to a recently applied border-transparent class. By removing border-transparent, the Select ghost variant properly inherits its expected visual borders while remaining transparent in background.


rorychatt:
@claude review this

PR #2692, opened 2026-03-18
fix: fix AppAttribute path argument syntax error

rorychatt:
Fix AppAttribute path argument syntax error in SignatureInputApp.cs


rorychatt:
@claude review this

PR #2696, opened 2026-03-19
fix(ui): use gap instead of space-x in dialog, alert-dialog and sheet footers

rorychatt:
Fixed missing spacing between form actions (Cancel/Save) by switching from \space-x-2\ to \gap-2\ in DialogFooter, AlertDialogFooter, and SheetFooter. This provides more reliable spacing when some siblings (like validation messages) are empty or growing.


rorychatt:
@claude review this

PR #2767, opened 2026-03-23
feat: adding a base-path support to ivy-framework

swaner:
Adding support to set a base path prefix like /foo/bar (can be tested by executing `dotnet run -- --path-base /foo/bar`)


artem-ivy-ai:
### Redeploy failed

Deployment stopped due to an error. I’m attaching the latest Sliplane events below.

### Logs

```
Redeploy triggered for 0 service(s).
```

rorychatt:
@claude review this. How would user set a base path for an ivy project using env variables / config files? Is this implemented in this PR?

src/Ivy/Auth/DefaultAuthApp.cs
```diff
@@ -145,7 +145,12 @@ public class OAuthFlowView(AuthOption option) : ViewBase
 
         var state = this.UseState(() => registry.RegisterPending(args.ConnectionId, option.Id ?? ""));
 
-        var oauthUriBuilder = new UriBuilder($"{args.BaseUrl}/ivy/auth/oauth-login")
+        // Build OAuth URI with path base prefix if present
+        var oauthPath = string.IsNullOrEmpty(args.PathBase)
```
rorychatt: we already have BaseUrl -> how is this different from pathbase? 

@claude tell me

src/Ivy/Apps/AppContext.cs
rorychatt: 1. Let's see if we can have BasePath part of BaseUrl here
2. Let's rename PathBase -> BasePath

src/Ivy/Auth/DefaultAuthApp.cs
swaner: No, I changed it to baseUrl. But actually I notice that the microsoft already calls it path base on request object.  So maybe it makes sense to follow that after all? Should I revert? 

See:  https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequest.pathbase?view=aspnetcore-10.0#microsoft-aspnetcore-http-httprequest-pathbase
and
https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-10.0#microsoft-aspnetcore-builder-usepathbaseextensions-usepathbase(microsoft-aspnetcore-builder-iapplicationbuilder-microsoft-aspnetcore-http-pathstring)

rorychatt: `BasePath` -> this is what AI-s will hallucinate and pathbase sounds like "legacy" code that somehow got into latest .NET. We should have this API, and then we can do whatever Microsoft is using in non-api section

(I did research this just now)

rorychatt: otherwise BaseUrl should be  UrlBase etc
it just makes no sense to have it called PathBase. It's not base of path. It's path of base

rorychatt:
(app context): needs documentation

rorychatt:
@claude review this

rorychatt:
Added documentation for App Context. Live preview: http://localhost:5015/onboarding/concepts/app-context

rorychatt:
@claude review this

artem-ivy-ai:
### Deleted

rorychatt:
(base path): set via env

The base path can only be set via CLI (--path-base), but PORT, VERBOSE, and HOST all support environment variables (Server.cs:98-111). For containerized deployments, env vars are typically preferred over CLI args.

Suggested fix in Server.cs around line 112:

if (_args.BasePath == null && Environment.GetEnvironmentVariable("BASE_PATH") is { } basePath)
{
    _args = _args with { BasePath = basePath };
}

PR #2776, opened 2026-03-24
fix(tiptap): remove default export to eliminate MIXED_EXPORTS warning

zachwolfe:
## Summary

Removes the default export from the Tiptap widget's frontend entry point to eliminate the `MIXED_EXPORTS` bundler warning. The module was exporting `TiptapInput` as both a named and default export, which triggers warnings in bundlers like esbuild/Rollup about mixing export styles. The named export alone is sufficient since consumers already use it by name.


artem-ivy-ai:
### Staging preview

- **Docs:** [Docs](https://ivy-staging-docs-fixtiptap-mixed-exports-warning.sliplane.app)
- **Samples:** [Samples](https://ivy-staging-samples-fixtiptap-mixed-exports-warning.sliplane.app)

rorychatt:
Good catch!

PR #2788, opened 2026-03-24
fix: let AppShell handle unknown nav app IDs instead of setting 404 …

dcrjodle:
This solved the refresh bug


rorychatt:
@claude review this. I am thinking we are treating a symptom and not the real cause. What do you think?

artem-ivy-ai:
### Deleted

### Logs

```
Deleted 2 service(s).
```

PR #2793, opened 2026-03-24
Fix MultipleSelector dropdown scroll position on open

nielsbosma:
## Summary

- Reset scroll position to top when MultipleSelector dropdown opens, preventing it from retaining scroll position from previous interactions


artem-ivy-ai:
### Deleted

### Logs

```
Deleted 1 service(s).
```

PR #2797, opened 2026-03-25
Fix shell=false query parameter app selection

nielsbosma:
## Summary
- Fix ?shell=false query parameter to select the first visible app when AppShell DefaultAppId is not configured


artem-ivy-ai:
### Deleted

### Logs

```
Deleted 2 service(s).
```

PR #2804, opened 2026-03-25
fix: handle exceptions in WidgetSerializer default value comparison

dcrjodle:
## Summary

- **Fixes a crash** (`InvalidOperationException: Trying to access an uninitialized WidgetBase Id`) that occurs when `System.Text.Json` lazily builds type metadata during the first serialization of an `AbstractWidget` subclass (e.g. `Button`)
- The `AddDefaultValueComparison` modifier constructs a default instance of each type to compare property values for diff-based serialization. This default instance is never registered in the widget tree, so its `Id` is null — and the `Id` getter throws by design as a safety check for real widgets
- Wraps `property.Get(defaultInstance)` in a try-catch so properties that throw on uninitialized instances are skipped (always serialized), preserving the safety check on `Id` for actual widget usage

## Test plan

- [x] `dotnet build` passes
- [ ] Verify the crash no longer occurs when serializing a widget type for the first time in a fresh process

### DIFF
--- a/src/Ivy/Core/WidgetSerializer.cs
+++ b/src/Ivy/Core/WidgetSerializer.cs
@@ -80,7 +80,16 @@ public static class WidgetSerializer
             if (property.Get == null)
                 continue;
 
-            var defaultValue = property.Get(defaultInstance);
+            object? defaultValue;
+            try
+            {
+                defaultValue = property.Get(defaultInstance);
+            }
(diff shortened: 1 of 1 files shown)


rorychatt:
@claude review this

artem-ivy-ai:
### Deleted

PR #2807, opened 2026-03-26
(autoscroll): add new AutoScroll container widget

ArtemKhvorostianyi:
### Summary

Introduces a standalone AutoScroll widget: a height-constrained, scrollable region that keeps the view pinned to the bottom as children change, with an Enabled flag to turn off follow behavior so users can read earlier content without being pulled down.

https://github.com/user-attachments/assets/eddce637-94a7-4e89-a14b-474fb6ddc39f

**Implementation**

Ivy.AutoScroll (WidgetBase): arbitrary children; Enabled (default true); AutoScrollExtensions.Enabled(...) for fluent configuration; inherits usual Width / Height from WidgetBase.
Frontend (AutoScrollWidget.tsx): maps props to layout; uses useAutoScroll with content: children and enabled.
use-auto-scroll: adds optional enabled; when false, growth/resize no longer forces scroll; when enabled flips back to true, scrolls to bottom once. Existing ChatMessageList behavior unchanged (defaults to enabled).
Registration: "Ivy.AutoScroll" in widgetMap and primitives barrel export.

### DIFF
--- a/src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/26_AutoScroll.md
+++ b/src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/26_AutoScroll.md
@@ -16,11 +16,13 @@ searchHints:
 Keep the newest content in view inside a fixed-height scroll area — ideal for live logs, activity feeds, or any append-only output in your [views](../../01_Onboarding/02_Concepts/02_Views.md).
 </Ingress>
 
-The `AutoScroll` [widget](../../01_Onboarding/02_Concepts/03_Widgets.md) is a scrollable container that automatically scrolls to the bottom when its children grow. Pass any widgets as children (not only text). Give the container an explicit [Size](../../04_ApiReference/Ivy/Size.md) — especially height — so the browser can show a scrollbar and the follow behavior can measure overflow.
+The `AutoScroll` [widget](../../01_Onboarding/02_Concepts/03_Widgets.md) is a scrollable container that follows new content to the bottom when its children grow. Pass any widgets as children (not only text). Give the container an explicit [Size](../../04_ApiReference/Ivy/Size.md) — especially height — so the browser can show a scrollbar and the follow behavior can measure overflow.
 
 ## Basic Usage
 
 `AutoScroll` takes a `params` array of children. Use `Height` (and optionally `Width`) from `WidgetBase` so the scroll region is bounded.
+
+When you build children from LINQ (for example `lines.Select(…)`) use `AutoScroll.FromChildren(…)` so you do not need `ToArray<object>()` — C# does not pass a `Widget[]` as `params object[]`.
 
 ```csharp demo-tabs
 public class AutoScrollBasicDemo : ViewBase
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Deployed

**Docs:** [https://ivy-staging-docs-2389-autoscroll-new-container-widget.sliplane.app](https://ivy-staging-docs-2389-autoscroll-new-container-widget.sliplane.app)
**Samples:** [https://ivy-staging-samples-2389-autoscroll-new-container-widget.sliplane.app](https://ivy-staging-samples-2389-autoscroll-new-container-widget.sliplane.app)

src/Ivy.Samples.Shared/Apps/Widgets/Primitives/AutoScrollApp.cs
```diff
@@ -0,0 +1,46 @@
...
+            "[AutoScroll] Turn off “Follow new lines” to read earlier lines without being pulled down.",
+            "[AutoScroll] Append lines to simulate a live log or activity feed.");
+
+    protected override object? BuildSample()
+    {
+        var logLines = UseState(AutoScrollLogSeed());
+        var autoscrollFollow = UseState(true);
+
+        var autoScrollPanel = new AutoScroll(
+                logLines.Value.Select(line => Text.Muted(line)).ToArray<object>())
```
rorychatt: I don't like this Cast -> is this needed?

Why can we just directly pass the array?

src/Ivy.Samples.Shared/Apps/Widgets/Primitives/AutoScrollApp.cs
```diff
@@ -0,0 +1,46 @@
...
+            .Height(Size.Px(100))
+            .Width(Size.Full())
+            .Enabled(autoscrollFollow.Value);
```
rorychatt: basic example should maybe not contain this Enabled API

src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/26_AutoScroll.md
```diff
@@ -0,0 +1,73 @@
...
+```csharp demo-tabs
+public class AutoScrollBasicDemo : ViewBase
+{
+    public override object? Build()
+    {
+        var lines = UseState(ImmutableArray.Create("First line", "Second line"));
+
+        return Layout.Vertical()
+            | new AutoScroll(
+                  lines.Value.Select(l => Text.Muted(l)).ToArray<object>())
```
rorychatt: same comment here, why we have to <object>?

src/Ivy.Docs.Shared/Docs/02_Widgets/01_Primitives/26_AutoScroll.md
```diff
@@ -0,0 +1,73 @@
...
+              .Width(Size.Full())
+            | new Button("Add line", () =>
+                lines.Set(lines.Value.Add($"Line at {DateTime.Now:HH:mm:ss}")));
+    }
+}
+```
+
+## Enabled follow
+
+When `Enabled` is `true` (default), new content scrolls the view to the bottom. When `false`, the user can scroll manually but the view does not jump when children update — useful when someone is reading older lines while new data arrives.
```
rorychatt: .Enabled() must be default and be true.
(also this is confusing with .Disabled() API ->
maybe we should just have a ScrollBehavior -> see what enums we already haver in the framework

ArtemKhvorostianyi: I see the point. 
I decided to implement Disabled API instead of Enabled prop, so it's now available to 'turn off' auto scrolling with setting .Disabled(true)

artem-ivy-ai:
### Deleted

PR #2817, opened 2026-03-26
fix(fe): ensure SpacerWidget grows by default

rorychatt:
Fixes the issue where `Spacer` width growth was failing because its default `Width = Size.Grow()` was not being serialized by the backend. The fix adds resilient defaults to `SpacerWidget.tsx` to ensure it grows even when the prop is missing.


artem-ivy-ai:
### Deleted

rorychatt:
@claude  review this. explain why this worked

PR #2820, opened 2026-03-26
Fix chart rendering issues

nielsbosma:
## Summary
- Fix chart rendering with non-string dimension values in generateDataProps
- Fix case-sensitive dataKey matching in LineChart and BarChart series


artem-ivy-ai:
### Deploy failed

Deployment stopped due to an error. I'm attaching the latest Sliplane events below.

### Logs

```
Internal server error
```

PR #2834, opened 2026-03-26
[SelectInput]: fix action icons positioning and interaction in SelectSingle

ArtemLazarchuk:
## Description
This PR addresses layout and interaction issues in the [SelectSingleVariant](cci:1://file:///Users/mak/git/Ivy-Interactive/Ivy-Framework/src/frontend/src/widgets/inputs/SelectSingleVariant.tsx:23:0-285:2) component.
### Changes Made:
- **Layout Fix:** Previously, action icons (Clear, Invalid, Loading) were positioned using an external absolute container that incorrectly overlaid the input text. They have now been moved directly inside the `SelectTrigger` to ensure stable and consistent positioning relative to the input field.
- **Interaction Bugfix:** Fixed an issue where clicking the "Invalid" icon or the "Loading" spinner would erroneously trigger the Radix Select dropdown. Added `onPointerDown` stop propagation to the icon container to isolate these actions from the main trigger.
- **Single-Click Clear:** Resolved the issue where the "Clear" (X) button required two clicks to function. It now correctly clears the state on the first click without triggering the popover.
- **Code Simplification:** Unified the event-handling logic by moving `stopPropagation` and `preventDefault` to a parent container, reducing code duplication across individual icons.
## Impact:
Improves UI consistency and prevents accidental dropdown activation when interacting with control icons (Clear/Invalid/Loading).

https://github.com/user-attachments/assets/67f96297-58f3-46d7-b68e-438017e4933a

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Widgets/Inputs/SelectInputApp.cs
+++ b/src/Ivy.Samples.Shared/Apps/Widgets/Inputs/SelectInputApp.cs
@@ -145,7 +145,7 @@ public class SelectInputVariantsExample : ViewBase
 
     public override object? Build()
     {
-        var colorState = UseState<Colors?>(Colors.Red);
+        var colorState = UseState(Colors.Red);
         var colorArrayState = UseState(Array.Empty<Colors>());
         var nullableColorArrayState = UseState<Colors[]?>(() => null);
         var iconsState = UseState<string>("bold");


src/Ivy.Samples.Shared/Apps/Widgets/Inputs/SelectInputApp.cs
```diff
@@ -145,7 +145,7 @@ private enum Colors { Red, Green, Blue, Yellow }
 
     public override object? Build()
     {
-        var colorState = UseState(Colors.Red);
```
rorychatt, line 148: let's keep the enum type of this usestate

rorychatt: or why did you change this?

rorychatt:
@claude review this

src/Ivy.Samples.Shared/Apps/Widgets/Inputs/SelectInputApp.cs
ArtemLazarchuk: > or why did you change this?

I changed this because enums use a default value. For example, if the default is Color.Red, and I change it to green, then press clear, it resets back to the initial default value (Red) instead of actually clearing it.

rorychatt: Sounds like a bug then, maybe
the purpose of enum there was to show that the type works with color enum values

ArtemLazarchuk:
@rorychatt In C#, an Enum (just like int or bool) is a Value Type, meaning it cannot be empty (null) by default. To make the field clearable, we must use the nullable version — <Colors?>. This is the only way for the component to correctly represent a 'nothing' state. Maybe...

rorychatt:
@ArtemLazarchuk if we have a separate col for nullable colors enum, then we keep this as colors. if we don't -> we can make this Colors?

ArtemLazarchuk:
We no longer need a separate column for nullable — this is now inferred automatically from the type.

- `Colors?` → nullable: Clear is available, and deselect is allowed.
- `Colors` → non-nullable: no Clear, and deselect is blocked.

### What was improved

- **Backend**
  - `SelectInput` now always serializes `nullable` to the frontend.
  - Added safe handling for empty enum input to prevent `Enum.Parse` crashes with nullable enum flows.

- **Frontend**
  - Clear (`X`) in Select is shown only when `nullable = true`.
  - Removed Clear button from **Toggle (single-select)**.
  - Improved deselect behavior:
    - nullable → deselect sends `null`
    - non-nullable → deselect is ignored (current value remains selected)
    
  https://github.com/user-attachments/assets/f75067e9-77db-4b39-8263-3a0ea837fa5b

artem-ivy-ai:
### Deleted

PR #2837, opened 2026-03-26
Add missing closing braces in `Ivy.Widgets.{Leaflet, Tiptap}` package.json files

zachwolfe:
156fa65 fixed this for the other two widgets only. I found this by building `Ivy/Ivy.slnx`, which uses Tiptap.

@rorychatt Do you think these external widget projects should be added to `Ivy-Framework.slnx` to make it easier to detect issues like this? Or would that slow down the build too much?


artem-ivy-ai:
### Deploying...

rorychatt:
@zachwolfe we want to keep the core package as light as possible.

PR #2843, opened 2026-03-29
(datatable): keep apps in samples in single app and fix multi-agg visibility

ArtemKhvorostianyi:
### Samples apps refactoring
Before:
image

Now:
Screenshot 2026-03-29 at 17 15 46

### Footer multi aggregations issues

Then I found issue with footer aggregration actions, it wasn't switchable:

https://github.com/user-attachments/assets/56ec9bf7-82dd-41f0-8c6c-9fe7647a075d

in app there are two options: total and average, but it wasn't able to switch between them. The logic was that on agg user could click and call expandable, which should show agg options
Fixed:

https://github.com/user-attachments/assets/88945ef8-6cdd-4b44-9602-188d4cffa320

### DIFF
--- a/src/frontend/src/widgets/dataTables/DataTableFooter.tsx
+++ b/src/frontend/src/widgets/dataTables/DataTableFooter.tsx
@@ -138,13 +138,13 @@ const FooterCell: React.FC<{
 
   useEffect(() => {
     if (!open) return;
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Deleted

PR #2848, opened 2026-03-29
[01088] Icons.X in Markdown should render with icon display next to it

nielsbosma:
## Summary

- Adds icon rendering for the `Icons.X` pattern in markdown inline code blocks
- When `Icons.ChevronDown` appears in inline code, the actual Lucide icon is displayed next to the code text
- Styled as `inline-flex` with centered alignment and `0.25em` gap

## Changes

- **MarkdownRenderer.tsx** — Added `Icons.X` pattern detection in the inline `code` component using regex. When matched, wraps the code + `<Icon>` in an inline-flex span.
- **MarkdownRenderer.test.ts** — Added test cases for the icon pattern regex (positive and negative matches).

## Commits

- `4c1586c` — Icon rendering implementation
- `2b323cf` — Tests and refinements


artem-ivy-ai:
### Deleted

PR #2850, opened 2026-03-30
[01100] Fix Assembly.Location for single-file desktop publishing

nielsbosma:
## Problem

`DeployDesktop` workflow fails with IL3000 errors during single-file publish because `Assembly.Location` returns an empty string in single-file apps, breaking OAuth token handler discovery in `Server.cs`.

## Solution

Replace `Assembly.Location`-based directory discovery with `AppContext.BaseDirectory` in `DiscoverAndRegisterOAuthTokenHandlers()`. This works correctly for both single-file and normal publish scenarios.

## Commits

- `a78392f8` — Replace Assembly.Location with AppContext.BaseDirectory for single-file publish
- `0b849030` — Fix AppContext namespace collision with System.AppContext

### DIFF
--- a/src/Ivy/Server.cs
+++ b/src/Ivy/Server.cs
@@ -277,7 +277,7 @@ public class Server
         try
         {
             // Load all "Ivy.Auth.*" assemblies eagerly to ensure their handlers are registered in the DI container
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Deleted

PR #2858, opened 2026-03-30
[01107] Fix CreateContext state propagation to child components

nielsbosma:
## Summary

- **Problem:** `CreateContext<T>()` cached context instances by type and never re-evaluated the factory on re-render, causing child components using `UseContext<T>()` to receive stale values when parent state changed
- **Solution:** Transformed `CreateContext` into a proper hook using `_callingIndex` tracking that re-evaluates the factory on every render, updating the service container so children always get fresh context
- **Tests:** Added `CreateContext_ReEvaluatesFactoryOnReRender` and `CreateContext_UpdatesServiceContainerForUseContext` unit tests

## Commits

- `7c3ed55c` — Fix CreateContext state propagation to child components

## Verification

- ✅ DotnetBuild
- ✅ DotnetFormat
- ✅ DotnetTest
- ✅ FrontendLint
- ✅ IvyFrameworkVerification

### DIFF
--- a/src/Ivy.Docs.Shared/Docs/03_Hooks/02_Core/12_UseContext.md
+++ b/src/Ivy.Docs.Shared/Docs/03_Hooks/02_Core/12_UseContext.md
@@ -325,12 +325,57 @@ public class ConditionalAuthView : ViewBase
 }
 ```
 
+## Reactive Context Pattern
+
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Deleted

PR #2860, opened 2026-03-30
[01140] Dialog Width Support In Ivy

nielsbosma:
## Summary

- **Fix:** Dialog's custom width was being silently capped by the `max-w-xl` CSS class (36rem) on `DialogContent`. When an inline width style is present, `maxWidth` is now also set inline to override the CSS constraint, allowing dialogs to render at any custom width.

## Commit
- `34a85b2` — Fix Dialog custom width being capped by max-w-xl CSS class


artem-ivy-ai:
### Deleted

PR #2869, opened 2026-03-31
[TextInput]: textarea height prop resizes container only, not content area

ArtemLazarchuk:
## Problem
When a `TextareaInput` had an explicit `.Height(...)` set, dragging the native resize handle would only move the border of the wrapper `<div>` — the actual textarea content area stayed the same size. This happened because the height style was applied to the wrapper `<div>` while the `<textarea>` used `h-full` to fill it. The browser's resize handle sets an inline style on the `<textarea>` element itself, not on the parent, so the two were out of sync.

## Fix
Move the height style from the wrapper `<div>` directly onto the `<Textarea>` element. The wrapper now only carries the width, and its height grows naturally to wrap the textarea. `h-full` is preserved on the textarea only when no explicit height prop is provided.

https://github.com/user-attachments/assets/42dffb6b-73a2-432a-aa44-0f9a8fd38793


artem-ivy-ai:
### Deleted

PR #2872, opened 2026-03-31
Fix: CodeBlock Layout & API Consistency

rorychatt:
This PR fixes a layout bug where CodeBlock widgets were incorrectly truncating text instead of expanding. It also ensures API consistency by adding width/height support to all specialized text widgets on the frontend and adding ergonomic .Grow() and .Height() extensions to TextBuilder on the backend.


src/frontend/src/widgets/primitives/CodeBlockWidget.tsx
```diff
@@ -113,7 +113,7 @@ const CodeWidget: React.FC<CodeWidgetProps> = memo(
     /** Pre (container): padding, dimensions, typography. Padding here applies to all lines. */
     const preStyle = useMemo<CSSProperties>(() => {
       const style: CSSProperties = {
-        ...getWidth(width),
```
rorychatt, line 116: this is bad

artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2889, opened 2026-03-31
Fix/restore routing dots

rorychatt:


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2898, opened 2026-04-01
[01356] Fix code block rendering without language specifier

nielsbosma:
## Changes

Fixed fenced code blocks without a language specifier rendering as inline code instead of block-level code elements. The root cause was that `MarkdownRenderer.tsx` used `!className` to detect inline code, but react-markdown doesn't set a className on code elements without a language — causing fenced blocks to be misidentified as inline. Fixed by checking the AST parent node (`node.parent.tagName === 'pre'`). Also removed the `match` guard in `CodeBlock.tsx` so no-language blocks reach the SyntaxHighlighter with a fallback language of `"text"`.

## API Changes

None.

## Files Modified

- **src/frontend/src/components/MarkdownRenderer.tsx** — Fixed inline detection logic to use AST parent node check instead of className
- **src/frontend/src/components/markdown/CodeBlock.tsx** — Removed `match &&` guard and added `"text"` language fallback

## Commits

- 14edf972 [01356] Fix code block rendering without language specifier

### Screenshots

edge-all-apps-final

### DIFF
--- a/src/frontend/src/components/MarkdownRenderer.tsx
+++ b/src/frontend/src/components/MarkdownRenderer.tsx
@@ -358,7 +358,9 @@ const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({
     () => ({
       code: memo((props: React.ComponentProps<"code">) => {
         const { children, className } = props;
-        const inline = !className;
+        const node = (props as any).node;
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2909, opened 2026-04-02
[01449] Tooltip should set aria-label on icon-only buttons

nielsbosma:
# Summary

Modified the `withTooltip` HOC to automatically set `aria-label` on wrapped components when a tooltip is present and no explicit `aria-label` is already provided. Also updated the explicit `TooltipWidget` to extract text content from its Content slot and apply it as `aria-label` on the trigger span wrapper.

## API Changes

- `withTooltip` HOC now passes `aria-label={tooltipText}` to wrapped components (non-breaking — only adds attribute when tooltip exists and no explicit aria-label is set)
- `TooltipWidget` span wrapper now includes `aria-label` extracted from text content in the Content slot

## Files Modified

- **src/frontend/src/hoc/withTooltip.tsx** — Added aria-label propagation from tooltip text to wrapped component
- **src/frontend/src/widgets/tooltip/TooltipWidget.tsx** — Added `extractText` helper and aria-label on trigger span

## Commits

- `95bc9738` [01449] Add aria-label to tooltip-wrapped components for accessibility

### DIFF
--- a/src/frontend/src/hoc/withTooltip.tsx
+++ b/src/frontend/src/hoc/withTooltip.tsx
@@ -12,8 +12,10 @@ function withTooltip<P extends React.JSX.IntrinsicAttributes>(Component: Compone
   return function TooltipHOC(props: WithTooltipProps<P>) {
     const { tooltipText, className, style, ...restProps } = props;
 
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2913, opened 2026-04-02
[01484] Dashboard table still not showing data rows

nielsbosma:
# Summary

Fixed the Dashboard DataTable showing only column headers with no data rows. The root cause was that `.Height(Size.Full())` relied on flex height resolution, but the DataTable's container had zero height at mount time because it was inside a nested vertical layout with a sibling chart. Also improved the Framework's `useContainerSize` hook to retry initial measurement across animation frames instead of failing silently when the container has zero height at mount.

## Files Modified

- **Ivy-Framework**
  - `src/frontend/src/widgets/dataTables/hooks/useContainerSize.ts` — Replaced synchronous initial measurement with a retry loop (up to 5 animation frames) for containers that start with zero dimensions

## Commits

- 2cff24eb [01484] Retry initial container measurement for nested flex layouts

### DIFF
--- a/src/frontend/src/widgets/dataTables/hooks/useContainerSize.ts
+++ b/src/frontend/src/widgets/dataTables/hooks/useContainerSize.ts
@@ -58,13 +58,21 @@ export const useContainerSize = () => {
 
     resizeObserver.observe(containerRef.current);
 
-    // Synchronous initial measurement to avoid first render with height=0.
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Deleted

artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2918, opened 2026-04-02
Fix Docker publish dropping rustserver native binary

rorychatt:
Fixes an issue where `dotnet publish` within Docker fails to package the rustserver native binaries during a clean checkout, because MSBuild static `ItemGroup` evaluation occurs before the `BuildRustServer` target writes the required files. Added a manual `cargo build --release` step before MSBuild invocation to guarantee the `.so` file exists. Also adds a `.dockerignore` to prevent local obj/bin folders from breaking Docker builds with `CS2001` absolute path errors.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+**/bin/
+**/obj/
+**/node_modules/
+**/.vs/
+**/.idea/
+**/.vscode/
+**/.git/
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2926, opened 2026-04-02
Fix: Package raw docs from Docs folder

rorychatt:
Removes `UseGitIgnore=false` globally to satisfy explicit rules. Fixes the underlying missing payload issue dynamically introduced during the Rust PR, where LLM-specific `.md` file generation was removed from the compilation output, making the `Generated/**/*.md` glob natively omit the packaging. By migrating the `<None>` glob explicitly to `Docs/**/*.md` we safely embed the raw unmutated `.md` files directly into the NuGet package without engaging GitIgnore logic.


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #2929, opened 2026-04-02
[01502] Fix Sheet width not applied when using Resizable

nielsbosma:
## Summary

Fixed the Sheet widget's width not being applied when using the Resizable feature (and also for non-resizable sheets). The `sheetVariant` in the shadcn `sheet.tsx` component applies `w-3/4` and `sm:max-w-sm` Tailwind classes that conflicted with the inline `style` width set by `SheetWidget.tsx`. Added `w-auto` and `sm:max-w-none` to the `className` prop so `tailwind-merge` removes the conflicting variant classes, allowing the inline styles to control width for both resizable and non-resizable paths.

## API Changes

None.

## Files Modified

- **Frontend:**
  - `src/frontend/src/widgets/sheet/SheetWidget.tsx` — Added Tailwind class overrides (`w-auto`, `sm:max-w-none`, `max-h-none`) to the `SheetContent` className to prevent variant width classes from conflicting with inline styles

## Commits

- 621a13ce [01502] Fix Sheet width not applied by overriding Tailwind variant classes

### Screenshots

01-basic-before-open

02-basic-sheet-open

03-props-custom-32rem

04-props-wide-600px

05-props-left-28rem

06-props-top-300px

07-resizable-default

08-resizable-custom-40rem

09-resizable-left

10-events-before

11-events-sheet-open

12-events-after-close

13-integration-sheet-open

14-integration-after-select

### DIFF
--- a/src/frontend/src/widgets/sheet/SheetWidget.tsx
+++ b/src/frontend/src/widgets/sheet/SheetWidget.tsx
@@ -164,7 +164,10 @@ export const SheetWidget: React.FC<SheetWidgetProps> = ({
       <SheetContent
         side={normalizedSide}
         style={styles}
-        className={cn("flex flex-col p-0 gap-0", isHorizontal && "h-full")}
+        className={cn(
+          "flex flex-col p-0 gap-0",
+          isHorizontal ? "h-full w-auto sm:max-w-none" : "max-h-none",
+        )}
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Deleted

PR #2942, opened 2026-04-02
[BadgeWidget]: render nothing for empty badges instead of tiny pill

defymecobra:
## Problem

When a `Badge` widget receives an empty string, `null`, or whitespace-only `title` with no icon, it still renders a small colored pill/bar. This is because the `BadgeWidget` component unconditionally renders the `<Badge>` wrapper `<div>`, and CSS styles (`padding`, `background-color`, `border-radius`) create a visible element even when there is no content inside.

### Observed Behavior

| Input | Result |
|---|---|
| `new Badge("")` | Colored bar |
| `new Badge(null)` | Colored bar |
| `new Badge("   ")` | Colored bar |
| `new Badge()` | Colored bar |
| Empty badge with any variant | Colored bar in variant color |

### Expected Behavior

Badges with no meaningful content (empty/null/whitespace title **and** no icon) should render nothing (`null`).

## Solution

Added an early return `null` in `BadgeWidget.tsx` when the badge has no meaningful content:

```tsx
const hasIcon = icon && icon !== "None";
if (!title?.trim() && !hasIcon) return null;
```

This check covers all edge cases:
- `null` / `undefined` title → `!title?.trim()` is `true`
- Empty string `""` → `"".trim()` is `""` (falsy)
- Whitespace-only `"   "` → `"   ".trim()` is `""` (falsy)

If an icon **is** present, the badge renders normally even without text (icon-only badges are a valid use case).

### Why frontend-only?

The backend (`Badge.cs`) correctly passes the data — it's the rendering layer that should decide whether to show an empty element. A backend fix would change the widget's API contract and potentially break users who create a Badge first and set the title later.

## Files Changed

- **`src/frontend/src/widgets/badge/BadgeWidget.tsx`** — moved `hasIcon` check to the top of the component and added early return when badge has no content

## Testing

A reproduction app (`EmptyBadgeTestApp.cs`) was created to verify:

- ✅ Empty string badges → render nothing
- ✅ Null badges → render nothing
- ✅ Whitespace-only badges → render nothing
- ✅ Empty badges in all variants (Primary, Secondary, Destructive, etc.) → render nothing
- ✅ Empty badges at all sizes (Small, Medium, Large) → render nothing
- ✅ Normal text badges → render correctly
- ✅ Icon-only badges → render correctly
- ✅ Text + icon badges → render correctly
- ✅ Whitespace title + icon badges → render correctly (icon-only)

On screenshots you can see how my test app looks after the fix applied

image
image

### DIFF
--- a/src/frontend/src/widgets/badge/BadgeWidget.tsx
+++ b/src/frontend/src/widgets/badge/BadgeWidget.tsx
@@ -30,9 +30,6 @@ export const BadgeWidget: React.FC<BadgeWidgetProps> = ({
   id,
   events = EMPTY_ARRAY,
 }) => {
-  const hasIcon = icon && icon !== "None";
-  if (!title?.trim() && !hasIcon) return null;
-
   const eventHandler = useEventHandler();
   const isClickable = events.includes("OnClick");
 
@@ -41,6 +38,10 @@ export const BadgeWidget: React.FC<BadgeWidgetProps> = ({
       eventHandler("OnClick", id, []);
     }
   }, [id, isClickable, eventHandler]);
+
+  const hasIcon = icon && icon !== "None";
+  if (!title?.trim() && !hasIcon) return null;
+
   let iconSize: number = 4;
 
   switch (density) {


defymecobra:
@claude review this

defymecobra:
Moved the early return after the hooks (useEventHandler, useCallback) to fix a React Rules of Hooks violation where hooks were being called conditionally.

defymecobra:
@claude review this

artem-ivy-ai:
### Staging deploy in progress

artem-ivy-ai:
### Deleted

PR #3034, opened 2026-04-04
[01703] Support multiple virtual columns from same root property in TableBuilder

nielsbosma:
# Summary

## Changes

Added support for multiple virtual columns derived from the same root navigation property in `TableBuilder<T>`. Previously, expressions like `e => e.Address.City` and `e => e.Address.Region` both resolved to the same column key `"Address"`, causing the second call to overwrite the first. Now each sub-property gets its own column keyed by the full property path (e.g., `"Address.City"`, `"Address.Region"`).

## API Changes

- **New:** `TypeHelper.GetFullPathFromMemberExpression(Expression)` — returns the complete dot-separated property path from a member expression (e.g., `"Address.City"` instead of `"Address"`).
- **Changed:** `TableBuilder<T>.GetField()` now uses `GetFullPathFromMemberExpression` instead of `GetNameFromMemberExpression` for column keying. Simple property expressions (e.g., `e => e.Name`) still resolve to `"Name"` (backward compatible).
- **New:** `TableBuilderColumn.ExpressionAccessor` property — compiled expression accessor used for dynamically created sub-property columns.

## Files Modified

- `src/Ivy/Helpers/TypeHelper.cs` — Added `GetFullPathFromMemberExpression` method
- `src/Ivy/Views/Tables/TableBuilder.cs` — Updated `GetField()` and `Remove()` to use full paths; added `ExpressionAccessor` to `TableBuilderColumn`
- `src/Ivy.Test/Helpers/TypeHelperTests.cs` — Added 3 tests for `GetFullPathFromMemberExpression`
- `src/Ivy.Test/TableBuilderTests.cs` — Added 3 tests for sub-property column support and backward compatibility

## Commits

- 162d8e3e [01703] Support multiple virtual columns from same root property

### DIFF
--- a/src/Ivy.Test/Helpers/TypeHelperTests.cs
+++ b/src/Ivy.Test/Helpers/TypeHelperTests.cs
@@ -210,6 +210,30 @@ public class TypeHelperTests
         Assert.True(TypeHelper.IsComplexExpression(expr.Body));
     }
 
+    [Fact]
+    public void GetFullPathFromMemberExpression_SimpleProperty_ReturnsName()
+    {
+        Expression<Func<TestModel, object>> expr = m => m.Name;
+        var name = TypeHelper.GetFullPathFromMemberExpression(expr.Body);
+        Assert.Equal("Name", name);
+    }
+
+    [Fact]
+    public void GetFullPathFromMemberExpression_NavigationProperty_ReturnsFullPath()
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Deploy timed out

### Logs

```
```

PR #3068, opened 2026-04-05
[01761] Investigate vite.config.mjs formatting issue

nielsbosma:
# Summary

## Changes

Added a `.gitattributes` file to `src/frontend/` that forces LF line endings for all frontend source files. This fixes a false positive from `vp fmt --check` on Windows where `core.autocrlf=true` converts stored LF to CRLF on checkout, but the oxfmt formatter normalizes back to LF.

## API Changes

None.

## Files Modified

- **src/frontend/.gitattributes** (new) — Forces `eol=lf` for all files in the frontend directory

---

## Commits

- 66106cf0 [01761] Add .gitattributes to force LF line endings in frontend


artem-ivy-ai:
### Deleted

PR #3157, opened 2026-04-05
[01882] Refactor YAML deserializer in Review ContentView

nielsbosma:
# Summary

## Changes

Replaced the inline `DeserializerBuilder` in `ContentView.cs` with the shared `PlanReaderService.YamlDeserializer` static field. Changed the field's visibility from `private` to `public` to enable direct access. This eliminates duplicate YAML configuration and ensures consistent deserialization settings (the shared instance includes `IgnoreUnmatchedProperties()` which the inline version lacked).

## API Changes

- `PlanReaderService.YamlDeserializer` — changed from `private static readonly` to `public static readonly`

## Files Modified

- **Services/PlanReaderService.cs** — Made `YamlDeserializer` field public
- **Apps/Review/ContentView.cs** — Replaced inline deserializer with `PlanReaderService.YamlDeserializer`

## Commits

- [01882] Refactor YAML deserializer in Review ContentView to use shared instance (5d4cc165)

### DIFF
--- a/src/tendril/Ivy.Tendril/Apps/Review/ContentView.cs
+++ b/src/tendril/Ivy.Tendril/Apps/Review/ContentView.cs
@@ -125,10 +125,7 @@ public class ContentView(
         var recommendations = new List<RecommendationYaml>();
         if (File.Exists(recommendationsPath))
         {
-            var recDeserializer = new DeserializerBuilder()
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3160, opened 2026-04-05
[01884] Optimize GetHourlyTokenBurn project filtering performance

nielsbosma:
# Summary

## Changes

Refactored `GetHourlyTokenBurn()` in `PlanReaderService` to eliminate inefficient server-side project filtering. The method now returns per-project hourly buckets with a `Project` property on `HourlyTokenBurn`, and only reads `plan.yaml` for plans that have both `costs.csv` and `logs/` (executed plans). `DashboardApp` filters client-side, reducing plan.yaml reads from all plans to only executed plans (~10x improvement).

## API Changes

- `HourlyTokenBurn.Project` — new `string` property (default `""`)
- `PlanReaderService.GetHourlyTokenBurn(int days, string? project)` — removed `project` parameter, now `GetHourlyTokenBurn(int days = 7)`

## Files Modified

- **Services/PlanReaderService.cs** — Refactored `GetHourlyTokenBurn()` to use `(Hour, Project)` composite key, read plan.yaml only for executed plans, and added `Project` property to `HourlyTokenBurn`
- **Apps/DashboardApp.cs** — Updated to call parameterless `GetHourlyTokenBurn()` and filter/aggregate client-side based on `selectedProject`

## Commits

- 65bc707e [01884] Optimize GetHourlyTokenBurn project filtering performance

### DIFF
--- a/src/tendril/Ivy.Tendril/Apps/DashboardApp.cs
+++ b/src/tendril/Ivy.Tendril/Apps/DashboardApp.cs
@@ -122,7 +122,20 @@ public class DashboardApp : ViewBase
         });
 
         // Hourly cost & tokens combined bar chart
-        var hourlyBurn = planService.GetHourlyTokenBurn(days: 7, project: selectedProject.Value);
+        var allHourlyBurn = planService.GetHourlyTokenBurn(days: 7);
+        var hourlyBurn = selectedProject.Value != null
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3162, opened 2026-04-05
[01890] Consolidate DeserializerBuilder Instances

nielsbosma:
# Summary

## Changes

Consolidated inline `DeserializerBuilder` calls into shared static readonly fields. `ConfigService` now uses a single `CamelCaseDeserializer` field (replacing two inline builders), and `ModelPricingService` uses a `DefaultDeserializer` field (replacing one inline builder).

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Services/ConfigService.cs** — Added `CamelCaseDeserializer` static field, replaced two inline `new DeserializerBuilder()` calls
- **src/tendril/Ivy.Tendril/Services/ModelPricingService.cs** — Added `DefaultDeserializer` static field, replaced one inline `new DeserializerBuilder()` call

## Commits

- 69fde098 [01890] Consolidate DeserializerBuilder instances into shared static fields

### DIFF
--- a/src/tendril/Ivy.Tendril/Services/ConfigService.cs
+++ b/src/tendril/Ivy.Tendril/Services/ConfigService.cs
@@ -108,6 +108,11 @@ public class TendrilSettings
 
 public class ConfigService
 {
+    private static readonly IDeserializer CamelCaseDeserializer = new DeserializerBuilder()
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3173, opened 2026-04-05
[01903] Recommendations Button Layout

nielsbosma:
# Summary

## Changes

Reorganized the button layout in the Recommendations ContentView. Moved the Accept button to the right edge of the header (after Spacer and count display) to make the primary action more prominent. Moved the Decline button from the header to the footer action bar as the first button, de-emphasizing the negative action.

## API Changes

None.

## Files Modified

- `src/tendril/Ivy.Tendril/Apps/Recommendations/ContentView.cs` — Reordered header and footer button composition

## Commits

- 3e8aba53 [01903] Reorganize Recommendations button layout


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3182, opened 2026-04-05
[01912] Disable Save button when no changes in General Settings

nielsbosma:
# Summary

## Changes

Added `.Disabled()` to the Save button in `GeneralSettingsView` so it is disabled when the current form values match the original configuration values. The button becomes enabled only when the user modifies either the Agent Command or Plan Template field.

## API Changes

None.

## Files Modified

- `src/tendril/Ivy.Tendril/Apps/Settings/GeneralSettingsView.cs` — Added `.Disabled()` condition comparing state values to config values

## Commits

- 8e95fa62 [01912] Disable Save button when no changes in General Settings


artem-ivy-ai:
### Staging removed

PR #3261, opened 2026-04-06
fix: removed duplicated app.UseFrontend and app.UseAssets from bad me…

rorychatt:
…rge in Server.cs


artem-ivy-ai:
### Staging removed

PR #3283, opened 2026-04-06
[01175] Close other tabs in TabsLayout overflow dropdown

rorychatt:
## Problem

When a `TabsLayout` has many tabs, hidden tabs overflow into a dropdown menu (chevron button). This dropdown only lets users select or close individual tabs. There is no way to close all tabs except the currently active one — a common UX pattern in tabbed interfaces (browsers, IDEs).

## Solution

Add an `OnCloseOthers` event to `TabsLayout` and a "Close other tabs" action in the overflow dropdown menu.

### Backend
- New `OnCloseOthers` event on `TabsLayout` (fires with the active tab index)
- Handler in `DefaultSidebarAppShell` to keep only the active tab
- Updated `TabsApp` sample to demonstrate the feature

### Frontend
- "Close other tabs" menu item in the overflow dropdown (visible when 2+ tabs exist)
- Wired through `TabsLayoutWidget` with `showCloseOthers` prop derived from events

## Commits

- `0f4cad5bb` — [01175] Add OnCloseOthers event to TabsLayout with 'Close other tabs' dropdown action
- `d923e43f7` — [01175] Fix OnCloseOthers wiring: use record with-expression instead of constructor param

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Widgets/Layouts/TabsApp.cs
+++ b/src/Ivy.Samples.Shared/Apps/Widgets/Layouts/TabsApp.cs
@@ -1,4 +1,5 @@
 using System.Collections.Immutable;
+using Ivy.Core;
 
 namespace Ivy.Samples.Shared.Apps.Widgets.Layouts;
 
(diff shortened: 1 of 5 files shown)


artem-ivy-ai:
### Staging removed

PR #3332, opened 2026-04-06
[02003] Priority levels do not need to be sorted

nielsbosma:
# Summary

## Changes

Added clarifying comments to three files explaining that priority levels are intentionally kept in config.yaml order (not sorted alphabetically), and that this order is user-controlled via the Settings UI.

## API Changes

None.

## Files Modified

- **ConfigService.cs** — Added comment above `Levels` property explaining order preservation
- **LevelsSettingsView.cs** — Added comment above levels usage explaining config.yaml ordering
- **YamlHelper.cs** — Extended XML doc on `SerializerCompact` noting collection order is preserved

## Commits

- c93d93d1b [02003] Add comments clarifying that priority levels preserve config.yaml order


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3338, opened 2026-04-06
[02013] Keyboard shortcuts trigger multiple times causing duplicate plan acceptance

nielsbosma:
# Summary

## Changes

Added debounce logic to the keyboard shortcut handler in `ButtonWidget.tsx` to prevent duplicate triggers when multiple Button components with the same shortcut key are mounted during UI transitions. A module-level `Map` tracks the last trigger time per button ID, and subsequent triggers within 300ms are ignored.

## API Changes

None.

## Files Modified

- **Frontend:**
  - `src/frontend/src/widgets/button/ButtonWidget.tsx` — Added module-level debounce state (`recentShortcuts` Map, `DEBOUNCE_MS` constant) and debounce check in the `handleKeyDown` function

## Commits

- [02013] Add debounce to keyboard shortcut handlers to prevent duplicate triggers (c2c68d0b4)

### DIFF
--- a/src/frontend/src/widgets/button/ButtonWidget.tsx
+++ b/src/frontend/src/widgets/button/ButtonWidget.tsx
@@ -21,6 +21,10 @@ import { parseShortcut, formatShortcutForDisplay, keyToCode } from "@/lib/shortc
 
 const ButtonWithTooltip = withTooltip(Button);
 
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #3340, opened 2026-04-06
[02014] Refactor FillDimensionGaps to reduce duplication

nielsbosma:
# Summary

## Changes

Refactored `FillDimensionGaps` in `PivotTable<T>` to extract the shared gap-filling logic into a generic `FillGaps<TValue>` helper method. The two nearly identical DateTime and int branches were replaced with a switch expression that dispatches to the single generic method, reducing 48 lines of duplicated code to 34 lines with no behavioral change.

## API Changes

None. The refactored methods are both `private static` — no public API surface was modified.

## Files Modified

- **src/Ivy/Views/Charts/PivotTable.cs** — Refactored `FillDimensionGaps` to use switch expression, added new private `FillGaps<TValue>` generic method

## Commits

- 291e5f93b [02014] Refactor FillDimensionGaps to extract generic FillGaps helper

### DIFF
--- a/src/Ivy/Views/Charts/PivotTable.cs
+++ b/src/Ivy/Views/Charts/PivotTable.cs
@@ -151,66 +151,52 @@ public class PivotTable<T>
         var firstValue = result[0][dimensionName];
         var lastValue = result[^1][dimensionName];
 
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #3347, opened 2026-04-06
[02022] Inline BuildStatCard int overload to reduce method count

nielsbosma:
# Summary

## Changes

Inlined the `BuildStatCard(int count, string label)` overload by adding `.ToString()` at all 6 call sites and removing the now-unnecessary int overload method, reducing the class method count by 1.

## API Changes

None. Both methods were `private static` — no public API surface affected.

## Files Modified

- `src/tendril/Ivy.Tendril/Apps/DashboardApp.cs` — Removed int overload of `BuildStatCard`, updated 6 call sites to use `.ToString()`

## Commits

- 19ce569a2 [02022] Inline BuildStatCard int overload to reduce method count


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3350, opened 2026-04-06
[02023] Replace dynamic with generic math interfaces in FillGaps

nielsbosma:
# Summary

## Changes

Replaced the single `FillGaps<TValue>` method that used `dynamic` for arithmetic with two type-safe methods: `FillDateTimeGaps` (handling `DateTime + TimeSpan`) and `FillGaps<TValue>` (constrained to `INumber<TValue>` for numeric types). Added `using System.Numerics` import.

## API Changes

None. Both methods are `private static` — no public API surface changed.

## Files Modified

- **src/Ivy/Views/Charts/PivotTable.cs** — Split `FillGaps<TValue>` into `FillDateTimeGaps` and `FillGaps<TValue>` with `INumber<TValue>` constraint; updated `FillDimensionGaps` to call `FillDateTimeGaps` for the DateTime case.

## Commits

- a4ff8f9a2 [02023] Replace dynamic with generic math interfaces in FillGaps

### DIFF
--- a/src/Ivy/Views/Charts/PivotTable.cs
+++ b/src/Ivy/Views/Charts/PivotTable.cs
@@ -1,4 +1,5 @@
 using System.Linq.Expressions;
+using System.Numerics;
 using System.Runtime.CompilerServices;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3351, opened 2026-04-06
[02025] Clean up button shortcut debounce map to prevent memory leak

nielsbosma:
# Summary

## Changes

Added cleanup logic to the ButtonWidget's keyboard shortcut `useEffect` to remove the button's entry from the module-level `recentShortcuts` Map when the component unmounts. This prevents unbounded growth of the Map during long-running sessions.

## API Changes

None.

## Files Modified

- `src/frontend/src/widgets/button/ButtonWidget.tsx` — Added `recentShortcuts.delete(id)` to the useEffect cleanup function

## Commits

- 3ca2060ad [02025] Clean up recentShortcuts map entry on button unmount


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #3761, opened 2026-04-10
[00005] Reduce containerWidth ResizeObserver re-renders in DataTable

nielsbosma:
# Summary

## Changes

Removed `containerWidth` from the column calculation pipeline in DataTable. The `convertToGridColumns` function no longer accepts a `containerWidth` parameter, and the last column always gets `grow: 1` to fill remaining space — previously this only happened when `containerWidth > 0`. This eliminates an unnecessary re-render cycle where ResizeObserver width changes triggered full column recalculation.

## API Changes

- `convertToGridColumns()` — removed 4th parameter `containerWidth: number`. Callers now pass `(columns, columnOrder, columnWidths, showGroups, ...)` instead of `(columns, columnOrder, columnWidths, containerWidth, showGroups, ...)`.
- `UseGridColumnsProps` interface — removed `containerWidth: number` property.
- `useGridColumns` hook — no longer accepts or passes `containerWidth`.

## Files Modified

- **columnHelpers.ts** — Removed `containerWidth` parameter and `containerWidth > 0` guard from `convertToGridColumns`
- **useGridColumns.ts** — Removed `containerWidth` from interface, destructuring, function call, and `useMemo` dependency array
- **DataTableEditor.tsx** — Removed `containerWidth` from `useGridColumns` call props
- **columnHelpers.test.ts** — Removed `containerWidth` arg from all 29 test calls, deleted obsolete "should not expand last column if containerWidth is 0" test, updated expected results for last column to include `grow: 1`

## Commits

- 517714d06

### DIFF
--- a/src/frontend/src/widgets/dataTables/dataTableEditor/DataTableEditor.tsx
+++ b/src/frontend/src/widgets/dataTables/dataTableEditor/DataTableEditor.tsx
@@ -202,7 +202,6 @@ export const DataTableEditor: React.FC<TableEditorProps> = ({
     columns,
     columnOrder,
     columnWidths,
-    containerWidth,
     showGroups: showGroups ?? false,
     showColumnTypeIcons: showColumnTypeIcons ?? true,
   });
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #3772, opened 2026-04-10
[02387] Worktree cleanup fails on Windows when directories are locked

nielsbosma:
# Summary

## Changes

Added a `ForceDeleteDirectory` helper method to `WorktreeCleanupService` that implements a three-stage directory deletion strategy: standard .NET deletion with read-only attribute clearing, a Windows-specific `cmd /c rmdir /s /q` fallback for locked directories, and a final retry after a 500ms delay for transient lock release. Updated the force-delete loop and worktrees directory cleanup to use the new helper. Added four new tests covering normal, read-only, nonexistent, and deeply nested directory scenarios.

## API Changes

- Added `internal static void ForceDeleteDirectory(string path)` to `WorktreeCleanupService`

## Files Modified

- **Services/WorktreeCleanupService.cs** — Added `ForceDeleteDirectory` method, added `System.Diagnostics` using, updated force-delete loop and worktrees directory cleanup to use the new helper
- **Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs** — Added 4 new test methods for `ForceDeleteDirectory`

## Commits

- `ce783f8eb` [02387] Add ForceDeleteDirectory with Windows rmdir fallback for locked worktree dirs
- `b4e2cd5bc` [02387] Fix ForceDeleteDirectory test for nonexistent path handling


artem-ivy-ai:
### Staging removed

PR #3773, opened 2026-04-10
[00023] Fix WorktreeCleanupService UnauthorizedAccessException on Windows

nielsbosma:
# Summary

## Changes

Added a Windows-specific `rmdir /s /q` fallback to the worktree cleanup path so `WorktreeCleanupService.CleanupPlanWorktrees` can delete deeply nested `node_modules` trees (like `helper-string-parser`) that previously threw `UnauthorizedAccessException` from `Directory.Delete`. The same fallback now backs `PlanReaderService.DeletePlan` for consistency.

## API Changes

- `WorktreeCleanupService.ForceDeleteDirectory(string path, ILogger? logger = null)` — new `internal static` helper. Clears read-only attributes, calls `Directory.Delete(path, true)`, and on Windows falls back to `cmd /c rmdir /s /q "<path>"` when the delete throws `UnauthorizedAccessException` or `IOException`. Throws `IOException` if the fallback also leaves the directory in place.

No public API changes.

## Files Modified

- `src/tendril/Ivy.Tendril/Services/WorktreeCleanupService.cs` — added `using System.Diagnostics`, replaced both inline force-delete blocks with calls to the new `ForceDeleteDirectory` helper, and introduced that helper.
- `src/tendril/Ivy.Tendril/Services/PlanReaderService.cs` — `DeletePlan` now delegates to `WorktreeCleanupService.ForceDeleteDirectory` instead of duplicating the `ClearReadOnlyAttributes` + `Directory.Delete` pair.
- `src/tendril/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs` — added `CleanupPlanWorktrees_ForceDeletes_DeepNodeModulesPath` (cross-platform, reproduces the deep-path scenario) and `ForceDeleteDirectory_Logs_Fallback_When_Initial_Delete_Fails` (Windows-only, exercises and asserts on the fallback log path).

## Commits

- cf4aa49c7

### DIFF
--- a/src/tendril/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs
+++ b/src/tendril/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs
@@ -240,6 +240,67 @@ public class WorktreeCleanupServiceTests : IDisposable
         Assert.Contains("TestRepo", logEntries[0]);
     }
 
+    [Fact]
+    public void CleanupPlanWorktrees_ForceDeletes_DeepNodeModulesPath()
+    {
+        // Simulates the real-world failure: cleanup must succeed over a deeply nested
+        // node_modules tree like the one that crashed on 'helper-string-parser'.
+        var dir = CreatePlan("11000-DeepNodeModules", "Completed", DateTime.UtcNow.AddHours(-2));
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #3787, opened 2026-04-10
[03005] Warn on silently dropped Gemini effort in GetAgentCommand

nielsbosma:
# Summary

## Changes

Added a `"gemini"` case to the per-agent effort flag switch in `GetAgentCommand` (Utils.ps1). The Gemini CLI has no effort flag, so instead of silently dropping the value, the function now emits a `Write-Warning` diagnostic that names the dropped effort level. Claude and Codex behavior is unchanged.

## API Changes

None. Internal PowerShell helper behavior only — no public API, config schema, or CLI surface changes.

## Files Modified

- `src/tendril/Ivy.Tendril/Promptwares/.shared/Utils.ps1` — added `"gemini"` switch case emitting `Write-Warning` in the effort-flag block of `GetAgentCommand`

## Commits

- ad5b2478c


artem-ivy-ai:
### Staging removed

PR #3788, opened 2026-04-10
[03007] Document --max-turns Risk in Claude Health Check

nielsbosma:
# Summary

## Changes

Added a seven-line inline comment above the Claude CLI health check in `SoftwareCheckStepView.cs` explaining why `--max-turns 1` is required (caps Claude at a single agentic turn so `claude -p` exits cleanly instead of blocking on stdin until the 30 s timeout fires) and warning that the flag is undocumented in `claude --help` and could be silently removed in a future release — mirroring the Gemini CLI 0.37.1+ regression fixed in plan 00030. Comment-only change; no behavioural impact.

## API Changes

None.

## Files Modified

- `src/tendril/Ivy.Tendril/Apps/Onboarding/SoftwareCheckStepView.cs` — added 7-line comment above `CheckHealth("claude", ...)` call.

## Commits

- d597e33f2


artem-ivy-ai:
### Staging removed

PR #3815, opened 2026-04-10
[01244] Limit Playwright Test Worker Concurrency

rorychatt:
# Summary

## Changes

Updated the Playwright test configuration to limit worker concurrency for local (non-CI) runs. Workers now default to 25% of CPU cores instead of unlimited, preventing CPU saturation when multiple Tendril jobs run Playwright test suites concurrently. Added support for a `PLAYWRIGHT_WORKERS` environment variable to allow per-environment overrides.

## API Changes

- New environment variable: `PLAYWRIGHT_WORKERS` — accepts a number (`'2'`) or percentage string (`'50%'`) to override the default 25% worker limit for local Playwright runs.

## Files Modified

- **src/frontend/playwright.config.ts** — Changed `workers` setting from `undefined` (half of cores) to `'25%'` for non-CI runs, with `PLAYWRIGHT_WORKERS` env var override support.

## Commits

- fcad9f515

### DIFF
--- a/src/frontend/playwright.config.ts
+++ b/src/frontend/playwright.config.ts
@@ -19,8 +19,9 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #3828, opened 2026-04-11
[01280] Fix error sheet scrolling for large content

rorychatt:
## Changes

Restructured the `ErrorDisplay` component to use a flex column layout so that the error title and message remain visible at the top of the sheet, the stack trace scrolls independently in the middle, and the Copy Details button is pinned at the bottom with a border separator. Updated `ErrorSheet` to pass flex layout through instead of wrapping everything in a single scrollable div.

## API Changes

None.

## Files Modified

- **src/frontend/src/components/ErrorDisplay.tsx** — Changed root layout from `space-y-4` to `flex flex-col h-full gap-4`, wrapped title/message in `shrink-0` div, made stack trace section `flex-1 min-h-0 overflow-auto`, pinned button with `shrink-0 pt-4 border-t`
- **src/frontend/src/components/ErrorSheet.tsx** — Changed inner wrapper from `overflow-y-auto` to `flex flex-col` so ErrorDisplay manages its own scrolling

## Commits

- `3a8629bd9` [01280] Fix error sheet scrolling for large content

### DIFF
--- a/src/frontend/src/components/ErrorDisplay.tsx
+++ b/src/frontend/src/components/ErrorDisplay.tsx
@@ -28,25 +28,27 @@ export const ErrorDisplay: React.FC<ErrorDisplayProps> = ({ title, message, stac
   };
 
   return (
-    <div className="space-y-4">
-      {title && (
-        <div>
-          <h4 className="text-sm font-medium mb-2">Type</h4>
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #3833, opened 2026-04-11
fix(frontend): resolve failing unit test in FileInputWidget

rorychatt:
This PR updates the unit tests for FileInputWidget to match the implementation. The test was previously failing because it expected a silent return for the uploadUrl guard, whereas the code now correctly uses a toast notification for better UX.


artem-ivy-ai:
### Staging removed

PR #3840, opened 2026-04-11
[01411] Add Density method to TableBuilder

rorychatt:
## Changes

Added a `.Density(Density density)` method to `TableBuilder<TModel>`, making the API consistent with `DataTableBuilder<TModel>` which already has this method. Also fully qualified existing `Density` enum references as `Ivy.Density` to avoid method/type name conflict (matching the DataTableBuilder pattern). Added four unit tests verifying the density behavior.

## API Changes

- `TableBuilder<TModel>.Density(Density density)` — new method that sets the table density directly from a `Density` enum value

## Files Modified

- **src/Ivy/Views/Tables/TableBuilder.cs** — Added `Density(Density)` method
- **src/Ivy.Test/TableBuilderTests.cs** — Added 4 density tests and `GetDensity` helper

## Commits

- `30b601704` — Add Density method to TableBuilder
- `0e65528fe` — Add unit tests for TableBuilder density

### DIFF
--- a/src/Ivy.Test/TableBuilderTests.cs
+++ b/src/Ivy.Test/TableBuilderTests.cs
@@ -24,6 +24,48 @@ public class TableBuilderTests
         public string Region { get; set; } = string.Empty;
     }
 
+    private static Density GetDensity<T>(TableBuilder<T> builder)
+    {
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #3904, opened 2026-04-11
[03089] Fix NullReferenceException Flash In ReviewApp

nielsbosma:
# Summary

## Changes

Added null guards for `_selectedPlan` in two `UseQuery` callbacks in `ContentView.cs` to prevent a `NullReferenceException` from flashing when opening or navigating between plans in the Review app. Removed the null-forgiving operator (`!`) in favor of proper null checks with early returns.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Apps/Review/ContentView.cs** — Added `if (_selectedPlan is null) return ...;` guards in `artifactContentQuery` and `commitQuery` callbacks.

## Commits

- 8963b12d3


artem-ivy-ai:
### Staging removed

PR #3912, opened 2026-04-11
[03099] Refine project filtering to prevent substring matches

nielsbosma:
# Summary

## Changes

Replaced `LIKE '%projectFilter%'` substring matching with exact `= @project` matching in `PlanDatabaseService` for both `GetDashboardData()` and `GetHourlyTokenBurn()`. This prevents incorrect project filtering when project names contain other project names as substrings. Added two tests verifying that filtering by "Framework" does not match "FrameworkExtended"/"MyFramework", and filtering by "Console" does not match "ConsoleApp".

## API Changes

None. The public method signatures are unchanged; only the SQL query behavior is tightened from substring to exact match.

## Files Modified

- `src/tendril/Ivy.Tendril/Services/PlanDatabaseService.cs` — Removed LIKE clauses and @projectPattern parameters from GetDashboardData and GetHourlyTokenBurn
- `src/tendril/Ivy.Tendril.Test/PlanDatabaseServiceTests.cs` — Added GetDashboardData_ProjectFilter_DoesNotMatchSubstrings and GetHourlyTokenBurn_ProjectFilter_DoesNotMatchSubstrings tests

## Commits

- 834aa3e1a [03099] Replace LIKE substring matching with exact match for project filtering

### DIFF
--- a/src/tendril/Ivy.Tendril.Test/PlanDatabaseServiceTests.cs
+++ b/src/tendril/Ivy.Tendril.Test/PlanDatabaseServiceTests.cs
@@ -1042,6 +1042,37 @@ public class PlanDatabaseServiceTests : IDisposable
     }
 
     [Fact]
+    public void GetDashboardData_ProjectFilter_DoesNotMatchSubstrings()
+    {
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #3954, opened 2026-04-12
[00017] Use backspace symbol in kbd/shortcut display instead of full word

rorychatt:
## Changes

Added a `keySymbols` mapping to `formatShortcutForDisplay` in `shortcut.ts` that converts special key names (Backspace, Delete, Enter, Escape, Tab, Space, arrow keys) to their standard Unicode symbols (⌫, ⌦, ↵, Esc, ⇥, ␣, ↑↓←→). On macOS, Shift and Alt/Option are also mapped to ⇧ and ⌥. Created unit tests for the new mappings.

## API Changes

None. This is a display-only change — the `formatShortcutForDisplay` function signature is unchanged.

## Files Modified

- **src/frontend/src/lib/shortcut.ts** — Added `keySymbols` lookup map and symbol resolution before the fallback case in `formatShortcutForDisplay`
- **src/frontend/src/lib/shortcut.test.ts** — New test file with unit tests for key symbol rendering

## Commits

- de9034bb1 — [00017] Add key symbol mapping for shortcut display

### DIFF
--- a/src/frontend/src/lib/shortcut.test.ts
+++ b/src/frontend/src/lib/shortcut.test.ts
@@ -1,0 +1,53 @@
+import { describe, expect, it } from "vitest";
+import { formatShortcutForDisplay } from "./shortcut";
+
+describe("formatShortcutForDisplay", () => {
+  it("renders Backspace as ⌫", () => {
+    const result = formatShortcutForDisplay("Backspace");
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #3966, opened 2026-04-12
[00044] Lazy-load heavy statically-imported widgets to reduce initial bundle size

rorychatt:
## Changes

Converted 8 statically-imported widgets in `widgetMap.ts` to lazy imports using the existing `lazyWithRetry` helper. The converted widgets (RichTextBlock, TabsLayout, Tab, DateTimeInput, SelectInput, AsyncSelectInput, DateRangeInput, List) pull in heavy third-party dependencies (katex, @dnd-kit/*, date-fns, react-day-picker, @tanstack/react-virtual, cmdk) that were previously bundled into the initial JS payload.

## API Changes

None. Only the loading mechanism changed from static to lazy; all widget APIs remain identical.

## Files Modified

- `src/frontend/src/widgets/widgetMap.ts` — Removed 8 static imports, replaced corresponding map entries with `lazyWithRetry(() => import(...))` calls using the `.then((m) => ({ default: m.WidgetName }))` pattern for named exports.

## Commits

- 161718ca4 [00044] Lazy-load heavy statically-imported widgets to reduce initial bundle size

### DIFF
--- a/src/frontend/src/widgets/widgetMap.ts
+++ b/src/frontend/src/widgets/widgetMap.ts
@@ -31,15 +31,11 @@ import {
   FieldWidget,
   TextInputWidget,
   BoolInputWidget,
-  DateTimeInputWidget,
   NumberInputWidget,
   NumberRangeInputWidget,
-  SelectInputWidget,
   ReadOnlyInputWidget,
   ColorInputWidget,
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #3983, opened 2026-04-12
[03155] Add Gemini health check to SoftwareCheckStepView

nielsbosma:
# Summary

## Changes

Added a Gemini authentication health check to `SoftwareCheckStepView` that verifies auth status by checking for `~/.gemini/oauth_creds.json` instead of running `gemini -p` (which opens a browser when unauthenticated). Updated the `anyAgentHealthy` logic to include Gemini in the proper health check evaluation rather than falling back to the installation-only check.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Apps/Onboarding/SoftwareCheckStepView.cs** — Added `CheckGeminiAuth()` method, wired Gemini health task into the pending tasks collection, and fixed `anyAgentHealthy` to use Gemini health status instead of installation-only fallback.

## Commits

- e76ca2bc8 [03155] Add Gemini health check to SoftwareCheckStepView


artem-ivy-ai:
### Staging removed

PR #3984, opened 2026-04-12
[03154] Refactor Windows KillProcessUsingPort to use GetPidsOnPort pattern

nielsbosma:
# Summary

## Changes

Refactored `KillProcessUsingPortWindows` in `ProcessHelper.cs` to use the existing `GetPidsOnPort` helper method instead of inline `Process` creation and output parsing. The Windows path now follows the same pattern as the Unix path, with a port-filtering lambda passed to `GetPidsOnPort`. Used `List<int>` instead of `yield return` since C# does not support iterator blocks in lambdas.

## API Changes

None.

## Files Modified

- **src/Ivy/Helpers/ProcessHelper.cs** — Replaced inline netstat process creation and parsing in `KillProcessUsingPortWindows` with a call to `GetPidsOnPort` using an inline parser lambda.

## Commits

- 701216ebf [03154] Refactor KillProcessUsingPortWindows to use GetPidsOnPort

### DIFF
--- a/src/Ivy/Helpers/ProcessHelper.cs
+++ b/src/Ivy/Helpers/ProcessHelper.cs
@@ -39,39 +39,37 @@ public static class ProcessHelper
 
     private static void KillProcessUsingPortWindows(int port)
     {
-        using var netstat = new Process
+        var pids = GetPidsOnPort("netstat", "-ano", output =>
         {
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #3990, opened 2026-04-13
[03160] Fix job timeout monitor race condition

nielsbosma:
# Summary

## Changes

Fixed race condition in `JobService.LaunchJob` monitor task where `ObjectDisposedException` was silently caught without completing the job, causing jobs to appear stuck indefinitely. Added specific `ObjectDisposedException` handler that exits gracefully (since the job is already being completed elsewhere), and added `CompleteJob` call in the general exception handler to prevent jobs from getting stuck.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Services/JobService.cs** — Added `ObjectDisposedException` catch before general `Exception` catch in monitor task; general catch now calls `CompleteJob` as fallback
- **src/tendril/Ivy.Tendril.Test/JobServiceTimeoutTests.cs** — Added `CompleteJob_AfterCtsDisposed_StillCompletes` test verifying job completion works after CTS disposal

## Commits

- afeedb035 [03160] Fix job timeout monitor race condition


artem-ivy-ai:
### Staging removed

PR #4005, opened 2026-04-13
[03187] SearchVariant missing disabled/invalid styles

nielsbosma:
# Summary

## Changes

Added missing `disabled` and `invalid` conditional styles to the SearchVariant's bordered container div, bringing it in line with DefaultVariant behavior. When `props.disabled` is true, the container now shows `cursor-not-allowed opacity-50`; when `props.invalid` is set, it shows `border-destructive`. Added three rendering tests verifying both states and the default (no-op) case.

## API Changes

None.

## Files Modified

- **src/frontend/src/widgets/inputs/TextInputWidget/variants/SearchVariant.tsx** — Added `props.invalid && "border-destructive"` and `props.disabled && "cursor-not-allowed opacity-50"` to the container's `cn()` call
- **src/frontend/src/widgets/inputs/TextInputWidget/__tests__/TextInputWidget.test.ts** — Added vi.mock setup for SearchVariant dependencies and three new test cases for disabled/invalid container styles

## Commits

- 24d486d26


artem-ivy-ai:
### Staging removed

PR #4006, opened 2026-04-13
[03178] Connection secret presets not auto-loaded into IConfiguration

nielsbosma:
# Summary

## Changes

Modified `Server.AddConnectionsFromAssembly()` to collect secret presets from connections implementing `IHaveSecrets` and inject them as lowest-priority in-memory configuration sources before calling `RegisterServices()`. This ensures preset values from `GetSecrets()` are available via `IConfiguration` without requiring manual user-secrets setup.

## API Changes

None — no public API surface changes. The existing `AddConnectionsFromAssembly()` method signature is unchanged; it now additionally populates `Configuration` with preset values.

## Files Modified

- **src/Ivy/Server.cs** — `AddConnectionsFromAssembly()` now collects presets from `IHaveSecrets` connections and rebuilds `Configuration` with in-memory presets as lowest-priority source
- **src/Ivy.Tests/ServerConfigurationTests.cs** — New test class with 3 tests verifying preset loading, environment variable override, and no-op when no presets exist

## Commits

- e3e1e6df3 [03178] Fix formatting in Server.cs
- bb39a78de [03178] Inject connection secret presets into IConfiguration

### DIFF
--- a/src/Ivy.Tests/ServerConfigurationTests.cs
+++ b/src/Ivy.Tests/ServerConfigurationTests.cs
@@ -1,0 +1,75 @@
+using Microsoft.Extensions.Configuration;
+
+namespace Ivy.Tests;
+
+public class TestConnectionWithPresets : IConnection, IHaveSecrets
+{
+    public string GetContext(string connectionPath) => "";
+    public string GetNamespace() => "Test";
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #4009, opened 2026-04-13
[00127] StackedProgress does not respect max width — overflows container bounds

rorychatt:
## Problem

Stacked progress bar does not respect its max width. In Tendril it extends out of its bounds.

## Solution

Fix StackedProgress max-width overflow by reordering CSS properties to ensure container bounds are respected.

## Commits

- `9848db39f` [00127] Fix StackedProgress max-width overflow by reordering CSS properties


artem-ivy-ai:
### Staging removed

PR #4028, opened 2026-04-13
[03208] Fix race condition in Create New Plan dialog

nielsbosma:
# Summary

## Changes

Added a loading state (`isCreating`) to the Create New Plan dialog that disables the "Create" button after the first click. This prevents duplicate plan creation from rapid clicks or repeated Ctrl+Enter key presses during the window between `onCreatePlan()` and `onClose()`.

## API Changes

None.

## Files Modified

- **CreatePlanDialog.cs** — Added `isCreating` state variable, `.Disabled(isCreating.Value)` on the Create button, guard check in OnClick handler, and `isCreating.Set(true)` at the start of the handler.

## Commits

- cbb6af945 [03208] Fix race condition in Create New Plan dialog


artem-ivy-ai:
### Staging removed

PR #4033, opened 2026-04-13
fix: correct NuGet login user to match Trusted Publisher policy

rorychatt:
Reverts the NuGet login user from Ivy-Interactive to rorychatt-ivy to match the account that holds the Trusted Publisher policy on NuGet.org. This resolves the 401 Unauthorized error in the publish-tendril workflow.


artem-ivy-ai:
### Staging removed

PR #4042, opened 2026-04-14
fix: horizontal overflow in StackedProgress widget

rorychatt:
Removes forced minWidth on segments and improves flex layout to prevent widget from expanding beyond horizontal bounds.


artem-ivy-ai:
### Staging removed

PR #4043, opened 2026-04-14
[00175] Remove push trigger from merge-main-into-development workflow

rorychatt:
## Problem

The GitHub Actions workflow `merge-main-into-development.yml` currently triggers on push to `development`, which can create a recursive trigger loop when the workflow itself merges `main` into `development` and pushes. Routine commits to `development` should not trigger a merge from `main`.

## Solution

Remove the `push` trigger from the workflow's `on:` block, keeping only `schedule` and `workflow_dispatch`.

## Commits

- 99141906e


artem-ivy-ai:
### Staging removed

PR #4062, opened 2026-04-14
[03244] Fix NullReferenceException in Review ContentView when planData is null

nielsbosma:
# Summary

## Changes

Added a null guard for `planData` in the Review app's `ContentView.Build()` method. When `planContentQuery.Value` is null after loading completes (e.g. due to a query failure), the UI now displays an error message instead of crashing with a `NullReferenceException`. The error message includes the actual exception message when available.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Apps/Review/ContentView.cs** — Added `else if (planData is null)` branch between the loading check and the main content rendering block

## Commits

- e34c472d8 [03244] Add null guard for planData in Review ContentView


artem-ivy-ai:
### Staging removed

PR #4069, opened 2026-04-14
[01463] Add visual distinction for selected vs hover overlay in DevTools

dcrjodle:
## Problem

After Plan 01462 added overlay persistence for the selected widget, both hover and selected states use the identical visual style: a `2px solid var(--primary)` border with `25% primary` background fill. Users cannot visually distinguish between "I'm hovering over this widget" and "I have this widget selected with the dialog open."

## Solution

Add a CSS modifier class `ivy-devtools-overlay--selected` when the overlay represents the selected widget (`dialogWidget` is set). The selected state uses:
- A slightly **lower** background opacity (15% vs 25%) — making the content more visible while selected
- An additional subtle outer `box-shadow` glow to create a "locked-in" feel
- The hover state retains its current stronger fill (25%) for immediate feedback

## Commits

- 9d1f43d61

### DIFF
--- a/src/frontend/src/components/DevTools.tsx
+++ b/src/frontend/src/components/DevTools.tsx
@@ -283,8 +283,9 @@ export function DevTools() {
     const { bounds, type } = activeWidget;
     if (bounds.width === 0 && bounds.height === 0) return;
 
+    const isSelected = !!dialogWidget;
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #4087, opened 2026-04-14
[03213] Fix markdown links rendering as literal text instead of clickable links

nielsbosma:
# Summary

## Changes

Added an `allowCustomProtocols` option to `validateLinkUrl()` that switches from a strict protocol allowlist (http/https only) to a blocklist of dangerous protocols (javascript:, data:, vbscript:, blob:). `MarkdownRenderer` now passes this option when an `onLinkClick` handler is registered, allowing custom URL schemes like `plan://` to render as clickable links instead of plain `<span>` text.

## API Changes

- `validateLinkUrl(url, options?)` — new optional second parameter `{ allowCustomProtocols?: boolean }`. Default behavior is unchanged.

## Files Modified

- `src/frontend/src/lib/url.ts` — Added `DANGEROUS_PROTOCOLS` set and `allowCustomProtocols` option to `validateLinkUrl`
- `src/frontend/src/components/MarkdownRenderer.tsx` — Pass `allowCustomProtocols: !!onLinkClick` in both the link component and `handleLinkClick` callback
- `src/frontend/src/lib/utils.test.ts` — Added test suite for `allowCustomProtocols` behavior

## Commits

- dd3653e74 [03213] Allow custom URL protocols in MarkdownRenderer when onLinkClick is registered

### DIFF
--- a/src/frontend/src/components/MarkdownRenderer.tsx
+++ b/src/frontend/src/components/MarkdownRenderer.tsx
@@ -159,7 +159,11 @@ const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({
 
       // Validate URL to prevent open redirect vulnerabilities
       // validateLinkUrl always returns a string ('#' for invalid URLs)
-      const validatedHref = validateLinkUrl(href);
+      // When onLinkClick is registered, allow custom protocols (e.g. plan://)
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #4091, opened 2026-04-14
[03270] Extract shared WithSlot helper to reduce duplication

nielsbosma:
# Summary

## Changes

Extracted the duplicated `WithSlot` helper method from 9 widget extension classes into a single generic `internal static object[] WithSlot<T>(this T widget, string slotName, object? value) where T : WidgetBase` method on `WidgetBaseExtensions`. Card's slightly different implementation (using `OfType<Slot>()`) was unified to use the same logic since Card only stores Slot children — the broader filter is strictly more correct.

## API Changes

- Added `WidgetBaseExtensions.WithSlot<T>(this T widget, string slotName, object? value)` — `internal` visibility, not a public API change.
- Removed 9 `private static WithSlot(...)` methods from individual extension classes.

## Files Modified

- **Shared helper:** `src/Ivy/Widgets/WidgetBase.cs` — added generic `WithSlot<T>` extension method
- **Input widgets (8 files):** TextInput.cs, NumberInput.cs, NumberRangeInput.cs, BoolInput.cs, SelectInput.cs, DateTimeInput.cs, DateRangeInput.cs, ColorInput.cs — removed private WithSlot, updated Prefix/Suffix to use shared helper
- **Card:** `src/Ivy/Widgets/Card.cs` — removed private WithSlot, updated Content/Footer to use shared helper

## Commits

- 3d7d69a8c

### DIFF
--- a/src/Ivy/Widgets/Card.cs
+++ b/src/Ivy/Widgets/Card.cs
@@ -53,13 +53,6 @@ public static class CardExtensions
 {
     internal static Slot? GetSlot(this Card card, string name) => card.Children.FirstOrDefault(e => e is Slot slot && slot.Name == name) as Slot;
 
-    private static object[] WithSlot(Card card, string slotName, object? value)
-    {
-        var others = card.Children.OfType<Slot>().Where(s => s.Name != slotName);
-        var result = value != null ? others.Append(new Slot(slotName, value)) : others;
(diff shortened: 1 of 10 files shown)


artem-ivy-ai:
### Staging removed

PR #4120, opened 2026-04-14
[03300] Restore periodic table refresh for Jobs with recently-completed jobs

nielsbosma:
# Summary

## Changes

Added a `UseInterval` block to JobsApp.cs that triggers a full table refresh every 5 seconds when there are Running or recently-completed jobs. This restores the periodic refresh behavior that was removed in plan 03211, ensuring the table stays fresh while jobs are active or within 5 seconds of completion.

## API Changes

None.

## Files Modified

- `src/tendril/Ivy.Tendril/Apps/JobsApp.cs` — Added UseInterval block (lines 51-62) that checks for active/recent jobs and refreshes the table accordingly.

## Commits

- d03a36d13 [03300] Restore periodic table refresh for Jobs with recently-completed jobs


artem-ivy-ai:
### Staging removed

PR #4126, opened 2026-04-14
fix: tendrildocs deployment Final

rorychatt:
Standardizes Dockerfile structure by using /src as root and adding the missing Ivy.Agent.Filter dependency.


artem-ivy-ai:
### Staging removed

PR #4155, opened 2026-04-14
security: upgrade follow-redirects to >=1.16.0

rorychatt:
Upgrades follow-redirects to fix a Dependabot alert. Verified with frontend and backend builds.


artem-ivy-ai:
### Staging removed

PR #4170, opened 2026-04-14
[03349] Fix Assignee Dropdown Showing Labels in Import Issues Dialog

nielsbosma:
# Summary

## Changes

Fixed a bug in the Import Issues dialog where the Assignee dropdown was displaying GitHub labels instead of users. The root cause was that both `assigneesQuery` and `labelsQuery` used the same cache key (`selectedRepo.Value`), causing UseQuery to conflate the two queries and mix up their cached data. Added unique prefixes to the query keys to ensure proper separation.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/AppShell/Dialogs/ImportIssuesDialog.cs**
  - Modified `assigneesQuery` to use key `$"assignees:{selectedRepo.Value ?? ""}"`
  - Modified `labelsQuery` to use key `$"labels:{selectedRepo.Value ?? ""}"`
  - Updated fetcher functions to strip the prefix before processing

## Commits

- 6ca50236d [03349] Fix assignee dropdown showing labels by using unique query keys


artem-ivy-ai:
### Staging removed

PR #4176, opened 2026-04-15
[01105] Add docs for auth cookie prefix support

zachwolfe:
## Problem

Plan [01100] added `Server.AuthCookiePrefix` property to support multiple Ivy auth apps on the same domain without cookie collisions. However, this feature is not yet documented, so users won't know when to use it or how to configure it.

## Solution

Added documentation for `Server.AuthCookiePrefix` to the existing "Customizing Authentication Cookies" section in AuthenticationOverview.md. The new "Namespacing Cookies for Multiple Apps" subsection includes:

- Use case explanation for multiple apps on the same domain
- Configuration example
- Cookie naming scheme with double-underscore delimiter
- Table showing cookie names with and without prefixes
- Note about brokered OAuth session cookies

## Commits

- 10ee69c9c [01105] Add documentation for AuthCookiePrefix to customize authentication cookies


artem-ivy-ai:
### Staging removed

PR #4179, opened 2026-04-15
[01110] Fix auth cookie prefix filtering in brokered session extraction

zachwolfe:
## Problem

After [Plan 01104](https://github.com/Ivy-Interactive/Ivy-Framework/pull?q=01104) refactored `AuthHelper.cs` to use the `PrefixCookieName()` method, the brokered session extraction methods lost their prefix-based cookie filtering. The original code filtered cookies with `key.StartsWith(pfx)`, but the refactored version only checks `key.EndsWith(suffix)`.

This causes cross-contamination when running multiple auth example apps in the same browser. For example:
1. Log into BasicAuthExample (prefix: `basic`) — sets cookie `basic__access_token`
2. Switch to ClerkExample (prefix: `clerk`) — its `ExtractBrokeredSessionsFromCookies()` finds `basic__access_token` because it ends with `_access_token`
3. The prefix stripping logic uses the *current* app's prefix length (8 chars for "clerk__"), but the cookie has a *different* prefix ("basic__")
4. For `basic__access_token`, the prefix doesn't match so it isn't stripped, yielding provider name `"basic_"` (the full cookie minus `_access_token`)
5. No handler exists for provider `"basic_"`, causing repeated warnings and errors

**Affected code:** Both `ExtractBrokeredSessionsFromCookies()` (line 202) and `ExtractBrokeredSessionsFromCookieHeader()` (line 244) in `src/Ivy/Core/Auth/AuthHelper.cs`.

## Solution

Re-add prefix filtering to both extraction methods. When a prefix is configured, only consider cookies that start with the current app's prefix. Cookies from other prefixes should be silently ignored.

**Changes:**

1. **`ExtractBrokeredSessionsFromCookies()` (line 206-211):** Add prefix filtering to the LINQ query. Compute the prefix string from the primary access token. Add a `key.StartsWith(prefix)` check when a prefix is configured.

2. **`ExtractBrokeredSessionsFromCookieHeader()` (line 257-263):** Apply the same prefix filtering.

This restores the prefix-scoping behavior that existed before the [01104] refactor, while keeping the `PrefixCookieName()` usage.

## Commits

- b94afce0a [01110] Add Collection attribute to prevent auth cookie test parallelism
- caa87a7a4 [01110] Fix auth cookie prefix filtering in brokered session extraction

### DIFF
--- a/src/Ivy.Tests/Auth/AuthCookiePrefixTests.cs
+++ b/src/Ivy.Tests/Auth/AuthCookiePrefixTests.cs
@@ -2,6 +2,7 @@ using Ivy.Core.Auth;
 
 namespace Ivy.Tests.Auth;
 
+[Collection("AuthCookiePrefix")]
 public class AuthCookiePrefixTests : IDisposable
 {
     public AuthCookiePrefixTests()
--- a/src/Ivy.Tests/Auth/AuthHelperBrokeredSessionTests.cs
+++ b/src/Ivy.Tests/Auth/AuthHelperBrokeredSessionTests.cs
@@ -1,4 +1,3 @@
-using System.Collections;
 using System.Diagnostics.CodeAnalysis;
 using Ivy.Core.Auth;
 using Microsoft.AspNetCore.Http;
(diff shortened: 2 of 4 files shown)


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-staging-docs-4179.sliplane.app](https://ivy-staging-docs-4179.sliplane.app)
**🧪 Samples:** [https://ivy-staging-samples-4179.sliplane.app](https://ivy-staging-samples-4179.sliplane.app)

artem-ivy-ai:
### Staging removed

PR #4185, opened 2026-04-15
[01510] Extract density gap class to shared utility

dcrjodle:
### Changes

Extracted the duplicated density-to-gap-class ternary from `TreeWidget.tsx` and `TreeItem.tsx` into a shared `densityTreeGap` constant in `density-scale.ts`. Both tree files now use a simple map lookup instead of inline ternaries, eliminating drift risk.

### API Changes

- **New export:** `densityTreeGap` from `src/frontend/src/components/ui/density-scale.ts` — maps `Densities` enum keys to Tailwind gap classes (`gap-0.5`, `gap-1`, `gap-1.5`).

### Files Modified

- `src/frontend/src/components/ui/density-scale.ts` — added `densityTreeGap` constant
- `src/frontend/src/widgets/tree/TreeWidget.tsx` — replaced inline ternary with `densityTreeGap[density]`
- `src/frontend/src/widgets/tree/TreeItem.tsx` — replaced inline ternary with `densityTreeGap[density ?? Densities.Medium]`

## Commits

- b75112ca3 — [01510] Extract density gap class to shared densityTreeGap utility

### DIFF
--- a/src/frontend/src/components/ui/density-scale.ts
+++ b/src/frontend/src/components/ui/density-scale.ts
@@ -17,3 +17,9 @@ export const densityText = {
   Medium: "text-sm",
   Large: "text-base",
 } as const;
+
+export const densityTreeGap = {
+  Small: "gap-0.5",
+  Medium: "gap-1",
+  Large: "gap-1.5",
+} as const;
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #4187, opened 2026-04-15
[01516] Tree node icon-to-chevron swap on hover

dcrjodle:
## Changes

Modified the Tree widget's `TreeItem` component to swap the node icon with a chevron (expand/collapse indicator) on row hover, for nodes that have both children and an icon. This matches the interaction pattern used in Ivy Studio's collapsible message groups. Nodes without icons retain the always-visible chevron behavior.

## API Changes

None.

## Files Modified

- **src/frontend/src/widgets/tree/TreeItem.tsx** — Added `hovered` state, `onMouseEnter`/`onMouseLeave` handlers on the tree item row, and conditional rendering that shows the icon by default but swaps to the chevron on hover (for nodes with both children and an icon).

## Commits

- `db7ed5b90` — [01516] Add icon-to-chevron swap on hover for tree nodes with children


artem-ivy-ai:
### Staging removed

PR #4223, opened 2026-04-21
Fix KBD styling in buttons to inherit parent colors

dcrjodle:
## Summary
- Removed `text-muted-foreground` and `bg-muted` from KBD elements inside buttons so they inherit the button's text color and have no independent background
- Added `border-current/20` to use the inherited text color at 20% opacity for the border
- Added `opacity-70` to subtly differentiate the shortcut badge from the button label

## Test plan
- [ ] Verify KBD badges in primary, secondary, destructive, and ghost button variants all inherit the correct text color
- [ ] Confirm the KBD border is visible but subtle across all variants
- [ ] Check dark mode rendering


artem-ivy-ai:
### Staging removed

PR #3135, opened 2026-04-05
[01857] Remove Make Draft button from Review Recommendations

nielsbosma:
# Summary

## Changes

Removed the "+ Make Draft" button from the Review app's Recommendations tab. This button was redundant since the dedicated Recommendations app provides proper Accept/Decline workflow with state management via `UpdateRecommendationState()`.

## API Changes

None.

## Files Modified

- **src/tendril/Ivy.Tendril/Apps/Review/ContentView.cs** — Removed the Button and its `_jobService.StartJob("MakePlan", ...)` handler from the recommendations loop (lines 324-329). Also removed the now-unused `recCapture` variable.

## Commits

- 6ec71d48 [01857] Remove Make Draft button from Review Recommendations tab

PR #4228, opened 2026-04-21
Fix missing tooltip on Bar/Line/Area chart hover

dcrjodle:
## Summary
- Bar/Line/Area chart tooltips silently failed to render (observed on the cost/token chart in Ivy Tendril's Dashboard app).
- The tooltip `formatter` indexed `params.value` with `[isVertical ? 0 : 1]`, but those widgets emit **scalar** series data via `data.map((d) => d[key])`. Indexing a number yielded `undefined`, and `formatTooltipValue` then called `undefined.toLocaleString()` — a thrown error which ECharts swallows, dropping the tooltip entirely.
- Regression was introduced by #03135 / commit `ec969b993` ("Add tooltip value formatting to cartesian charts"), which ported the scatter-chart indexing to cartesian widgets whose data isn't a `[x, y]` tuple.

## Fix
- Extract the value with `Array.isArray(p.value) ? p.value[isVertical ? 1 : 0] : p.value` in all three cartesian widgets (Bar, Line, Area).
- Also corrects the inverted index for the tuple case: the value axis is opposite the category axis (vertical → index 1, horizontal → index 0).
- ScatterChartWidget is untouched — its data really is `[x, y]` tuples.

## Test plan
- [x] `npm run build` in `src/frontend/` succeeds
- [ ] Manual: hover the hourly cost/token bar chart on the Ivy Tendril dashboard and confirm the tooltip renders with category header + both series values
- [ ] Manual: hover a single-series bar chart and confirm the tooltip renders
- [ ] Manual: hover a line chart and an area chart and confirm tooltips render

### DIFF
--- a/src/frontend/src/widgets/charts/AreaChartWidget.tsx
+++ b/src/frontend/src/widgets/charts/AreaChartWidget.tsx
@@ -195,12 +195,14 @@ const AreaChartWidget: React.FC<AreaChartWidgetProps> = ({
           mutedForeground: themeColors.mutedForeground,
         }),
         formatter: (params: any) => {
+          const extractValue = (p: any) =>
+            Array.isArray(p.value) ? p.value[isVertical ? 1 : 0] : p.value;
           if (Array.isArray(params)) {
             // Multi-series: add category header from first param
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #4233, opened 2026-04-22
(security): fix alerts

ArtemKhvorostianyi:
### What CodeQL complained about

Screenshot 2026-04-22 at 04 05 29

TriggerMachineAuthSync came from SetAuthCookiesRequest (JSON body). Anyone who could call PATCH /ivy/auth/set-auth-cookies with a valid cookieJarId could set that flag to true and push SyncAuthFromCookies to other sessions that share the X-Machine-Id header. The gate for that sensitive path was effectively client-controlled.

Redeeming a jar still requires a fresh server-issued id, but the rule is about not letting request fields decide whether privileged follow-up runs.

CookieJarEntry / AppSessionStore — When registering a cookie jar, the server now stores TriggerMachineAuthSync on the entry. TryRemoveCookies returns it via an out parameter (default false if redeem fails).

RegisterAuthSessionCookies — Takes triggerMachineAuthSync (default false) and passes it into RegisterCookies.

AuthService / ConnectedAccountsService — They already choose when sync is needed; that flag is passed only at RegisterAuthSessionCookies time. IClientProvider.SetAuthCookies no longer sends triggerMachineAuthSync over SignalR (it was redundant for security once the server owns the flag).

AuthController.SetAuthCookies — Drops TriggerMachineAuthSync from the request DTO and uses the value returned from WriteCookiesToResponse (i.e. from the redeemed jar).

Frontend use-backend.tsx — Stops sending triggerMachineAuthSync in the PATCH body.

Screenshot 2026-04-22 at 04 06 18

Bump the override from 3.3.2 to 3.4.0 (meets “≥ 3.4.0”).

P.S.

Screenshot 2026-04-22 at 04 07 49

These two alerts send [Ivy.Database.Generator.Toolkit/Ivy.Database.Generator.Toolkit.csproj](https://github.com/Ivy-Interactive/Ivy-Framework/blob/-/Ivy.Database.Generator.Toolkit/Ivy.Database.Generator.Toolkit.csproj) and it seems that I don't have access to it, so these two alerts weren't fixed

Also about auth changes I think it will be good to set Zack as a reviewer.

### DIFF
--- a/src/Ivy/Auth/AuthService.cs
+++ b/src/Ivy/Auth/AuthService.cs
@@ -171,8 +171,8 @@ public class AuthService : AuthTokenHandlerService, IAuthService
         if (hasChanges)
         {
             // Pass removed providers so their cookies get deleted
-            var cookieJarId = _sessionStore.RegisterAuthSessionCookies(_authSession, removedProviders);
-            _client.SetAuthCookies(cookieJarId, reloadPage: false, triggerMachineReload: null, triggerMachineAuthSync: true);
+            var cookieJarId = _sessionStore.RegisterAuthSessionCookies(_authSession, removedProviders, triggerMachineAuthSync: true);
+            _client.SetAuthCookies(cookieJarId, reloadPage: false, triggerMachineReload: null);
         }
 
         return BrokeredSessionsResult.Success(filteredSessions);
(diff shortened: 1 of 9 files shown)


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-staging-docs-4233.sliplane.app](https://ivy-staging-docs-4233.sliplane.app)
**🧪 Samples:** [https://ivy-staging-samples-4233.sliplane.app](https://ivy-staging-samples-4233.sliplane.app)

zachwolfe (approved):
Getting `triggerMachineAuthSync` out of the frontend is a good idea. I tested a few of the auth examples (including the new GitHub connected account support in BasicAuthExample) and everything seems to work.

artem-ivy-ai:
### Staging removed

PR #4258, opened 2026-04-23
Fix SidebarNews to use feedUrl prop instead of hardcoded broken URL

nielsbosma:
## Summary
- SidebarNewsWidget ignored the `feedUrl` prop and fetched from hardcoded `https://ivy.app/news/news.json` which returns 404
- Now uses the `feedUrl` prop passed from C# (`https://cdn.ivy.app/tendril/news.json`)
- Image base URL derived from feed URL instead of hardcoded broken path


artem-ivy-ai:
### Staging removed

PR #4260, opened 2026-04-23
Add ImageBaseUrl prop to SidebarNews for CORS proxy support

nielsbosma:
## Summary
- Add optional `ImageBaseUrl` prop to `SidebarNews` widget (C# and TSX)
- When provided, uses it for image URLs instead of deriving from feedUrl
- Needed when feedUrl is a local proxy endpoint and images live on a CDN


artem-ivy-ai:
### Staging removed

PR #4264, opened 2026-04-24
[00057] NumberInput focus border should wrap entire widget including affixes

dcrjodle:
# Summary

Applied the same focus-border pattern from TextInput's DefaultVariant to NumberInput's NumberVariant. When a NumberInput with prefix/suffix affixes is focused, the border now wraps the entire widget container (input + affixes) instead of just the inner input element. Affix separator borders are conditionally hidden during focus for a clean unified appearance.

## API Changes

None.

## Files Modified

- **src/frontend/src/widgets/inputs/NumberInputWidget.tsx** — Added `useState` import, `isFocused` state with `handleFocus`/`handleBlur` callbacks in `NumberVariant`, updated outer container div to use focus-aware border classes (`border-ring` when focused, `border-input` when not), and made prefix/suffix separator borders conditional on `!isFocused`.

## Commits

- `79e76f8d7` [00057] Add focus border wrapping for NumberInput with affixes


artem-ivy-ai:
### Staging removed

PR #4266, opened 2026-04-24
[00062] Enhance FolderDialog and UseFolderDialog to return selected folder path

dcrjodle:
## Problem

The existing `FolderDialogWidget.tsx` uses the browser `showDirectoryPicker()` API which only returns folder entries (files/subdirectories), not the folder's absolute path. The `UseFolderDialog` hook has no way to return the selected folder's absolute path to C# consumers. However, `browserSupport.ts` already has a `pickDirectoryFullPath()` function that uses the desktop bridge to return `{ name, path }`.

## Solution

- Add `OnPathSelected` event to `FolderDialog.cs` that returns the folder path as a string
- Update `FolderDialogWidget.tsx` to use `pickDirectoryFullPath()` from `browserSupport.ts` instead of calling `showDirectoryPicker()` directly, firing `OnPathSelected` with the absolute path
- Expose path in `UseFolderDialog` hook via a new `selectedPath` state, wired through the `OnPathSelected` event

## Commits

- `8dbd56af4` [00062] Add OnPathSelected event to FolderDialog and expose path in UseFolderDialog hook

## Verification

- [x] DotnetBuild
- [x] DotnetFormat
- [x] FrameworkFrontendLint
- [x] VitePlusCheck
- [x] CheckResult

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Widgets/FileDialogApp.cs
+++ b/src/Ivy.Samples.Shared/Apps/Widgets/FileDialogApp.cs
@@ -166,7 +166,7 @@ public class FolderPickerDemo : ViewBase
     {
         var entries = UseState<FolderDialogEntry[]?>();
 
-        var (folderDialogView, showFolderDialog) = UseFolderDialog();
+        var (folderDialogView, showFolderDialog, selectedPath) = UseFolderDialog();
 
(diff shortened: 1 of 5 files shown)


artem-ivy-ai:
### Staging removed

PR #4269, opened 2026-04-24
Fix Antlr4BuildTasks download on CI

rorychatt:
Sets the download directory to IntermediateOutputPath so it does not fail on Linux CI.


artem-ivy-ai:
### Staging removed

PR #4296, opened 2026-04-26
[03562] DataTable Cell Deselection on Outside Click

nielsbosma:
# Summary

## Changes

Added click-outside detection to the DataTableEditor component to automatically clear cell and column selections when users click anywhere outside the grid container. Fixed an issue where the previous implementation's optimization check prevented column selections from being properly cleared by removing the conditional check and always clearing selections on outside clicks.

## API Changes

None.

## Files Modified

- **src/frontend/src/widgets/dataTables/dataTableEditor/DataTableEditor.tsx**
  - Removed optimization check `if (gridSelection.rows.length > 0 || gridSelection.columns.length > 0)`
  - Now always clears selection when clicking outside (React optimizes away no-op state updates)
  - Removed `gridSelection` from useEffect dependency array to prevent re-registering event listener on every selection change
  - Updated comment to reflect the simplified approach

## Commits

- 998427694 [03562] Add click-outside detection to clear cell selections in DataTable
- c11af799d [03562] Fix TypeScript error: use length property instead of isEmpty() method
- 6c96b5a6d [03562] Remove optimization check to ensure column selections are cleared on outside click


artem-ivy-ai:
### Staging removed

PR #4302, opened 2026-04-27
Emit OnPathSelected from FolderDialog in browser mode

dcrjodle:
## Summary
- `UseFolderDialog().selectedPath` previously stayed `null` in any plain browser because the frontend only emitted `OnPathSelected` when the Ivy desktop bridge was present.
- The widget now also emits the folder name in browser mode: `dirHandle.name` from `showDirectoryPicker`, and the first segment of `webkitRelativePath` from the `webkitdirectory` fallback.
- Browsers do not expose absolute filesystem paths through any web API, so the folder name is the most a non-desktop host can supply. Desktop bridge behavior (full absolute path) is unchanged.
- XML doc on `UseFolderDialog` updated to describe both modes.

## Test plan
- [ ] Pick a folder in a Chromium browser → `selectedPath` reflects the folder name.
- [ ] Pick a folder in a non-Chromium browser (webkitdirectory fallback) → `selectedPath` reflects the root folder name.
- [ ] Run inside the Ivy desktop host → `selectedPath` still reflects the absolute path.
- [ ] Cancel the picker in each mode → no `OnPathSelected` is emitted.

### DIFF
--- a/src/Ivy/Hooks/UseFolderDialog.cs
+++ b/src/Ivy/Hooks/UseFolderDialog.cs
@@ -16,7 +16,8 @@ public static class UseFolderDialogExtensions
     /// On other browsers, falls back to input[webkitdirectory].
     /// Returns (dialogView, showFolderDialog, selectedPath) tuple following the UseAlert pattern.
     /// Call showFolderDialog(callback) to open the dialog; the callback receives folder entries.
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #4305, opened 2026-04-27
[Auth]: Enhance password handling and validation in forms

### LINKED ISSUE #4278: Basic auth password validation rejects 5-character passwords
## Summary
When setting up basic authentication for Ivy Tendril, the system enforces a minimum password length of 8 characters, but the user attempted to use a 5-character password ("bosma") with username "niels".

**Expected:** User should be able to set a password with fewer than 8 characters, or the minimum length requirement should be clearly documented.

**Actual:** The basic auth validation rejects passwords shorter than 8 characters.

## Additional Context
See attached screenshot showing the Ivy Tendril login screen with username "niels" and a 5-character password field marked with an error indicator.

---

[Slack thread](https://ivy-interactive.slack.com/archives/C09PMA599RC/p1777147775028169?thread_ts=1777147775.028169&cid=C09PMA599RC)

ArtemLazarchuk:
## Summary

Password fields in scaffolded forms now respect `[MinLength]`, `[StringLength]`, and `[Length]` instead of always enforcing a hardcoded minimum. The default remains **8 characters** when nothing is specified. Login (`DefaultAuthApp`) no longer blocks short passwords that already match the auth provider (e.g. Basic Auth hashes from the CLI).

## Motivation

Users could create short passwords via tooling (e.g. `ivy auth add`), but the email/password login form always required at least 8 characters before the request reached the provider. That mismatch blocked valid credentials.

## Changes

- **Login form** — `LoginFormModel.Password` uses `[MinLength(0)]` and `ToPasswordInput().MinLength(0)` so only **required** is enforced; length policy stays with the auth backend.
- **FormHelpers** — optional `skipAttribute` filter on `GetValidators`; new `GetMinLength` helpers for property/field (reads `MinLength`, `StringLength`, `Length`).
- **FormScaffolder** — password widgets get `.MinLength(...)` from model metadata when present; skips duplicate `[MinLength]` DataAnnotations for password fields (single password validator + message); `CreatePasswordValidator` receives `GetMinLength() ?? DefaultPasswordMinLength`.
- **Validators** — `DefaultPasswordMinLength = 8`; `minLength <= 0` skips length check; `GetEffectiveValidators` uses `TextInputBase.MinLength` for password variant when set; `ForVariant` accepts optional password min override.
- **TextInput** — mirrors `MinLength` into attached metadata so on-blur validation sees values set after `ToPasswordInput()` (record `with` / chaining).

https://github.com/user-attachments/assets/d16b1dd6-0906-491e-bc45-434e35ae2a50

### DIFF
--- a/src/Ivy/Auth/DefaultAuthApp.cs
+++ b/src/Ivy/Auth/DefaultAuthApp.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel.DataAnnotations;
 using System.Reflection;
 using Ivy.Core.Auth;
 
@@ -58,7 +59,7 @@ public class DefaultAuthApp : ViewBase
 
 public class PasswordEmailFlowView(IState<string?> errorMessage) : ViewBase
 {
-    private record LoginFormModel(string User, string Password);
+    private record LoginFormModel(string User, [property: MinLength(0)] string Password);
 
     public override object Build()
     {
@@ -104,7 +105,7 @@ public class PasswordEmailFlowView(IState<string?> errorMessage) : ViewBase
             .Label(m => m.User, "User")
             .Label(m => m.Password, "Password")
             .Builder(m => m.User, state => state.ToTextInput())
-            .Builder(m => m.Password, state => state.ToPasswordInput())
+            .Builder(m => m.Password, state => state.ToPasswordInput().MinLength(0))
             .SubmitStrategy(FormSubmitStrategy.OnSubmit)
             .SubmitTitle("Login")
             .OnSubmit(async model => await HandleLoginAsync(model));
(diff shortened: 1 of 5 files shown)


rorychatt (changes requested):
I am thinking about this solution. You implemented stuff for specific case of MinLength. But also usually there are other requirements there, like max length. Or having a special character. My guess is that with this solution, only min length will be respected.

I also see that this solution kind of becomes "ugly".

Maybe it is better if we make it so with basic auth cli `ivy auth add` -> choose `basic` we prompt the user for the limits/default constraints?

like

```bash
> Choose auth provider
$ basic

> Choose minimum length (default 8)? Y or N
$ Y

> Should it contain special symbols? Y or N
$ N
```

@nielsbosma your thoughts?

@ArtemLazarchuk I am putting `ON HOLD` for this until we figure out what we want to do exactly.

artem-ivy-ai:
### Staging removed

PR #4323, opened 2026-04-28
feat: add self-improving feedback loop to all 21 skills

nielsbosma:
Each skill now reads .ivy/learnings/{skill-name}.md at the start (if it
exists) and evaluates its run at the end, appending lessons learned when
corrections were needed. This lets skills improve across sessions rather
than repeating the same mistakes.


artem-ivy-ai:
### Staging removed

PR #4344, opened 2026-04-30
[00005] Fix race condition in job monitor when killing already-exited process

nielsbosma:
# Summary

## Changes

Fixed a race condition in `ProcessExtensions.KillProcess` and `ProcessExtensions.KillProcessAsync` where accessing `Process.Id` after the process had already exited caused an `InvalidOperationException`. The methods now capture the process ID before attempting to kill, so logging statements use the cached value instead of accessing the disposed process object.

## API Changes

None.

## Files Modified

- `Ivy-Framework/src/Ivy/Helpers/ProcessExtensions.cs` — Added process ID capture before kill operations in both sync and async methods

## Commits

- 115e8a24e [00005] Fix race condition in ProcessExtensions when killing already-exited process


artem-ivy-ai:
### Staging removed

PR #4351, opened 2026-04-30
[00133] Fix race condition in auth endpoints when AppSession disconnects during OAuth flow

### LINKED ISSUE #4253: (auth): Race condition in auth endpoints
### Description

`/ivy/auth/oauth-login` and `/ivy/auth/callback` both depend (the latter only sometimes) on being able to access the `AppSession` that referred the user to them. Problem: there's absolutely no guarantee that it will still exist.

dcrjodle:
## Summary

Added deferred session removal to prevent a race condition where the OAuth callback endpoint fails because the `AppSession` was already removed when the SignalR connection dropped during the OAuth redirect. When `AppHub.OnDisconnectedAsync` detects a pending OAuth callback for the disconnecting connection, it schedules a 2-minute deferred removal instead of immediately removing the session. The deferred removal is cancelled if the browser reconnects after the OAuth redirect completes.

## API Changes

- `IOAuthCallbackRegistry.HasPendingForConnection(string connectionId)` — new interface method to check if any pending OAuth callbacks exist for a connection
- `AppSessionStore.ScheduleDeferredRemoval(string connectionId, TimeSpan delay, Func<string, Task> cleanupAction)` — schedules delayed session removal with a callback
- `AppSessionStore.CancelDeferredRemoval(string connectionId)` — cancels a pending deferred removal (returns bool)
- `AppSessionStore.HasDeferredRemoval(string connectionId)` — checks if a deferred removal is pending
- `AppHub` constructor now accepts `IOAuthCallbackRegistry` (injected via DI, already registered as singleton)

## Files Modified

- **src/Ivy/Core/Auth/OAuthCallbackRegistry.cs** — Added `HasPendingForConnection` to interface and implementation
- **src/Ivy/Core/Server/AppSessionStore.cs** — Added deferred removal scheduling, cancellation, and disposal cleanup
- **src/Ivy/Core/Server/AppHub.cs** — Injected `IOAuthCallbackRegistry`, added OAuth-aware disconnect deferral, extracted `CleanupSessionAsync` shared method, added deferred removal cancellation on reconnect
- **src/Ivy.Test/Auth/OAuthCallbackRegistryTests.cs** — New: 6 tests for `HasPendingForConnection`
- **src/Ivy.Test/AppSessionStoreTests.cs** — New: 5 tests for deferred removal scheduling and cancellation

## Commits

- `e66713349` [00133] Fix race condition in auth endpoints when AppSession disconnects during OAuth flow
- `228b2035b` [00133] Add tests for OAuthCallbackRegistry and AppSessionStore deferred removal

## Verifications

- ✅ DotnetBuild
- ✅ DotnetFormat
- ✅ DotnetTest
- ✅ CheckResult

Closes https://github.com/Ivy-Interactive/Ivy-Framework/issues/4253

### DIFF
--- a/src/Ivy.Test/AppSessionStoreTests.cs
+++ b/src/Ivy.Test/AppSessionStoreTests.cs
@@ -1,0 +1,92 @@
+using Ivy.Core.Server;
+
+namespace Ivy.Test;
+
+public class AppSessionStoreTests
+{
+    [Fact]
+    public async Task ScheduleDeferredRemoval_RemovesAfterDelay()
+    {
+        using var store = new AppSessionStore();
+        var removed = false;
+
+        store.ScheduleDeferredRemoval("conn-1", TimeSpan.FromMilliseconds(50), async connId =>
+        {
+            removed = true;
+            await Task.CompletedTask;
+        });
+
+        Assert.True(store.HasDeferredRemoval("conn-1"));
+        Assert.False(removed);
+
+        await Task.Delay(200);
+
+        Assert.True(removed);
+        Assert.False(store.HasDeferredRemoval("conn-1"));
+    }
+
+    [Fact]
+    public async Task CancelDeferredRemoval_PreventsRemoval()
+    {
+        using var store = new AppSessionStore();
+        var removed = false;
+
+        store.ScheduleDeferredRemoval("conn-1", TimeSpan.FromMilliseconds(200), async connId =>
+        {
+            removed = true;
+            await Task.CompletedTask;
+        });
+
+        Assert.True(store.HasDeferredRemoval("conn-1"));
+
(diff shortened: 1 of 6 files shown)


zachwolfe:
@dcrjodle Thanks for looking into this with Tendril. I added a commit that opens connected account logins in the same tab, since that is now possible thanks to this PR. On the one hand, prolonging the `AppSession` lifetime like this feels like a bit of a hack. On the other hand, the only case I'm aware of that wouldn't be solved by this PR is signing in with Clerk (without an extra tab), since that requires an active frontend connection in each endpoint instead of just the server-side `AppSession`. And solving that case would probably require much worse hacks. So I think we should merge this.

artem-ivy-ai:
### Staging removed

Issue #4371, opened 2026-05-01
tree: I feel like there are some things off with the left vertical line


rorychatt:
### Description

<img width="250" height="195" alt="Image" src="https://github.com/user-attachments/assets/6dcb9db9-28e9-4210-acc3-bf931d8aad95" />
<img width="279" height="440" alt="Image" src="https://github.com/user-attachments/assets/71e49070-0391-4afb-8b7c-bd5a1ec31abf" />

The files maybe need to be all the way to the left (bottom md files)?

phonofidelic:
Here I am

rorychatt:
src/Ivy.Samples -> dotnet run

`samples.ivy.app` locally

rorychatt:
`Tree`

rorychatt:
https://samples-ivy-staging.sliplane.app/widgets/tree

phonofidelic:
Made a PR in my forked repo: [fix: remove indentation on leaf nodes in Tree widget
](https://github.com/phonofidelic/Ivy-Framework/pull/1).

rorychatt:
@phonofidelic when you link an issue, you don't need to write comments like this above ^ it gets automated in GitHub.

you can simply add "closes #4371" as an example.

PR #4377, opened 2026-05-05
fix: remove indentation on leaf nodes in Tree widget

phonofidelic:
Addresses https://github.com/Ivy-Interactive/Ivy-Framework/issues/4371

## Problem
A span was rendered to the left of leaf nodes that had siblings with children which made the leaf nodes appear to be on the same level as their siblings' children. 

## Solution
Removed `hasSiblingWithChildren` and `isNested` props from `TreeItem`.

### Before:
ivy-framework-tree- widget-indentation_before

### After:
ivy-framework-tree- widget-indentation_after

### DIFF
--- a/src/frontend/src/widgets/tree/TreeItem.tsx
+++ b/src/frontend/src/widgets/tree/TreeItem.tsx
@@ -53,8 +53,6 @@ interface TreeItemWidgetProps {
   item: MenuItem;
   density?: Densities;
   rowActions?: MenuItem[];
-  hasSiblingWithChildren?: boolean;
-  isNested?: boolean;
   onItemClick: (item: MenuItem) => void;
   onRowActionClick?: (item: MenuItem, action: MenuItem) => void;
 }
@@ -63,8 +61,6 @@ export const TreeItem: React.FC<TreeItemWidgetProps> = ({
   item,
   density,
   rowActions,
-  hasSiblingWithChildren,
-  isNested,
   onItemClick,
   onRowActionClick,
 }) => {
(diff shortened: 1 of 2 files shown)


rorychatt:
@ArtemKhvorostianyi review this pls

src/frontend/src/widgets/tree/TreeItem.tsx
```diff
@@ -53,8 +53,6 @@ interface TreeItemWidgetProps {
   item: MenuItem;
   density?: Densities;
   rowActions?: MenuItem[];
-  hasSiblingWithChildren?: boolean;
```
rorychatt, line 56: this removed in FE api. Does this exist in BE api?

src/frontend/src/widgets/tree/TreeItem.tsx
```diff
@@ -63,8 +61,6 @@ export const TreeItem: React.FC<TreeItemWidgetProps> = ({
   item,
   density,
   rowActions,
-  hasSiblingWithChildren,
```
rorychatt, line 66: same comment here

src/frontend/src/widgets/tree/TreeItem.tsx
ArtemKhvorostianyi: This was never part of BE API; it was an internal FE-only prop used for rendering alignment.

hasSiblingWithChildren was an internal frontend-only TreeItem prop, not part of the FE/BE MenuItem contract.
It does not exist in backend API (MenuItem in src/Ivy/Shared/MenuItem.cs), so removing it here does not affect BE integration.

ArtemKhvorostianyi (approved):
Built on my machine
LGTM, verified examples in samples and docs
Also left a comment on @rorychatt feedback. Verified, code implementation is fine

PR #4411, opened 2026-05-12
[Inputs]: restore clear button visibility in SearchVariant

### LINKED ISSUE #4385: No button to clear input in samples search
## Summary
There is no button to clear the input field in the samples search functionality.

## Additional Context
Reported in Slack: https://ivy-interactive.slack.com/archives/C09PMA599RC/p1778070879777469?thread_ts=1778070879.777469&cid=C09PMA599RC

## Reported by
@dcrjodle

---

ArtemLazarchuk:
## Problem

The X (clear) button in **search** inputs stopped appearing after the ghost/affix layout refactor in `2bf207004`.

The refactor introduced a `showClear` variable that tied button visibility to `props.nullable`, but `nullable` defaults to **false** for non-nullable string types — so the button was silently hidden for most search inputs.

## Root cause

Before the refactor, the clear button rendered directly in JSX with a simple `hasValue && !disabled` check.

The refactor unified the trailing cluster logic into a `showClear` flag, but incorrectly applied the **nullable** guard (which is appropriate for text/default inputs) to the **search** variant as well.

Search inputs should **always** show a clear button when they have a value — that is standard UX for search fields. Whether the underlying state type is nullable is an implementation detail and should not control button visibility here.

## Fix

Simplified `showClear` in `SearchVariant` to `!props.disabled && hasValue`, restoring the pre-refactor behavior.
Знімок екрана 2026-05-13 о 02 43 39

### DIFF
--- a/src/frontend/src/widgets/inputs/TextInputWidget/variants/SearchVariant.tsx
+++ b/src/frontend/src/widgets/inputs/TextInputWidget/variants/SearchVariant.tsx
@@ -91,9 +91,7 @@ export const SearchVariant: React.FC<SearchVariantProps> = ({
   const hasAffixes = hasPrefix || hasSuffix;
   const ghostAffixChrome = Boolean(props.ghost && hasAffixes);
   const ghostSuffixLayout = Boolean(props.ghost && hasSuffix);
-  const showClear = ghostSuffixLayout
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4420, opened 2026-05-14
Add unconfigured plugin state

zachwolfe:
## Summary

- Plugins with missing/invalid configuration are now kept loaded in an **Unconfigured** state instead of being unloaded or moved to failed plugins
- Adds `ReconfigurePlugin(pluginId)` for the host to trigger re-validation after updating config — transitions plugins between Active/Unconfigured as needed
- Removes unused `ConfigureServices` from `IIvyPlugin` interface (all implementations were empty)
- Renames `GetLoadedPluginIds` → `GetActivePluginIds` (returns only fully-configured plugins)
- Adds `GetUnconfiguredPlugins()` returning schema + validation errors for UI consumption
- Updates `PluginManagerApp` with a new "Unconfigured Plugins" section

## State model

```
Unloaded  →  Loaded (Unconfigured)  →  Loaded (Active)
                                     ←  (via ReconfigurePlugin)
```

### DIFF
--- a/src/Ivy.Plugin.Abstractions/IIvyPlugin.cs
+++ b/src/Ivy.Plugin.Abstractions/IIvyPlugin.cs
@@ -1,12 +1,8 @@
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
-
 namespace Ivy.Plugins;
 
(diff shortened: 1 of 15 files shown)


artem-ivy-ai:
### Staging removed

PR #4430, opened 2026-05-15
Port filter-query-editor package into the frontend

dcrjodle:
## Summary

Vendors the **filter-query-editor** package (formerly the external `Ivy-Query-Editor` repo / npm `filter-query-editor@2.2.0`) directly into the frontend at `src/frontend/src/lib/filter-query-editor/`, so the filter query parser/editor is no longer a separate repo or npm dependency. Also fixes two reported bugs.

### What changed

- **Vendored the full package** (42 files): parser, ANTLR-generated parser code, CodeMirror editor component + extensions, formatter, evaluator, semantic validator, types, hooks, styles, plus a local `index.ts` re-exporting the same public API. Demo app and tests were excluded.
  - The ANTLR4-generated parser (`generated/FiltersParser.ts` etc.) is **checked in**. The source repo regenerated it at build time via `antlr-ng`; this build does not run ANTLR codegen, so vendoring the generated output is what keeps the `parseQuery` path (used by the LLM filter fallback in `DataTableFilterOption`) working reliably.

- **Fixed RangeError in the CodeMirror newline-prevention transaction filter** (`useCodeMirror.ts`): the replacement transaction used `tr.newDoc.length` for the range bound, but the change applies against the transaction's *start state*. Any edit that both grew the document and introduced a newline (e.g. pasting multi-line text) produced an out-of-bounds range and threw a console `RangeError`. Now uses `tr.startState.doc.length`.

- **Dependency swap**: removed `filter-query-editor@2.2.0`; added the two transitive deps the frontend didn't already have (`antlr4ng`, `@codemirror/autocomplete` — pinned to `6.20.1` to avoid downgrading the version already resolved transitively for `@uiw/react-codemirror`). Repointed `DataTableFilterOption.tsx` from `"filter-query-editor"` to `"@/lib/filter-query-editor"`.

- **Conformed vendored sources** to the frontend's stricter `noUnusedLocals`/`noUnusedParameters` tsconfig (the source repo did not enforce these), and applied the repo's lint/format conventions via the pre-commit hook.

### Not addressed (separate issue)

This PR does not change the LLM `ParseFilter` fallback *behavior* — it only ensures the fallback's `parseQuery` works by vendoring the generated parser. The reported case of a seemingly-valid query (`[Age] < 45`) falling through to the LLM is a column-name match / column-loading-timing issue in `DataTableFilterOption.tsx`'s `queryEditorColumns` mapping, which is intentionally left out of scope here.

## Test plan

- [x] `tsc -b` passes (0 errors)
- [x] `pnpm run build` (full Vite production build) passes — editor bundles into `DataTableWidget`
- [x] `vp lint` on vendored dir: 0 errors (1 pre-existing upstream warning, unchanged)
- [ ] Manual: open a DataTable with filtering, type a valid filter (e.g. `[Age] > 100`), confirm it parses without falling back to the LLM
- [ ] Manual: paste a multi-line string into the filter input, confirm no `RangeError` in console and newlines are collapsed to spaces
- [ ] Manual: confirm autocomplete, syntax highlighting, and the LLM fallback (invalid query + Enter) still work


artem-ivy-ai:
### Staging removed

PR #4432, opened 2026-05-15
Fix Enter inserting a space and empty filter results hiding the DataTable

dcrjodle:
## Summary

Two bugs in the DataTable filter UI, found after the filter-query-editor port (#4430):

### Bug 1 — pressing Enter to submit a filter appended a space
The single-line filter editor's `Enter` keymap returned `false` when no `onApply` was wired (the host handles Enter at the React level). Returning `false` let `defaultKeymap` insert a `"\n"`, which the single-line transaction filter then rewrote to `" "` — leaving a stray trailing space in the query every time you pressed Enter.

- `Enter` / `Mod-Enter` / `Shift-Enter` are now always consumed (`return true`) so no newline is ever produced in this single-line editor.
- The transaction filter now **drops a pure newline insertion entirely** (no-op transaction) instead of substituting a space. Genuine multi-line **pastes** are still flattened to a single line (newlines → space, runs collapsed), and the RangeError-safe replacement range from #4430 is preserved.

### Bug 2 — a filter that matched no rows removed the whole table
`useEmptyRows` short-circuited filler-row calculation when `visibleRows === 0`, so a *loaded-but-empty* result produced `totalRows = 0`. The grid then collapsed to ~0 height inside its flex container and the entire table appeared to be removed from the page.

- `useEmptyRows` no longer short-circuits at `visibleRows === 0`; it still computes filler rows so the grid keeps its header and a **full-height empty body**. (`hasMore` / `scrollContainerHeight === 0` guards are unchanged.)
- `useDataLoading` now explicitly clears the stale Arrow table and sets `visibleRows = 0` (marking the load complete) when a query returns no Arrow stream, instead of leaving the previous result in place.

Expected behavior after this PR: an empty filter result keeps the table widget rendered — header, columns, toolbar — with an empty body and no rows, exactly as requested.

## Test plan

- [x] `tsc -b` passes (after `pnpm install`; the pre-existing `playwright.config.ts` `@types/node` noise is unrelated and resolves on a clean install)
- [x] `pnpm run build` passes
- [x] `vp lint` on the 3 changed files: 0 warnings, 0 errors
- [ ] Manual: type a filter, press Enter — no trailing space is added to the input
- [ ] Manual: paste a multi-line string — it flattens to one line (no RangeError)
- [ ] Manual: enter a filter that matches no rows — the table stays visible with header/columns and an empty body (widget not removed)
- [ ] Manual: clear the filter — rows return as before

### DIFF
--- a/src/frontend/src/lib/filter-query-editor/components/useCodeMirror.ts
+++ b/src/frontend/src/lib/filter-query-editor/components/useCodeMirror.ts
@@ -78,25 +78,30 @@ export function useCodeMirror({
       // Basic setup
       history(),
       keymap.of([
-        // Custom keymaps for applying filter
+        // Custom keymaps for applying filter. Enter is always consumed so a
+        // newline is never inserted into this single-line editor: the host
+        // component handles Enter (submitting the filter) at the React level.
+        // Returning false here would let defaultKeymap insert "\n", which the
+        // transaction filter then turned into a stray trailing space.
         {
           key: "Enter",
           run: () => {
-            if (onApplyRef.current) {
-              onApplyRef.current();
-              return true;
-            }
-            return false;
+            onApplyRef.current?.();
+            return true;
           },
         },
         {
           key: "Mod-Enter", // Cmd+Enter on Mac, Ctrl+Enter on Windows/Linux
           run: () => {
-            if (onApplyRef.current) {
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #4434, opened 2026-05-15
[00273] Fix unused catch variable in filter-query-editor formatting

dcrjodle:
Replace `catch (error)` with `catch` in `src/frontend/src/lib/filter-query-editor/components/extensions/formatting.ts` to suppress the unused variable lint warning.

---

**Commits:**
- 4231059d1 — Remove unused catch variable in filter-query-editor formatting


artem-ivy-ai:
### Staging removed

PR #4437, opened 2026-05-15
[00284] Remove 'No rows to display' message from DataTable

dcrjodle:
Removes the "No rows to display" message div from the DataTable component. When there are no visible rows, the table now simply shows as empty (headers only) without any message text. The table collapse behavior remains intact.

---

**Commits:**
- 664654d52 [00284] Remove 'No rows to display' message from DataTable


artem-ivy-ai:
### Staging removed

PR #4443, opened 2026-05-18
(datatable): fix header cell styling when sorting

### LINKED ISSUE #4424: Column sorting breaks styling
## Summary
Clicking a column to sort it causes the column styling to break completely.

**Expected:** Clicking a column should sort it while maintaining proper styling.

**Actual:** After clicking to sort, the styling is broken. The sorted column appears with incorrect styling, including misaligned content and visual artifacts.

## Additional Context
The issue is visible in the provided screenshots. When sorting the "Age" column, the header styling becomes corrupted with black backgrounds extending incorrectly, and content alignment is lost.

---

**Reported by** @dcrjodle

[Slack thread](https://ivy-interactive.slack.com/archives/C09PMA599RC/p1778831432827619?thread_ts=1778831432.827619&cid=C09PMA599RC)

---

ArtemKhvorostianyi:
Had a problem:
image

Fixed:

https://github.com/user-attachments/assets/d8f34516-df4c-446a-ac2d-dccf34fa2d4d

The black header came from Glide’s column selection. That paints the header with accentColor (your theme’s primary, often near-black).

The pointer guard wasn’t reliable enough, so the fix is more direct:

- columnSelect: "none" when sorting is enabled — Glide no longer selects columns on header mousedown, so the accent fill never appears.
- Strip column selection in useGridSelection — Any column selection that still slips through is cleared immediately.
- Clear columns after sort — Safety net in useHeaderMenu so a sorted column doesn’t stay selected.
- Sort arrows still use indicatorIcon only; header background should match other columns.

### DIFF
--- a/src/frontend/src/widgets/dataTables/components/GridContainer.tsx
+++ b/src/frontend/src/widgets/dataTables/components/GridContainer.tsx
@@ -7,6 +7,7 @@ import DataEditor, {
   GridMouseEventArgs,
   GridSelection,
   GroupHeaderClickedEventArgs,
+  HeaderClickedEventArgs,
   Highlight,
   Item,
   SpriteMap,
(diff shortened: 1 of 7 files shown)


artem-ivy-ai:
### Staging removed

PR #4452, opened 2026-05-20
fix: format chart tooltip headers using active axis tick formatters

rorychatt:
This PR formats chart tooltip category headers using the tick formatter defined on the active category axis (X-axis or Y-axis depending on chart orientation). This resolves the issue where raw ISO dates like \2026-05-19T20:00:00Z\ were displayed in tooltips instead of formatted labels.


artem-ivy-ai:
### Staging removed

PR #4487, opened 2026-05-25
[Staging]: show deploy timestamp instead of "Version 0" on staging

ArtemLazarchuk:
## Problem

On staging (PR preview) deployments, the sidebar showed **"Version 0"** — 
meaningless and confusing. This happens because staging builds don't set 
an assembly version, so .NET defaults to `0.0.0.0`.

### Before
Знімок екрана 2026-05-25 о 15 17 40
 
### After
Знімок екрана 2026-05-25 о 15 15 59

## How it works

The fix checks the assembly version at startup:

- **`0`** (no version set) → shows `"Deployed May 25 · 11:19 UTC"` — 
  the time the server process started
- **`1.2.58`** (real release version) → shows `"Version 1.2.58"` — 
  behaviour unchanged

```csharp
var version = typeof(Server).Assembly.GetName().Version!.ToString().EatRight(".0");

string versionLabel;
if (version == "0")
{
    var buildDate = DateTime.UtcNow.ToString("MMM d · HH:mm UTC");
    versionLabel = $"Deployed {buildDate}";
}
else
{
    versionLabel = $"Version {version}";
}
```

## When does each case apply?

**Shows "Deployed ..."**
- PR preview deployed on Sliplane — built from source without 
  `IvyPackageVersion`, so `GenerateAssemblyInfo` is skipped and 
  version defaults to `0.0.0.0`
- Local development via `dotnet watch` — same reason, no version 
  is set locally

**Shows "Version X.Y.Z"**  
- Production builds that use `IvyPackageVersion` (e.g. `1.2.58`) 
  — the NuGet package DLL already has the version baked in from 
  the release pipeline

## Changed files

- `src/Ivy.Samples.Shared/SamplesServer.cs`
- `src/Ivy.Docs.Shared/DocsServer.cs`

### DIFF
--- a/src/Ivy.Docs.Shared/DocsServer.cs
+++ b/src/Ivy.Docs.Shared/DocsServer.cs
@@ -24,13 +24,26 @@ public static class DocsServer
         server.Services.AddTransient<IIvyDocsQuestionsClient>(sp => sp.GetRequiredService<IvyDocsQuestionsClient>());
 
         var version = typeof(Server).Assembly.GetName().Version!.ToString().EatRight(".0");
-        server.SetMetaTitle($"Ivy Docs {version}");
+
+        // For staging deployments (version = "0"): show build date from DLL timestamp
+        string versionLabel;
+        if (version == "0")
+        {
+            var buildDate = DateTime.UtcNow.ToString("MMM d · HH:mm UTC");
+            versionLabel = $"Deployed {buildDate}";
+        }
+        else
+        {
+            versionLabel = $"Version {version}";
+        }
+
+        server.SetMetaTitle($"Ivy Docs {versionLabel}");
 
         var appShellSettings = new AppShellSettings()
             .Header(
                 Layout.Vertical().Padding(2)
                 | new IvyLogo()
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

src/Ivy.Docs.Shared/DocsServer.cs
```diff
@@ -24,13 +24,26 @@ public static async Task RunAsync(ServerArgs? args = null)
         server.Services.AddTransient<IIvyDocsQuestionsClient>(sp => sp.GetRequiredService<IvyDocsQuestionsClient>());
 
         var version = typeof(Server).Assembly.GetName().Version!.ToString().EatRight(".0");
-        server.SetMetaTitle($"Ivy Docs {version}");
+
+        // For staging deployments (version = "0"): show build date from DLL timestamp
```
rorychatt, line 28: this logic is reused in samples server and should be brought out as a helper function exposed only internally

rorychatt (commented):
.

PR #4500, opened 2026-05-26
[Details]: page crash when model contains nested ToDetails()

### LINKED ISSUE #4488: [Details]: page fails to open with TargetInvocationException
## Summary

The `Details` page does not open on both `samples.ivy.app` and `docs.ivy.app`.

Instead of rendering the page, the application throws a `TargetInvocationException`.

---

## Steps to Reproduce

1. Open `samples.ivy.app` or `docs.ivy.app`
2. Search for `details`
3. Click on the `Details` page/item
4. Observe the result

---

## Actual Result

The page fails to load and shows the following error:

```txt
TargetInvocationException
Exception has been thrown by the target of an invocation.
```

---

## Expected Result

The `Details` page should open normally and render its content without throwing an exception.

---

## Additional Information

* Reproducible on both environments:

  * `samples.ivy.app`
  * `docs.ivy.app`
* The issue occurs immediately after clicking the `Details` item.

Image

ArtemLazarchuk:
## Summary

Fixes a crash when opening the Details sample/docs page (`TargetInvocationException`). Nested `.ToDetails()` fields (e.g. `Address = new { ... }.ToDetails()`) were incorrectly handled as navigation properties, causing `NavigationPropertyBuilder` to read uninitialized `ViewBase.Id`.

## Problem

After navigation-property support in `DetailsBuilder._Scaffold()`, any complex class field got `NavigationPropertyBuilder`. Fields whose value is a nested `DetailsBuilder` (from `.ToDetails()` in an anonymous type) match that rule.

`NavigationPropertyBuilder` resolves display text via `Name` / `Title` / `Id`. `ViewBase.Id` is only set when the view is mounted in the widget tree. A `DetailsBuilder` stored only in the model has no id yet → `InvalidOperationException`, surfaced as `TargetInvocationException`.

This broke `DetailsApp` tabs that use nested details (Nested, Density) and failed immediately on app open because all tabs are built at once.

## Solution

- **DetailsBuilder**: skip `NavigationPropertyBuilder` when the field type implements `IView` (e.g. `DetailsBuilder<T>`).
- **NavigationPropertyBuilder**: return `IView` values as-is for rendering; do not read `Id` on `ViewBase` when resolving entity display strings.

https://github.com/user-attachments/assets/1a4a22e9-0f92-4039-8710-406ba827b20d

### DIFF
--- a/src/Ivy.Test/Views/DetailsBuilderNavigationTests.cs
+++ b/src/Ivy.Test/Views/DetailsBuilderNavigationTests.cs
@@ -122,4 +122,36 @@ public class DetailsBuilderNavigationTests
         Assert.Equal("hello", defaultBuilder.Build("hello", new Product()));
         Assert.Equal(3.14m, defaultBuilder.Build(3.14m, new Product()));
     }
+
+    [Fact]
+    public void NavigationPropertyBuilder_Build_NestedDetailsBuilder_ReturnsView()
+    {
+        var address = new { Street = "123 Elm St", City = "Springfield" }.ToDetails();
+        var builder = new NavigationPropertyBuilder<object>();
+        var result = builder.Build(address, new object());
+        Assert.Same(address, result);
+    }
+
+    [Fact]
+    public void ToDetails_WithNestedToDetails_BuildsWithoutThrowing()
+    {
+        var record = new
+        {
+            FirstName = "John",
+            Address = new { Street = "123 Elm St", City = "Springfield" }.ToDetails(),
+            Employment = new { Company = "Acme" }.ToDetails(),
+        };
+
+        var details = record.ToDetails().RemoveEmpty().Build();
+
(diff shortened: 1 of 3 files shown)


rorychatt:
@ArtemLazarchuk test if diff viewer in tendril still properly works - can click a file in the list and navigation happens

rorychatt (changes requested):
needs verification in ivy tendril diff viewer

ArtemLazarchuk:
@rorychatt Tested the diff viewer in Tendril. Verified locally

- File selection from the list works correctly
- Navigation opens the expected file diff
- No errors or unexpected behavior observed


https://github.com/user-attachments/assets/200aa3a4-bc40-420f-bd70-11c5d2ca1418

artem-ivy-ai:
### Staging removed

PR #4521, opened 2026-05-28
(ActivityHeatmap): add builder API and harden heatmap behavior

phonofidelic:
## Summary

- Add a fluent `ToActivityHeatmap()` builder over `IEnumerable<T>`/`IQueryable<T>` that projects data into the heatmap via a **dimension** (date/time axis) and a **measure** (aggregated value per cell), using the pivot-table engine. The dimension is truncated to cell granularity (day/hour) before pivoting so the chosen aggregation sees every row in a cell instead of degenerating into a per-row sum. Default aggregation is **Sum** (numeric types and nullable variants); use `.Measure(name, q => …)` on the builder for custom pivot aggregators. The measure name is forwarded as **`ValueLabel`** in tooltips.
- Support hourly intervals end to end: new `ActivityInterval` (`Daily`/`Hourly`) and `ValueLabel` props on the widget, with the builder auto-detecting `Hourly` when dimension values carry a non-zero time component. The frontend renders an hourly "punchcard" grid (24 rows × day columns, default 7-day window) with sticky day/hour labels, scroll-to-latest, and a tooltip that uses the measure name as its value label.
- Move grid construction and level bucketing entirely into the frontend; remove the now-unused C# `ActivityHeatmapGrid` (`BuildGrid`/`BuildGridFromRange`/`GetLevel`) and its tests. `Activity` now carries an optional `Hour`.
- Rewrite README around the builder API, refresh `.samples` with basic/optional-props/hourly examples, and rewrite tests to cover the builder's truncation and value-conversion helpers (`InternalsVisibleTo` added for the test project).

**Updated basic usage example:**
ActivityHeatmap updated builder API

**Hourly interval example:**
ActivityHeatmap with hourly interval - removed aggregation prop

ActivityHeatmap with hourly interval - dark

## Test plan

# Test plan
- [ ] `dotnet test` for `Ivy.Widgets.ActivityHeatmap.Tests` (builder truncation/aggregation/`ToHour`/`ToDateOnly`).
- [ ] In the `.samples` project, verify the Basic Usage, Optional Properties, and Hourly Interval cards render and that day-click logs fire.
- [ ] Toggle day/month labels, color scheme, and date range in the Optional Properties example.
- [ ] Confirm the hourly punchcard scrolls to the latest day on load and the tooltip shows `<measure name>: <value>`.


phonofidelic:
@claude review this

src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
```diff
@@ -254,20 +255,21 @@ export function ActivityHeatmap({
 
       {showTooltip && <div
         ref={tooltipDiv}
-        className="fixed opacity-0 z-50 bg-card text-xs text-foreground pointer-events-none rounded-[4px] px-2 py-3"
+        className="fixed opacity-0 z-50 bg-card text-xs text-foreground pointer-events-none rounded-[4px] px-2 py-3 min-w-40 max-w-64"
```
rorychatt: width and stuff should be controlled with backend.

see Size API it supports both min and max

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -26,20 +26,16 @@ public override object Build()
...
-
-        return Layout
-            .Vertical()
-            .Gap(4)
-            .Width(Size.Auto().At(Breakpoint.Tablet).And(Breakpoint.Desktop, Size.Fit()))
+            .Select(d => new { Date = d, Count = rng.Next(1, 20) });
 
-            | Text.H1("ActivityHeatmap")
-            | Text.H2("Basic Usage")
+        var basicUsageHeading = new ArticleHeading("basic-usage", "Basic Usage", 1);
```
rorychatt: Fluent API over constructors like this

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +44,90 @@ public override object Build()
...
             .Select(start.AddDays)
             .Where(_ => rng.NextDouble() > 0.4)
-            .Select(d => new Activity {{ Date = d, Count = rng.Next(1, 20) }})
-            .ToArray();
+            .Select(d => new Activity {{ Date = d, Count = rng.Next(1, 20) }});
 
-        return new ActivityHeatmap().Data(data);
+        return Layout.Vertical()
+            | data.ToActivityHeatmap()
+                .Dimension(""Days"", d => d.Date)
```
rorychatt: First time I see syntax like this, likely this should be an enum ie Activity.Days

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +44,90 @@ public override object Build()
...
-                        | Text.P(selectedColor.Value.ToString()))
-                    | (Layout.Horizontal().Gap(2).Width(Size.Grow())
-                        | showDayLabels.ToBoolInput().Label("Show day labels")
-                        | showMonthLabels.ToBoolInput().Label("Show month labels"))
-                    | nullableRange.ToDateRangeInput())
+        var themeSelector = new DropDownMenu(@evt =>
+            {
+                ThemeMode selectedTheme = @evt.Value switch
+                {
+                    "Light" => ThemeMode.Light,
```
rorychatt: use enum

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +44,90 @@ public override object Build()
...
+                };
+                client.SetThemeMode(selectedTheme);
+            },
+            new Button("Theme").Variant(ButtonVariant.Link).Icon(Icons.SunMoon),
+            MenuItem.Default("Light").Icon(Icons.Sun),
+            MenuItem.Default("Dark").Icon(Icons.Moon),
+            MenuItem.Default("System").Icon(Icons.Computer));
+
+        var mainContent = Layout.Vertical().Width(Size.Full())
+            .Gap(20)
```
rorychatt: usually better not to use any custom margins-paddings-gaps and let framework just handle it with nice defaults

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +44,90 @@ public override object Build()
...
-            | new Card(new ActivityHeatmap()
-                    .Data(data)
-                    .StartDate(startDate)
-                    .EndDate(endDate)
-                    .ColorScheme(selectedColor.Value)
-                    .ShowDayLabels(showDayLabels.Value)
-                    .ShowMonthLabels(showMonthLabels.Value)
-                    .OnDayClick(day => Console.WriteLine($"Clicked {day.Date}: {day.Count}")))
+        var article = new Article(
+            new object[] { mainContent })
```
rorychatt: type unsafe code

src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
rorychatt: can we reuse styling from other components?

src/widgets/Ivy.Widgets.ActivityHeatmap/ActivityHeatmap.cs
```diff
@@ -19,6 +19,8 @@ public record ActivityHeatmap : WidgetBase<ActivityHeatmap>
 
     [Prop] public bool ShowDayLabels { get; init; } = true;
 
+    [Prop] public string ValueLabel { get; init; } = "Count";
```
rorychatt: does it need this default? why not nullable

rorychatt (changes requested):
See my feedback

rorychatt:
@claude review this

src/widgets/Ivy.Widgets.ActivityHeatmap/ActivityHeatmap.cs
phonofidelic: The frontend uses `valueLabel` in the tooltip.  There it also defaults to "Count", but maybe the tooltip  should just show the value and no label if it is not set?

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
phonofidelic: I was looking at the [BarChart documentation](https://docs.ivy.app/widgets/charts/bar-chart#barchart). But maybe it would be better to use an enum?
```csharp
public enum ActivityDimension
{
    Days
}

public ActivityHeatmapBuilder<TSource> Dimension(ActivityDimension name, Expression<Func<TSource, object>> selector)
{
    _dimension = new Dimension<TSource>(name.ToString(), selector);
    return this;
}
```

src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
phonofidelic: From what I understand, the chart tooltips are rendered by the `echarts-for-react` library in the frontend. The backend provides props through the `ChartTooltip` record, but do not provide specific styling. I've tried to match the tooltip styling based on the rendered `ReactECharts` tooltip  component.

Image

I'll take a closer look at matching the `ChartTooltip` props sent from the backend in other chart widgets.

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
rorychatt: -> ai will vibe code based on this

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +45,88 @@ public override object Build()
...
+            | Text.H2(optionalPropsUsageHeading.Text).Anchor(optionalPropsUsageHeading.Id)
+            | new CodeBlock(@$"public class ActivityHeatmapOptionalProps : ViewBase
+{{
+    public override object Build()
+    {{
+        return Layout.Vertical()
+            | data
+                .ToActivityHeatmap()
+                .Dimension(ActivityDimension.Days, d => d.Date)
+                .Measure(""Downloads"", e => e.Sum(d => d.Count))
```
rorychatt: there should be some type regarding the structure upon which the sample is based (i.e Video that has downloads, views, likes etc and you use those types)

src/widgets/Ivy.Widgets.ActivityHeatmap/.samples/Program.cs
```diff
@@ -48,62 +45,88 @@ public override object Build()
...
+                MenuItem.Default(nameof(ThemeMode.Dark)).Icon(Icons.Moon),
+                MenuItem.Default(nameof(ThemeMode.System)).Icon(Icons.Computer))
+            .OnSelect(@evt =>
+            {
+                if (Enum.TryParse<ThemeMode>(@evt.Value.ToString(), out var theme))
+                    client.SetThemeMode(theme);
+            });
+
+        var mainContent = Layout.Vertical().Width(Size.Full())
+            | Text.H1("ActivityHeatmap")
```
rorychatt: not a big fan of H1 usage

rorychatt: h2

src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
rorychatt: break out css of tooltips in charts into charttooltip ,.,.,, -> like common styling files and reuse it

so you dont get a dublicate place with basically same code, Edit one place and hte other is outdated

src/widgets/Ivy.Widgets.ActivityHeatmap/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Ivy.Widgets.ActivityHeatmap.Tests")]
```
rorychatt, line 3: is this safe? (yes?)

src/widgets/Ivy.Widgets.ActivityHeatmap/ActivityHeatmap.cs
rorychatt: enum

src/widgets/Ivy.Widgets.ActivityHeatmap/ActivityHeatmap.cs
```diff
@@ -128,3 +133,8 @@ public record Activity
     public DateOnly Date { get; init; }
     public int Count { get; init; }
```
rorychatt, line 84: use selectors

rorychatt (changes requested):

phonofidelic:
@claude review this

phonofidelic:
@claude review this

rorychatt (changes requested):
gave feedback in DM-s

- don't use true as argument in self-descriptive fluent methods like `isTrueish(true)` -> `isTrueish()`
- console writelines are generally okay, but in this case for a sample app for users it is better to use toast
- noted a weird behavior with cropping at the bottom

PR #4536, opened 2026-05-29
[Docs]: remove redundant .Gap(4) calls from docs and samples

### LINKED ISSUE #4528: (docs): usage of bad .Gap(), .Margin() and .Padding()
### Description

there is default values of 4 and setting gap(4) makes no sense = likely consistency with other projects

defymecobra:
## 📝 The Issue
In issue, the author pointed out redundant usage of spacing methods across the framework, samples, and documentation: `.Gap(4)`, `.Margin(4)`, and `.Padding(4)`. According to the issue description, the value `4` (medium size) is applied by default, making explicit declarations redundant and adding unnecessary visual noise to the codebase.

## 🎯 What Needed to Be Done
The goal was to analyze all `C#` / `XAML` source files, along with the source `Markdown` documentation files, find occurrences of these methods with the parameter `4`, remove them, and verify that the UI layout does not break.

## 🛠️ What Was Done & WHY
A mass replacement was performed, but with a critical correction to the logic proposed in the original issue:

1. **Only `.Gap(4)` calls were removed**
   - **Why:** We investigated the framework's source code (specifically `LayoutView.cs`). The `_rowGap` and `_columnGap` variables are indeed initialized to `4` by default. Therefore, calling `.Gap(4)` is entirely redundant. These calls were successfully removed across 82+ files.

2. **`.Margin(4)` and `.Padding(4)` were kept unchanged**
   - **Why:** The issue author's assumption that Margin and Padding also default to `4` turned out to be **incorrect**. 
   - Inspecting the framework's core components revealed that, for instance, in `LayoutView`, the default paddings and margins are `null` (i.e., 0). In the `Box` primitive, the default `Padding` is `2` and `Margin` is `0`. 
   - An initial attempt to remove `.Padding(4)` and `.Margin(4)` caused actual UI bugs (e.g., the padding of the search bar in the documentation completely broke in `SmartSearch.cs`).
   - Consequently, these calls are genuinely meaningful (they override the actual default zeros and twos to fours) and were preserved.

## ✅ Result
- The codebase is clean of meaningless `.Gap(4)` calls.
- There are no visual regressions in the UI, as all meaningful spacing values (Margin and Padding) were intentionally kept.
- The project builds without errors. The generated documentation updates automatically based on the corrected `.md` source files.


artem-ivy-ai:
### Staging removed

PR #4540, opened 2026-05-30
[TextInputVariant]: force foreground color on embedded buttons

### LINKED ISSUE #4533: Folder Select button should be styled like an attix
## Summary
The Folder Select button ("Browse" button) in the "Add a project" dialog should be styled like an attix component.

## Additional Context
The button appears in the "Add one or more Git repositories" section of the project creation dialog, next to the "Repository URL or Local Path" input field.

**Slack thread:** https://ivy-interactive.slack.com/archives/C09PMA599RC/p1780053049647959?thread_ts=1780053049.647959&cid=C09PMA599RC

**Reported by:** niels

---

defymecobra:
## Problem

Buttons placed inside `.Suffix()` / `.Prefix()` slots of a `TextInput` inherited the `text-muted-foreground` color from the affix container. This caused button labels and icons to appear gray/muted instead of using the standard foreground color, making them look disabled or de-emphasized — even though they are fully interactive.

This was especially noticeable for `.Ghost()` buttons, which have a transparent background and rely entirely on text color for visual presence.

## What needed to be done

Ensure that any button rendered inside an affix cell always uses the normal foreground color (`text-foreground`), regardless of the muted color applied to the surrounding container (which is intentional for non-interactive content like icons or unit labels).

## What was changed and why

Added `[&_button]:text-foreground` to the `affixEmbeddedButtonClasses` constant in `text-input-variant.ts`.

This CSS selector targets any `<button>` element nested inside the affix cell and overrides the inherited `text-muted-foreground` with `text-foreground`. The override is scoped specifically to buttons, so non-button affix content (e.g. plain text labels, icons used as indicators) still renders in the muted color as designed.

### Changed file

- `src/frontend/src/components/ui/input/text-input-variant.ts`

Before:
image

After:
image

### DIFF
--- a/src/frontend/src/components/ui/input/text-input-variant.ts
+++ b/src/frontend/src/components/ui/input/text-input-variant.ts
@@ -8,7 +8,7 @@ import { cn } from "@/lib/utils";
  * `size-7`/`size-9` target is larger than the glyph, which reads as extra padding.
  */
 export const affixEmbeddedButtonClasses =
-  "[&_button]:!px-0 [&_button]:shadow-none [&_button]:rounded [&_button]:hover:bg-accent [&_button]:cursor-pointer [&_button]:transition-colors [&_button.size-7]:!size-4 [&_button.size-9]:!size-6";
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4546, opened 2026-05-31
[DataTable]: fix grow column widths distribution and container flexbox shrinking

### LINKED ISSUE #843: Dashboard columns are no longer divided
## Summary
The dashboard columns are no longer properly divided/separated. Based on the screenshot provided, the dashboard layout appears to have lost its column structure.

**Expected:** Dashboard metrics (Total Plans, Drafts, In Progress, Ready for Review, Completed, Failed, Avg Cost/Plan) should be displayed in properly divided columns.

**Actual:** Columns are no longer divided as expected.

## Additional Context
Screenshot shows the dashboard at `localhost:5010/dashboard` with metric cards and a data table below. The visual layout suggests the column dividers or spacing may have been affected by a recent change.

---

[Slack thread](https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1779608925533909?thread_ts=1779608925.533909&cid=C0AS51ABBB6)

Reported by niels

---

defymecobra:
## Overview
This PR addresses critical rendering bugs in the `DataTableWidget` where proportional columns (`Fraction` / `Grow`) failed to distribute space equally, and the table's height failed to shrink correctly inside Flexbox layouts.

## Issue 1: Uneven Column Width Distribution
**Problem:** 
Columns explicitly set to equal fractions (e.g., `Size.Fraction(1/8f)`) were rendered with slightly varying widths. In some scenarios, only the very last column would stretch to fill the remaining space.

**Root Causes:**
1. The `syncColumns` hook inadvertently wiped out the `originalWidth` metadata (which contains the raw `Fraction:0.125` string) by overwriting it with `columnsProp`. This caused `parseSizeGrow` to return `undefined`, forcing the table to apply the default fallback where only the last column gets `grow: 1`.
2. `glide-data-grid` calculates final column widths by first allocating a base width, then dividing the *remaining* space according to `grow` factors. Because base widths were derived from header text lengths, columns with longer text got a larger base width. Distributing equal grow factors over unequal base widths resulted in uneven final widths.

**The Fix:**
1. Modified `parseSizeGrow` in `columnHelpers.ts` to fall back to `col.width` (as a string) when `col.originalWidth` is missing. This ensures the grow factor is always preserved.
2. Implemented a two-pass initialization in `useColumnManagement.ts`. It now identifies all proportional columns, finds the maximum base width among them, and normalizes all of them to share that same base width. This forces `glide-data-grid` to distribute the remaining space perfectly evenly.

## Issue 2: Improper Flexbox Shrinking
**Problem:** 
When a DataTable had an explicitly defined height (e.g., `307px`), it would not shrink properly if its flex parent (like `Layout.Vertical()`) ran out of vertical space. 

**Root Cause:** 
The explicit `height` property created a rigid constraint. Previous workarounds attempting to use `maxHeight: 100%` and `min(height, 100%)` were brittle and failed to leverage native flexbox mechanics.

**The Fix:** 
Updated `DataTableWidget.tsx` to move explicit heights into the `flexBasis` property instead of `height`, while keeping `flexShrink: 1`. This correctly communicates the table's ideal size to the flex engine while allowing it to gracefully shrink when the parent container lacks space.

Fixes [Issue 843](https://github.com/Ivy-Interactive/Ivy-Tendril/issues/843)

Before:
image

After:
image

### DIFF
--- a/src/frontend/src/widgets/dataTables/DataTableWidget.test.ts
+++ b/src/frontend/src/widgets/dataTables/DataTableWidget.test.ts
@@ -18,8 +18,9 @@ describe("DataTableWidget - container style for height modes", () => {
     expect(source).toContain("containerStyle.flexGrow = 1");
   });
 
-  it("should cap explicit heights with min(preferred, 100%) so tables shrink in tabs", () => {
-    expect(source).toContain("containerStyle.height = `min(${containerStyle.height}, 100%)`");
+  it("should move explicit heights to flex-basis so tables shrink in tabs", () => {
+    expect(source).toContain("containerStyle.flexBasis = containerStyle.height");
+    expect(source).toContain("delete containerStyle.height");
   });
 
   it("should compute a density-aware minimum height before page scroll", () => {
--- a/src/frontend/src/widgets/dataTables/DataTableWidget.tsx
+++ b/src/frontend/src/widgets/dataTables/DataTableWidget.tsx
@@ -116,9 +116,9 @@ export const DataTable: React.FC<DataTableWidgetProps> = ({
     containerStyle.flexDirection = "column";
     containerStyle.flexShrink = 1;
     containerStyle.minHeight = minHeight;
-    containerStyle.maxHeight = "100%";
     if (containerStyle.height) {
-      containerStyle.height = `min(${containerStyle.height}, 100%)`;
+      containerStyle.flexBasis = containerStyle.height;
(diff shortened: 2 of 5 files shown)


defymecobra:
Also updated a test

rorychatt (approved):

artem-ivy-ai:
### Staging removed

PR #4569, opened 2026-06-04
[00086] Checkbox labels should be clickable to toggle state

nielsbosma:
# Summary

## Changes

Updated the `BoolInputWidget` component to make labels and descriptions clickable for toggling checkbox, switch, and toggle inputs. Clicking the label or description text now toggles the input state, bringing the behavior in line with standard web UX.

The implementation correctly handles:
- Nullable checkbox cycle (null → true → false → null)
- Disabled and loading states (prevents clicks)
- Appropriate cursor styling (pointer when enabled, not-allowed when disabled)
- All three variants (Checkbox, Switch, Toggle)

## API Changes

None — this is a purely behavioral change. The `InputLabel` component is an internal implementation detail, not a public API. No exported interfaces or props were modified.

## Files Modified

- **src/frontend/src/widgets/inputs/BoolInputWidget.tsx**
  - Enhanced internal `InputLabel` component with `onClick` and `disabled` props
  - Added conditional cursor styling (`cursor-pointer` / `cursor-not-allowed`)
  - Implemented `handleLabelClick` in all three variants (Checkbox, Switch, Toggle)
  - Checkbox variant correctly replicates the nullable cycle logic from checkbox.tsx

---

## Commits

- 8783dbeca: Fix formatting in BoolInputWidget
- 515561105: Make checkbox, switch, and toggle labels clickable to toggle state

### DIFF
--- a/src/frontend/src/widgets/inputs/BoolInputWidget.tsx
+++ b/src/frontend/src/widgets/inputs/BoolInputWidget.tsx
@@ -73,11 +73,19 @@ const InputLabel: React.FC<{
   label?: string;
   description?: string;
   density?: Densities;
-}> = React.memo(({ id, label, description, density = Densities.Medium }) => {
+  onClick?: () => void;
+  disabled?: boolean;
+}> = React.memo(({ id, label, description, density = Densities.Medium, onClick, disabled }) => {
   if (!label && !description) return null;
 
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4582, opened 2026-06-09
fix(ActivityHeatmap): dimension label localization and style adjustments for cell size change

phonofidelic:
## Summary

- Replace the hardcoded `MONTH_NAMES` array and manual `"Month Day, Year"` template with locale-aware `Intl`/`toLocaleDateString` formatting, so tooltip headers and column labels respect the user's locale:
**Before**
ActivityHeatmap sv-SE tooltip header - before

**After**
ActivityHeatmap sv-SE tooltip header - after

- Fix a timezone off-by-one bug: tooltip header dates are now parsed with an explicit local-time component (`+ "T00:00:00"`) so single-day entries no longer render as the previous day for users in negative-UTC timezones.
- Consolidate label formatting into a single `formatDimensionLabel` helper that drives both daily (month) and hourly (month + day) x-axis labels, and add `text-nowrap` so wider labels don't wrap inside the fixed-width column cell.
- Add unit tests covering `formatTooltipHeader` (daily, hourly, hour 0/23 edge cases, and null/undefined hour handling).
- Adjust frontend component styles after cell size increase:

**Before:**
ActivityHeatmap size adjustment - before

**After:**
ActivityHeatmap size adjustment - after

### DIFF
--- a/src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
+++ b/src/widgets/Ivy.Widgets.ActivityHeatmap/frontend/src/ActivityHeatmap.tsx
@@ -2,7 +2,7 @@ import { useEffect, useRef, useState } from "react";
 import "./style.css";
 import { ActivityHeatmapProps, Activity } from "./types";
 import { computeMaxCount, getLevel } from "./levelUtils";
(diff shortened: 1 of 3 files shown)


phonofidelic:
@claude review this

PR #4591, opened 2026-06-12
fix: stop root connection reconnect storm when navigating with appArgs

nielsbosma:
The root (shell) SignalR connection used raw appArgs in its connection URL
and effect deps, so navigating to an app carrying appArgs (e.g. the agent
app via "Continue with Agent") tore down and rebuilt the live root
connection mid-handshake, producing a 'WebSocket is not in the OPEN state'
reconnect loop and a stuck 'Connection lost' modal.

- use-backend: add stableAppArgs gated by syncStableAppValues (mirroring
  stableAppId/stableAppShell) so the root no longer reconnects when only
  appArgs change; child app-hosts still receive args immediately.
- AppShell: add AppShellSettings.IncludeArgsInUrl (default true) with
  HideArgsInUrl()/IncludeArgsInUrl(bool) and a GetUrl(includeArgs) overload
  to keep navigation args out of the browser URL; DefaultSidebarAppShell
  honors it.

### DIFF
--- a/src/Ivy/AppShell/DefaultSidebarAppShell.cs
+++ b/src/Ivy/AppShell/DefaultSidebarAppShell.cs
@@ -140,7 +140,7 @@ public class DefaultSidebarAppShell(AppShellSettings settings) : ViewBase
         void RedirectToAppIfNotError(NavigateArgs navigateArgs, bool replaceHistory = false, string? tabId = null)
         {
             if (IsErrorApp(navigateArgs.AppId)) return;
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-framework-staging-docs-4591.sliplane.app](https://ivy-framework-staging-docs-4591.sliplane.app)
**🧪 Samples:** _not available_

PR #4592, opened 2026-06-12
Fix xterm rendering to match Windows Terminal

nielsbosma:
- Switch to the canvas renderer (@xterm/addon-canvas) so box-drawing and
  block-element glyphs (e.g. the Claude Code logo) tile seamlessly via
  customGlyphs instead of relying on font metrics.
- Clear the texture atlas once document.fonts are ready to avoid tofu from
  the webfont-load race.
- Add a doFit() correction so the canvas renderer's per-row pixel rounding
  no longer clips the last line in full-height terminals.


artem-ivy-ai:
### Staging removed

PR #4593, opened 2026-06-12
Add loading overlay to Xterm Terminal widget

nielsbosma:
New Loading/LoadingText props show a spinner overlay until the first
visible output arrives on the PTY stream. ANSI escape preambles from
ConPTY (title set, cursor homing) are filtered out so the overlay
stays up until the process actually prints something.


artem-ivy-ai:
### Staging removed

PR #4581, opened 2026-06-09
Add CollapsibleOnMobile to SidebarLayout + stable mobile tab-header height

dcrjodle:
## Summary

Two framework changes that enable responsive (mobile/tablet) layouts for apps built on Ivy. Motivated by making Ivy Tendril usable on phones/tablets, but both are generic framework improvements.

### 1. `SidebarLayout.CollapsibleOnMobile`
A new opt-in (`[Prop] CollapsibleOnMobile` + `.CollapsibleOnMobile()` extension) that lets a **nested, non-main** sidebar auto-collapse below the mobile breakpoint, reusing the same media-query logic that already powers the main app sidebar.

- Frontend gates the auto-collapse + initial-state logic behind a new `collapseEnabled = mainAppSidebar || collapsibleOnMobile` flag.
- The **Ctrl+B shortcut** and the **in-pane toggle button** remain exclusive to `mainAppSidebar`, so a nested master-detail layout collapses its list on small screens but does not sprout app-chrome controls. The hosting app is expected to provide its own navigation when collapsed (e.g. a header dropdown).
- Fully backward compatible: defaults to `false`, no behavior change for existing layouts.

### 2. Stable tab-header height on mobile
`TabsVariant` header now reserves a minimum height (`min-h-11`) and bottom-aligns its triggers. Previously the header sized only to its content, so when there were no tabs the header shrank and an absolutely-positioned sidebar toggle overflowed the header's bottom border. The `pl-10` slot already reserved horizontal room for the toggle; this reserves the vertical room too.

## Testing
- `dotnet build src/Ivy/Ivy.csproj` — clean (0 warnings, 0 errors)
- `tsc -b` (frontend typecheck) — clean
- `vp lint` on changed `.tsx` — 0 warnings, 0 errors
- Manual verification at 390px pending (consumer-side, in Ivy Tendril)

## Notes
The consuming changes live in a separate Ivy-Tendril branch (nested master-detail apps adopt `.CollapsibleOnMobile()` and render a mobile header-title dropdown for navigation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4604, opened 2026-06-16
Fix dropdown trigger forcing full width and wrapping in toolbars

dcrjodle:
## Problem

In Ivy Tendril, the action-buttons footer in the **draft** and **review** views has a `...` (ellipsis) "more actions" button rendered via `Button().WithDropDown(...)`. That dropdown trigger was stretching to **full width**, pushing it onto its own line inside the `Layout.Horizontal().Wrap()` footer.

## Root cause

The trigger isn't styled in Tendril — the full width is forced in the framework. In #4581 (CollapsibleOnMobile / mobile-responsive work), the `DropDownMenu` trigger wrapper was changed:

```diff
- <div>{slots.Trigger}</div>
+ <div style={{ width: "100%", minWidth: 0 }}>{slots.Trigger}</div>
```

That was meant so a header dropdown could fill a sidebar drawer on mobile, but it applied to **every** dropdown trigger, making standalone toolbar dropdown buttons stretch full width and wrap.

## Fix

Use `maxWidth: 100%` instead of `width: 100%` so the trigger sizes to its content by default (no spurious wrapping) while staying constrained to its container. Consumers that genuinely want a full-width trigger can set the width on the trigger button itself. `minWidth: 0` is kept for flex-shrink safety.

```diff
- <div style={{ width: "100%", minWidth: 0 }}>{slots.Trigger}</div>
+ <div style={{ maxWidth: "100%", minWidth: 0 }}>{slots.Trigger}</div>
```

## Verification

- `npx tsc --noEmit` passes.
- The fix is framework-side; the affected Tendril call sites (`Apps/Drafts/ActionBarView.cs`, `Apps/Review/ContentView.cs`) need no changes.

### DIFF
--- a/src/frontend/src/widgets/dropDownMenu/DropDownMenuWidget.tsx
+++ b/src/frontend/src/widgets/dropDownMenu/DropDownMenuWidget.tsx
@@ -208,7 +208,7 @@ export const DropDownMenuWidget: React.FC<DropDownMenuWidgetProps> = ({
   return (
     <DropdownMenu open={open} onOpenChange={handleOpenChange}>
       <DropdownMenuTrigger ref={triggerRef} asChild>
-        <div style={{ width: "100%", minWidth: 0 }}>{slots.Trigger}</div>
+        <div style={{ maxWidth: "100%", minWidth: 0 }}>{slots.Trigger}</div>
       </DropdownMenuTrigger>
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4605, opened 2026-06-16
fix(sheet): preserve responsive breakpoints when Resizable()

dcrjodle:
## Why is the fix specific to the Sheet widget?

Responsive width resolution is centralized and works correctly for every widget: `widgetRenderer.tsx` resolves `props.width` to the active breakpoint before the widget renders. Sheet is the **only** widget that rewrites its own `Width` in C# inside an extension method (`Resizable()`), upstream of that resolver — so it's the only place a `Responsive<Size>` can be silently flattened before it ever reaches the frontend.

(The narrower limitation in the `Responsive<T>` `.At`/`.And` API that makes this easy to hit is tracked as a follow-up.)

## Problem

Resizable sheets with a multi-breakpoint responsive width collapse to a single fixed width (~384px) and overflow on larger screens.

`Resizable()` injects the resize min/max by reading **only** `Width?.Default`:

```csharp
var width = sheet.Width?.Default ?? Sheet.DefaultWidth;
```

A `Responsive<Size>` that sets only `Mobile`/`Tablet`/`Desktop`/`Wide` (with `Default == null`) resolves to `Sheet.DefaultWidth` (`Size.Rem(24)` ≈ 384px), re-wrapped as a flat, non-responsive width — discarding every breakpoint rule. (`Resizable()` rewrites the width because the resizable frontend reads its initial/min/max from the serialized `"value,min,max"` string.)

## Fix

Apply the default min/max to **every populated breakpoint** (`ApplyConstraints`), preserving the responsive ramp. Keeps each breakpoint's explicit `Min`/`Max` when set, falls back to a single constrained default only when no breakpoint is set, and applies the same handling to the height axis for top/bottom sheets.

## Tests

New `SheetResizableTests` (4 cases) — the first reproduces the failing ramp; all would fail under the old `.Default`-only behavior. Build clean, 47/47 Sheet+Responsive tests pass, `dotnet format` clean.

### DIFF
--- a/src/Ivy/Widgets/Sheet.cs
+++ b/src/Ivy/Widgets/Sheet.cs
@@ -91,30 +91,53 @@ public static class SheetExtensions
 
         if (isHorizontal)
         {
-            var width = sheet.Width?.Default ?? Sheet.DefaultWidth;
-            if (width.Min == null)
-            {
-                width = width.Min(Size.Px(200));
-            }
-            if (width.Max == null)
-            {
-                width = width.Max(Size.Px(1200));
-            }
+            var width = ApplyConstraints(sheet.Width, Sheet.DefaultWidth, Size.Px(200), Size.Px(1200));
             return (sheet with { Resizable = true }).Width(width);
         }
         else
         {
-            var height = sheet.Height?.Default ?? Sheet.DefaultHeight;
-            if (height.Min == null)
-            {
-                height = height.Min(Size.Px(100));
-            }
-            if (height.Max == null)
-            {
-                height = height.Max(Size.Px(900));
-            }
+            var height = ApplyConstraints(sheet.Height, Sheet.DefaultHeight, Size.Px(100), Size.Px(900));
             return (sheet with { Resizable = true }).Height(height);
         }
+    }
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Ivy/Widgets/Sheet.cs
```diff
@@ -91,32 +91,55 @@ public static Sheet Resizable(this Sheet sheet, bool resizable = true)
...
-            var height = sheet.Height?.Default ?? Sheet.DefaultHeight;
-            if (height.Min == null)
-            {
-                height = height.Min(Size.Px(100));
-            }
-            if (height.Max == null)
-            {
-                height = height.Max(Size.Px(900));
-            }
+            var height = ApplyConstraints(sheet.Height, Sheet.DefaultHeight, Size.Px(100), Size.Px(900));
```
rorychatt, line 99: maybe this could and should have been hard-coded as a default value in the frontend (like some other default values in frontend for other widgets)

rorychatt (commented):
Left minor feedback

-> I see that some other edge cases might brake, but time will tell.

The comment is about an already existing bad pattern, so IMO this is merge-able, but maybe we should fix/open issue about the default value coded in BE, not in FE

artem-ivy-ai:
### Staging removed

PR #4607, opened 2026-06-17
Make breakpoint detection container-aware so responsive layouts respect the sidebar

dcrjodle:
## Problem

Responsive resolution (`responsiveVisible`, `width`/`height`/`density`, and any `.ShowOn()`/`.HideOn()` collapse logic) was driven by `useBreakpoint()`, which measured **`window.innerWidth`** — the whole viewport.

When the main app sidebar is **expanded**, the content area shrinks: `SidebarLayoutWidget` lays the app out as a CSS grid with `gridTemplateColumns: ${sidebarWidth} 1fr`, so the `1fr` content track loses ~the sidebar's width. But the viewport doesn't change, so the breakpoint stayed at `wide` even though the content was much narrower. Progressive-collapse action bars (e.g. Drafts/Recommendations/Review in Ivy-Tendril) therefore kept all buttons inline and overflowed instead of collapsing into the `⋮` dropdown.

With the sidebar **collapsed**, content width ≈ viewport, so the math happened to line up and it worked. That's the reported "works collapsed, broken expanded" symptom.

Surfaced by **Ivy-Tendril#1208** (fixes Ivy-Tendril#724). The Tendril-side progressive-collapse logic was already correct — the defect was here, in how the breakpoint is measured.

## Fix

- **`use-responsive.ts`** — `useBreakpoint(containerRef?)` now optionally derives the breakpoint from a container element's `clientWidth` via `ResizeObserver`, re-evaluating on sidebar open/close and resize-drag. Falls back to `window.innerWidth` when no ref is supplied (or the element hasn't measured yet), so existing window-based callers are unchanged.
- **`use-breakpoint-context.tsx`** — `BreakpointProvider` accepts an optional `containerRef` and forwards it.
- **`AppHostWidget.tsx`** — passes its content-container ref to the provider, making **per-app** responsive resolution sidebar-aware. The global `App.tsx` provider stays window-based for chrome outside any app host.
- Clarifying comment on the misleading `BREAKPOINTS.wide = 1280` constant: it is **not** used as a threshold (`wide` resolves at `>=1024`); value left unchanged to avoid an app-wide behavior shift.

## Tests

New `use-responsive.test.tsx` covers:
- window fallback when no ref is given,
- breakpoint derived from container width (narrow container under a wide viewport → narrower band),
- re-evaluation when the observed container changes width.

## Verification

- `tsc -b` ✓, lint ✓
- All hook/responsive/breakpoint suites pass (49 tests)
- Full production frontend build ✓ (embedded bundle confirmed to contain the new `ResizeObserver` path)
- `dotnet build` of Ivy-Tendril against this framework ✓ (0 errors), fresh frontend embed confirmed

### DIFF
--- a/src/frontend/src/hooks/__tests__/use-responsive.test.tsx
+++ b/src/frontend/src/hooks/__tests__/use-responsive.test.tsx
@@ -1,0 +1,139 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import React, { act, useRef } from "react";
+import { createRoot, Root } from "react-dom/client";
+import { useBreakpoint, type BreakpointName } from "../use-responsive";
+
+let container: HTMLDivElement;
+let root: Root;
+
+function mount(element: React.ReactElement) {
+  act(() => {
+    root.render(element);
+  });
+}
+
+beforeEach(() => {
+  container = document.createElement("div");
+  document.body.appendChild(container);
+  root = createRoot(container);
+});
+
+afterEach(() => {
+  act(() => {
+    root.unmount();
+  });
+  container.remove();
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+});
+
+function setViewportWidth(width: number) {
+  Object.defineProperty(window, "innerWidth", { value: width, configurable: true, writable: true });
+}
+
+/**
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #4610, opened 2026-06-17
Show fields with an explicit Builder() in DetailsBuilder

nielsbosma:
## Problem

Collection-typed fields (e.g. `List<Link>`) are auto-removed during `ToDetails()` scaffolding — `_Scaffold` sets `IsRemoved = true` for any collection type. Supplying a custom `.Builder(...)` for such a field only swapped the builder and left `IsRemoved` set, so `Build()` filtered it out and it never rendered.

This surfaced in Ivy.Tendril's `DetailsTabView`, where `RelatedPlans` and `DependsOn` (both `List<Link>`) were never shown despite having explicit builders.

## Fix

Reset `IsRemoved = false` in all four `Builder(...)` overloads. Explicitly giving a field a builder now re-includes it, so collection fields render as intended. No change needed in downstream callers.

### DIFF
--- a/src/Ivy/Views/Builders/DetailsBuilder.cs
+++ b/src/Ivy/Views/Builders/DetailsBuilder.cs
@@ -146,6 +146,7 @@ public class DetailsBuilder<TModel> : ViewBase, IStateless
     {
         var column = GetField(field);
         column.Builder = builder(_builderFactory);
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4626, opened 2026-06-22
[00549] Text widget: thread id prop through all non-heading variants

### LINKED ISSUE #4623: Text widget: Anchor id is dropped by all non-heading variants (breaks #fragment scroll)
## Summary

`Text.Block(...).Anchor("foo")` (and any non-heading `Text` variant with an anchor) renders **no `id`** on the DOM element, so `document.getElementById("foo")` returns `null` and anchor/`#fragment` scrolling silently does nothing.

## Root cause

In `src/frontend/src/widgets/primitives/TextBlockWidget.tsx`, the `anchor` prop is passed to the variant component as `id={anchor}` (line ~238), but **only the `H1`–`H6` variants actually destructure and apply `id`**. All 14 other variants — `Literal`, `Block`, `P`, `Inline`, `Blockquote`, `Monospaced`, `Lead`, `Muted`, `Danger`, `Warning`, `Success`, `Label`, `Strong`, `Display` — have the signature `({ children, className, style }) => ...` and never render the `id`.

So for non-heading variants the id is dropped on the floor.

## Impact / repro

In Ivy-Tendril's Review → Changes tab (`ChangesTabView.cs`), each file diff is preceded by `Text.Block("").Anchor(path)` and the file tree calls `client.Redirect($"#{path}")` on select. Because the `Block` variant drops the id:
1. `Redirect("#<path>")` → frontend `extractAnchorId` → `document.getElementById("<path>")` → **null**
2. `scrollIntoView` is never called → clicking a file in the tree does not scroll to its diff.

(That code path predates the recent mobile-dropdown work; this is not a regression from PR #1294.)

## Suggested fix

Make every variant honor the `id` it's already being given (the `VariantMap` type already declares `id?: string`). Either:
- thread `id` into each variant's rendered element, or
- wrap the rendered component in an anchor element that carries the `id` when `anchor` is set (headings already self-apply `id`, so avoid emitting a duplicate for those).

## Affected file

- `src/frontend/src/widgets/primitives/TextBlockWidget.tsx`

dcrjodle:
Fixes #4623

# Summary

## Changes

Added the `id` prop to all 14 non-heading text widget variants (Literal, Block, P, Inline, Blockquote, Monospaced, Lead, Muted, Danger, Warning, Success, Label, Strong, Display) so that `Text.Block("...").Anchor("foo")` correctly renders the `id` attribute on the DOM element. Added comprehensive test coverage for id attribute rendering across all variants.

## API Changes

All non-heading variant components now accept and render the `id` prop:
- `Literal`, `P`, `Inline`, `Blockquote`, `Monospaced`, `Lead`, `Muted`, `Danger`, `Warning`, `Success`, `Label`, `Strong`, `Display`: Added `id` parameter to component signature and applied to rendered element
- `Block`: Added `id` parameter and applied to outer `<div>` wrapper

## Files Modified

- `src/frontend/src/widgets/primitives/TextBlockWidget.tsx` — Added `id` prop to 14 variant component signatures and rendered elements
- `src/frontend/src/widgets/primitives/TextBlockWidget.test.tsx` — New test file with comprehensive coverage for id attribute rendering

---

## Commits

- 316529925
- b41006c4f8d74e8c651a5d72c8cebe06083d59f7

### DIFF
--- a/src/frontend/src/widgets/primitives/TextBlockWidget.test.tsx
+++ b/src/frontend/src/widgets/primitives/TextBlockWidget.test.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import { render } from "@testing-library/react";
+import { describe, it, expect } from "vitest";
+import { renderToString } from "react-dom/server";
 import { TextBlockWidget } from "./TextBlockWidget";
 
 describe("TextBlockWidget id attribute", () => {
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #4635, opened 2026-06-25
Fix plugin reload loop causing OOM on persistent failures

zachwolfe:
## Summary
- Uncaught exceptions (e.g. `TypeLoadException`) in `PluginReloadScheduler` would bypass the cooldown, causing infinite retry loops that exhausted memory
- Added exponential backoff and a max consecutive failure limit (5 attempts), after which reload attempts are suspended
- Failure counter resets when a source file changes, allowing retries after the developer fixes the issue


zachwolfe:
I'm not at all confident in this fix. I will investigate more thoroughly later.

artem-ivy-ai:
### Staging removed

PR #4650, opened 2026-06-25
fix: code input copy button position and hover visibility

rorychatt:
This PR fixes the layout and hover behavior of the copy code button in CodeInputWidget:

1. Moves the copy button to the top right using textareaTrailingOverlayClasses.
2. Shows the copy button only on hover over the input container (using group-hover).
3. Moves scroll padding to the CodeMirror content element so the scrollbar remains at the far right edge of the container.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

PR #4652, opened 2026-06-25
[00595] Fix inline code rendering in bullet points

### LINKED ISSUE #4640: Markdown rendering issue with inline code in bullet points
## Summary
Markdown rendering is not working correctly when inline code is placed within bullet points.

## Additional Context
This issue needs to be fixed in both the main codebase and in Tendrils DraftMarkdown.

The screenshot shows a "Local Changes Detected" dialog where inline code formatting (e.g., `origin/main`) appears within bullet point lists and is not rendering properly.

## Related
- Slack thread: https://ivy-interactive.slack.com/archives/C09PMA599RC/p1782377926899559?thread_ts=1782377926.899559&cid=C09PMA599RC

## Reported by
Reported by @nielsbosma

---

dcrjodle:
Fixes #4640

## Problem

Inline code (backtick text) inside bullet points renders as a full-width block element instead of inline with the surrounding text. In the "Local Changes Detected" dialog, text like "4 commits ahead of `origin/main`" renders with `origin/main` on its own line spanning the full width.

The root cause is in the `li` component rendering logic in MarkdownRenderer.tsx. When a list item contains block-level children, all children are wrapped in a `flex flex-col` container, causing inline content to stack vertically.

## Solution

**Ivy-Framework:** Changed the `liContent` typography class in `styles.ts` from `"flex flex-col gap-y-1 [&>p]:my-0"` to `"[&>*+*]:mt-1 [&>p]:my-0"`. This removes the flex column layout, allowing inline content to flow naturally while maintaining spacing between block-level children.

---

## Commits

- 96c8918b3 Fix inline code rendering in list items

### DIFF
--- a/src/frontend/src/lib/styles.ts
+++ b/src/frontend/src/lib/styles.ts
@@ -771,7 +771,7 @@ export const typography: Record<string, string> = {
   ul: "list-disc list-outside ps-5 flex flex-col gap-y-1",
   ol: "list-decimal list-outside ps-5 flex flex-col gap-y-1",
   li: "list-item",
-  liContent: "flex flex-col gap-y-1 [&>p]:my-0",
+  liContent: "[&>*+*]:mt-1 [&>p]:my-0",
 
   // Links
   a: "text-primary underline underline-offset-[3px] brightness-90 hover:brightness-100",


artem-ivy-ai:
### Staging removed

PR #4654, opened 2026-06-25
feat(Xterm): add AutoFocus prop to Terminal

nielsbosma:
Adds an `AutoFocus` prop (default `true`) to the Terminal widget so the xterm automatically receives keyboard input on mount.

## Changes
- **C# (`Terminal.cs`)**: new `[Prop] bool AutoFocus = true` plus a fluent `.AutoFocus(bool = true)` extension.
- **React (`Terminal.tsx`)**: `autoFocus` prop (default `true` to match the C# default since defaults aren't serialized); calls `term.focus()` once the terminal is opened and ready, gated by `autoFocus && !isReadOnly` so read-only/closed terminals are unaffected.
- **README**: documented the new prop.

Frontend bundle rebuilds successfully.


artem-ivy-ai:
### Staging removed

PR #4656, opened 2026-06-25
fix(Table): shrink columns to content when all are content-sized

### LINKED ISSUE #1378: Columns not shrinking to content
## Summary
Columns should shrink to fit their content instead of expanding to fill available space.

## Additional Context
As shown in the attached screenshot, the table columns are not properly sizing themselves to their content. The "Status" and "Name" columns should be narrower and only take up the space needed for their content.

**Slack thread:** https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1782379271826209?thread_ts=1782379271.826209&cid=C0AS51ABBB6

---

dcrjodle:
## Problem

The simple `Table` widget stretches its columns to fill the container instead of shrinking them to their content. In the [reported case](https://github.com/Ivy-Interactive/Ivy-Tendril/issues/1378) a `Status` (badge) and `Name` (link) column each balloon to fill the available width, leaving large empty gaps.

## Root cause

Auto-scaffolded tables render at `Size.Full()` with `table-layout: auto`. That combination sizes columns to content **and then redistributes all leftover space back into them to reach 100% width**, so columns stretch. The per-header `Size.Fit()` width is only a minimum hint under `table-layout: auto` and has no effect against a 100%-wide table.

## Fix (opt-in, default unchanged)

A table now renders at `fit-content` width **only when every visible column is content-sized** (`Size.Fit`/`Auto`/`MinContent`/`MaxContent`). With no slack to redistribute, columns collapse to their content. If any column is not content-sized, or an explicit table `Width(...)` is set, behavior is unchanged (full width).

Opt in per-column via the existing API — no new public surface:

```csharp
items.ToTable()
     .ColumnWidth(x => x.Status, Size.Fit())
     .ColumnWidth(x => x.Name, Size.Fit())
```

## Changes

- **`TableBuilder.cs`** — when all visible columns are content-sized (and no explicit table width), use `Size.Fit()` instead of `Size.Full()`.
- **`TableWidget.tsx`** — when width is `Fit`, apply `w-fit` to the wrapper and `<table>` so the `fit-content` width actually takes effect (overrides the base component's hardcoded `w-full` via tailwind-merge).
- **`TableWidget.test.tsx`** (new) — covers default full-width, the `Fit` shrink path, and `auto`/`fixed` table-layout.

## Verification

- `dotnet build src/Ivy/Ivy.csproj` — 0 warnings/errors
- frontend `tsc --noEmit` — clean
- `TableWidget` tests — 3/3 pass

Fixes Ivy-Interactive/Ivy-Tendril#1378

### DIFF
--- a/src/Ivy/Views/Tables/TableBuilder.cs
+++ b/src/Ivy/Views/Tables/TableBuilder.cs
@@ -384,11 +384,15 @@ public class TableBuilder<TModel> : ViewBase, IStateless
 
         Table RenderTable(TableRow[] tableRows)
         {
-            var tableWidth = _width ?? Size.Full();
+            var visibleColumns = _columns.Values.Where(e => !e.Removed).ToList();
+
+            bool allColumnsContentSized = visibleColumns.Count > 0 &&
+                visibleColumns.All(c => c.Width != null && c.Width.IsContentSized);
+
+            var tableWidth = _width ?? (allColumnsContentSized ? Size.Fit() : Size.Full());
             var table = new Table(tableRows).Width(tableWidth).Density(_density);
 
-            bool hasContentSizedColumn = _columns.Values
-                .Where(e => !e.Removed)
+            bool hasContentSizedColumn = visibleColumns
                 .Any(c => c.Width == null || c.Width.IsContentSized);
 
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #4664, opened 2026-06-26
Fix plugin loader rejecting plugins with both Debug and Release builds

zachwolfe:
Deduplicate DLLs by filename when scanning plugin directories, preferring Debug over Release. Fixes source plugins being rejected with "Multiple [assembly: IvyPlugin] attributes found" when both build configurations exist.


artem-ivy-ai:
### Staging removed

PR #4665, opened 2026-06-26
Fix plugin loading for NuGet-referenced shared assemblies

zachwolfe:
Plugins referencing shared assemblies via NuGet failed to load on a locally-built host with `FileNotFoundException` because returning `null` from the plugin ALC didn't reliably resolve through the Default context during native type resolution. Explicitly load shared assemblies by name from `AssemblyLoadContext.Default` instead.


artem-ivy-ai:
### Staging removed

PR #4666, opened 2026-06-27
Fix duplicate plugin loading during discovery

zachwolfe:
Plugins that exist both as a subdirectory in the plugins folder and in `plugin-references.yaml` were being loaded twice, causing duplicate registrations. This deduplicates by manifest ID and gives explicit references priority over auto-discovered subdirectories.


artem-ivy-ai:
### Staging removed

PR #4673, opened 2026-06-29
Fix localhost connection error in webview on macOS/Linux

rorychatt:
Default useTls to false on macOS/Linux in DesktopWindow and Server to avoid connection errors on platforms where WKWebView/WebKitGTK cannot bypass self-signed certificate validation errors.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

zachwolfe:
@rorychatt This change breaks auth by default on macOS and Linux. Also on macOS, in my experience, trusting the dotnet dev-certs made Rustino work with local HTTPS Ivy apps. And I think I put a command to do that in a Tendril install script a while ago. So are you sure we still need this?

(admittedly, last time I looked into this, the situation on Linux seemed more complicated than on macOS)

PR #4679, opened 2026-06-30
Plugin system fixes

zachwolfe:
## Summary
- Detect source plugins with only a .csproj (no .cs files required)
- Don't set reload cooldown when plugin load fails
- Clean up failed plugins when their directory is deleted


artem-ivy-ai:
### Staging removed

PR #4686, opened 2026-07-01
[00630] Fix Text Input Suffix Button Clipping

dcrjodle:
When a button (e.g., "Browse") is placed in the suffix slot of a TextInput, it gets visually clipped at the edges. This is visible in the ProjectRepoPickerView where a "Browse" button is used as a suffix on the repository URL input field.

The root cause is in `textInputFieldShellClasses` which applies `overflow-hidden` on the field shell container. This clips the suffix/prefix content at the container's rounded corners.

## Solution

Replace `overflow-hidden` with `overflow-clip` on the field shell, and add proper border radius to first prefix and last suffix cells so button backgrounds fill correctly to the edge.

---

**Commits:**
- c7c0da21c Fix TextInput suffix button clipping
- c0dc45cb6 Add vertical padding to TextInput field shell

---

### DIFF
--- a/src/frontend/src/components/ui/input/text-input-variant.ts
+++ b/src/frontend/src/components/ui/input/text-input-variant.ts
@@ -71,8 +71,8 @@ export const textInputAffixCellVariant = cva(
   {
     variants: {
       side: {
-        prefix: "rounded-none",
-        suffix: "rounded-none",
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4696, opened 2026-07-02
Add optional Tools slot to Field widget

nielsbosma:
Adds an optional `Tools` slot to the `Field` widget so an arbitrary widget can be rendered alongside the label.

- **Backend** (`Field.cs`): new `[Slot("Tools")]` and additive `.Tools(...)` extension method (binary-safe for plugins).
- **Frontend** (`FieldWidget.tsx`):
  - `LabelPosition.Top`: tools render right-aligned on the label row, absolutely positioned so the label→input padding is identical with or without tools; label is bottom-aligned.
  - `LabelPosition.Left`: tools render below the label.

Verified live that a field with tools and one without have pixel-identical label padding (14px row, 10px gap to input).


artem-ivy-ai:
### Staging removed

PR #4701, opened 2026-07-03
[00031] Fix Copy Paste in Xterm Widget

nielsbosma:
# Summary

## Changes

Fixed copy/paste in the Xterm terminal widget. Ctrl+V paste previously never worked because xterm.js
captures keyboard input on its own internal hidden `<textarea>` inside the Shadow DOM, so the
container-level `keydown`/`paste` listeners never fired. Replaced them with xterm's
`attachCustomKeyEventHandler` API (which intercepts key events before xterm processes them) plus a
`paste` listener attached directly to xterm's own textarea for right-click paste.

## API Changes

None. This is a frontend-only fix inside `src/widgets/Ivy.Widgets.Xterm/frontend/src/Terminal.tsx` —
no C# widget API, props, or events changed.

## Files Modified

- `src/widgets/Ivy.Widgets.Xterm/frontend/src/Terminal.tsx` — replaced container-level clipboard event
  listeners with `term.attachCustomKeyEventHandler` (Ctrl+V/Cmd+V) and a textarea-level `paste` listener
  (right-click paste), keeping the existing `ClipboardAddon` for OSC 52 copy support.

## Manual Testing

1. Run a sample app that hosts the Xterm widget (e.g. `Ivy.Widgets.Xterm/.samples`) with an input-enabled
   terminal (`OnInput` event registered, not read-only).
2. Click into the terminal to focus it, copy some text to the system clipboard from anywhere, then press
   Ctrl+V (or Cmd+V on macOS). The copied text should be pasted into the terminal — previously nothing
   happened.
3. Right-click inside the terminal and choose "Paste" from the browser's context menu (if available) —
   text should also paste correctly.
4. Select text in the terminal and press Ctrl+C — copy should continue to work as before (unaffected by
   this change, handled natively by `ClipboardAddon`).
5. Confirm no console errors appear and the terminal still resizes/renders correctly, since the
   `attachCustomKeyEventHandler` runs on every key event across the whole terminal, not just paste.

---
Commits:
- 192137ffb Fix copy/paste in Xterm widget via attachCustomKeyEventHandler

---

### DIFF
--- a/src/widgets/Ivy.Widgets.Xterm/frontend/src/Terminal.tsx
+++ b/src/widgets/Ivy.Widgets.Xterm/frontend/src/Terminal.tsx
@@ -343,8 +343,10 @@ export const Terminal: React.FC<TerminalProps> = ({
       term.loadAddon(clipboardAddon);
     }
 
-    // Manual paste handler for Shadow DOM compatibility
-    const handlePaste = (e: ClipboardEvent) => {
+    // Paste handler for browser paste events (right-click paste, etc.) on
+    // xterm's internal textarea. Defined here (outer scope) so the cleanup
+    // closure below can remove the exact same listener reference.
+    const handleTextareaPaste = (e: ClipboardEvent) => {
       if (!allowClipboardRef.current || term.options.disableStdin) return;
       e.preventDefault();
       const text = e.clipboardData?.getData("text");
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #4731, opened 2026-07-25
Make AppRepository thread-safe to prevent duplicate apps in the sidebar

zachwolfe:
Tendril loads plugins on a background task at startup, so two Reloads ran concurrently; both re-read the shared Root and filled one tree, duplicating nearly every menu item. Reload now builds an immutable snapshot published with one Volatile.Write, and is serialized so a slow reload can't republish a stale tree over a later one's result.

The tree is now built from the resolved app dictionary, so each id gets one node. No public signatures changed.


artem-ivy-ai:
### Staging removed

PR #4746, opened 2026-08-12
feat(webmcp): expose view capabilities to browser AI agents via UseWebMcpTool

nielsbosma:
## What

Adds [WebMCP](https://webmachinelearning.github.io/webmcp/) support to Ivy. WebMCP is a W3C Web Machine Learning CG spec (Google + Microsoft, currently in a Chrome origin trial) that lets a page declare functions as **tools** a browser-resident AI agent can discover and invoke — instead of the agent screenshotting the DOM and guessing where to click.

Ivy is an unusually good fit: because apps are pure C# with server-held state, a tool handler runs against the *same* state the human is looking at, and the widget diff that follows re-renders the UI live as the agent works. Authors write one hook call and never touch JavaScript.

```csharp
[App(icon: Icons.Bot)]
public class TodoApp : ViewBase
{
    {
        var todos = UseState(ImmutableArray<string>.Empty);

        UseWebMcpTool("add-todo", "Add a new item to the user's active todo list",
            (AddTodoArgs args) => { todos.Set(todos.Value.Add(args.Text)); return $"Added \"{args.Text}\""; });

        UseWebMcpTool("list-todos", "List the user's current todo items",
            () => todos.Value,
            new WebMcpToolOptions { ReadOnly = true, UntrustedContent = true });

        return /* … */;
    }
}

public record AddTodoArgs(
    [property: Description("The text content of the todo item")] string Text);
```

Enable it in `Program.cs`. Without this the hook is inert and nothing reaches the browser:

```csharp
server.UseWebMcp();                                        // local dev
server.UseWebMcp(o => o.OriginTrialToken = "…");           // deployed origin
```

## API surface

| Member | Purpose |
|---|---|
| `UseWebMcpTool(...)` | Declares a tool for the lifetime of the view. 8 overloads: sync/async × with/without a typed args record × void/returning. |
| `UseWebMcpAvailability()` | `IState<WebMcpAvailability>` — `Unknown` → `Available`/`Unavailable`, so a view can offer a fallback instead of silently registering tools nobody can call. |
| `WebMcpToolOptions` | `Title`, `ReadOnly`, `UntrustedContent`, `Enabled`. |
| `WebMcpToolResult` | `Text` / `Error` / `Empty` for full control over the result. |
| `server.UseWebMcp(...)` | Opt-in, plus optional origin-trial token injection. |

Return values normalize exactly the way [`GoogleChromeLabs/use-webmcp-tool`](https://github.com/GoogleChromeLabs/use-webmcp-tool) does — `string` → text block, `null` → empty content, other objects → JSON, thrown or returned exceptions → `isError: true`. Input schemas come from `JsonSchemaExporter` (built into net10, no new package), with `[Description]` flowing through to per-argument docs.

## How it works

```
Build() ──UseWebMcpTool──▶ IWebMcpToolService (per connection)
                                  │ Send("WebMcpTools", […])
                                  ▼
                          lib/webmcp.ts ──registerTool──▶ document.modelContext
                                  ▲                              │ execute(input)
          Send("WebMcpToolResult", {callId, content, isError})   ▼
        AppHub.WebMcpToolCall(callId, toolId, argsJson) ◀──invoke── pending-call map
                    └─▶ AppSession.EventQueue.Enqueue(…)   (serialized with UI events)
```

Tool calls run on the session's existing `EventDispatchQueue`, the same queue UI events use, so agent actions serialize with human clicks and re-render through the normal widget diff.

## Design notes worth reviewing

Four constraints shaped this more than anything else:

- **The hub method doesn't await the handler.** `MaximumParallelInvocationsPerClient` is left at its default of 1 (`Server.cs`), so awaiting a slow tool would stall every other invocation on that connection — including `Event`. Raising the limit isn't an option either: parallel invocations would let two `Event` calls enqueue out of order. So the result comes back out-of-band keyed on a `callId`, mirroring the existing HTTP tunnel pattern inverted.

- **Registration is one mount effect plus a latest-ref.** `EffectQueue` only drains returned disposables at unmount, so a re-running effect would leak registrations rather than replace them; and `ViewContext.UseEffectHook` pins the first build's closure, which would otherwise capture stale state.

- **The browser registry is keyed by a per-`useBackend` scope, not the SignalR connection id.** That id changes on automatic reconnect, which would strand registrations with no owner left to release them. Apps aren't iframed — `AppHostWidget` mounts a second `useBackend` — so the AppShell and the hosted app share one `document.modelContext` and tool names are claimed document-wide.

- **Syncs are serialized and diffed.** `registerTool` is async and the server sends one message per tool as views mount. An abort-all-then-re-register-all approach raced against itself, collided on duplicate names, and silently dropped tools — reproducibly 2 of 4. Each scope now queues syncs behind one another and diffs by descriptor signature, leaving unchanged tools untouched.

## Verification

Driven end-to-end against the **real** Chrome WebMCP API (Chrome 151, `--enable-features=WebMCPTesting`), not a stub:

| call | result |
|---|---|
| `getTools()` | all 4 tools, correct schemas incl. `[Description]` text and annotations |
| `add-todo` | correct text **and** the item appears in the rendered UI with no reload |
| `list-todos` | `["native call"]` |
| `remove-todo` (bad arg) | `isError: true` |
| `clear-todos` | `Cleared 1 todo item(s)`, list emptied |
| availability | callout flips to the `Available` branch |

Also verified: closing the app's shell tab unregisters all tools; with no WebMCP present the page renders and behaves exactly as before (feature detection), and Ivy logs one actionable console warning naming the exact remedy rather than failing silently.

Solution builds clean (0 warnings), frontend typecheck/lint/format pass, `dotnet format --verify-no-changes` passes. Frontend test suite unchanged at 1284 passing — the 5 failures in `charts/sharedUtils.test.ts` are pre-existing locale-dependent assertions unrelated to this change.

## Not included

- Tests. Deliberately scoped out: `Ivy.Test` for schema generation and the service, vitest for the browser registry (`_resetForTesting` hooks are already in place), and a Playwright e2e injecting a fake `document.modelContext` via `addInitScript`.
- A docs page under `Ivy.Docs.Shared/Docs/03_Hooks/02_Core/`.
- AppShell-derived tools (`list-apps`, `navigate-to-app`) — the highest-leverage follow-up, since it would make every existing Ivy app agent-navigable with zero author effort.
- Cross-origin `exposedTo` / `fromOrigins`.

## Security posture

A tool executes server-side with the user's full auth session — the same authority as a button click. Only what `UseWebMcpTool` declares is ever exposed, `exposedTo` is never set (same-origin only), and `UntrustedContent` marks tools returning database rows so the agent treats them as data rather than instructions.

## Risk

The spec is young — `navigator.modelContext` → `document.modelContext` already happened once mid-origin-trial. All browser-API contact is confined to `src/frontend/src/lib/webmcp.ts`, so the next rename touches one file.

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Concepts/WebMcpToolsApp.cs
+++ b/src/Ivy.Samples.Shared/Apps/Concepts/WebMcpToolsApp.cs
@@ -1,0 +1,115 @@
+using System.Collections.Immutable;
+using System.ComponentModel;
+
+namespace Ivy.Samples.Shared.Apps.Concepts;
+
+/// <summary>
+/// Exposes this view's capabilities to a browser-resident AI agent through WebMCP.
+///
+/// The view renders the todo list but offers no way to edit it: there is no input, no add button and
+/// no delete action. Every mutation has to arrive through <c>document.modelContext</c>, which makes
+/// the point of WebMCP hard to miss — the handlers run server-side against the very state the UI is
+/// built from, so the list re-renders live as the agent works.
+/// </summary>
+[App(icon: Icons.Bot, searchHints: ["webmcp", "mcp", "agent", "ai", "tools", "modelcontext"])]
+public class WebMcpToolsApp : SampleBase
+{
+    private const string AddTool = "add-todo";
+    private const string AddDescription = "Add a new item to the user's active todo list";
+
+    private const string RemoveTool = "remove-todo";
+    private const string RemoveDescription = "Remove an item from the user's active todo list";
+
+    private const string ListTool = "list-todos";
+    private const string ListDescription = "List the user's current todo items";
+
+    private const string ClearTool = "clear-todos";
+    private const string ClearDescription = "Remove every item from the user's todo list";
+
+    public record AddTodoArgs(
+        [property: Description("The text content of the todo item")] string Text);
+
+    public record RemoveTodoArgs(
+        [property: Description("The exact text of the todo item to remove")] string Text);
+
+    protected override object? BuildSample()
+    {
+        var todos = UseState(ImmutableArray<string>.Empty);
+        var webMcp = UseWebMcpAvailability();
+
+        UseWebMcpTool(AddTool, AddDescription,
+            (AddTodoArgs args) =>
+            {
+                if (string.IsNullOrWhiteSpace(args.Text))
+                {
+                    return WebMcpToolResult.Error("A todo item needs some text.");
+                }
+
+                todos.Set(todos.Value.Add(args.Text));
+                return $"Added todo item: \"{args.Text}\"";
+            });
+
+        UseWebMcpTool(RemoveTool, RemoveDescription,
+            (RemoveTodoArgs args) =>
+            {
+                if (!todos.Value.Contains(args.Text))
+                {
+                    return WebMcpToolResult.Error($"There is no todo item called \"{args.Text}\".");
+                }
+
+                todos.Set(todos.Value.Remove(args.Text));
+                return $"Removed todo item: \"{args.Text}\"";
+            });
+
+        UseWebMcpTool(ListTool, ListDescription,
+            () => todos.Value,
+            new WebMcpToolOptions { ReadOnly = true, UntrustedContent = true });
+
+        UseWebMcpTool(ClearTool, ClearDescription,
+            () =>
+            {
+                var removed = todos.Value.Length;
+                todos.Set(ImmutableArray<string>.Empty);
+                return $"Cleared {removed} todo item(s).";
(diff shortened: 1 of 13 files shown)


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-framework-staging-docs-4746.sliplane.app](https://ivy-framework-staging-docs-4746.sliplane.app)
**🧪 Samples:** [https://ivy-framework-staging-samples-4746.sliplane.app](https://ivy-framework-staging-samples-4746.sliplane.app)

rorychatt (commented):
needs docs

PR #4750, opened 2026-08-13
[00076] Fix UTF-8 Corruption and Add Output Capture to Ivy.Hooks.Pty

nielsbosma:
# Summary

## Changes

Fixed a UTF-8 chunk-boundary decoding bug in `Ivy.Hooks.Pty`'s PTY reader loop (multi-byte
characters split across a 4KB read previously decoded to `�` replacement characters) by hoisting a
single stateful `Encoding.UTF8.GetDecoder()` into a new internal `PtyOutputDecoder`. Added an
opt-in, size-capped output transcript (`PtyOptions.CaptureOutput` / `MaxCaptureLength` /
`PtyHandle.Output`) sharing the same decode stream as the existing `OnOutput` callback, plus a new
public `AnsiEscape.Strip` helper (a C# port of the frontend's ANSI-stripping regexes) for reading
captured output as plain text. Documented the previously-undocumented three-way
`OnInput`/`Stream`/`OnOutput` data-path split and rewrote the stale Xterm widget README.

## API Changes

- **`Ivy.Hooks.Pty.PtyOptions`** — new properties: `bool CaptureOutput` (default `false`), `int MaxCaptureLength` (default `1_000_000`).
- **`Ivy.Hooks.Pty.PtyHandle`** — new property: `string Output` (empty string unless `CaptureOutput` is set; capped, keeps newest output). No change to the record's positional constructor parameters, so all existing `new PtyHandle(...)` call sites are unaffected.
- **`Ivy.Hooks.Pty.AnsiEscape`** (new public static class) — `static string Strip(string? text)`.
- **`Ivy.Hooks.Pty.PtyOutputDecoder`** (new, `internal`) — not part of the public API surface.

## Files Modified

**Core fix**
- `src/Ivy.Hooks.Pty/PtyOutputDecoder.cs` (new) — stateful UTF-8 decoder + capped transcript.
- `src/Ivy.Hooks.Pty/AnsiEscape.cs` (new) — ANSI/control-char stripper.
- `src/Ivy.Hooks.Pty/UsePty.cs` — wires the decoder into the reader loop and spawn-failure path; adds `PtyOptions`/`PtyHandle` members.

**Tests**
- `src/Ivy.Hooks.Pty.Test/` (new project) — `PtyOutputDecoderTests.cs`, `AnsiEscapeTests.cs`, registered in `Ivy-Framework.slnx` alongside `Ivy.Hooks.Pty` (neither was previously part of the solution's CI build/test path).
- `src/Ivy.Hooks.Pty/Ivy.Hooks.Pty.csproj` — `InternalsVisibleTo` for the new test project.

**Documentation**
- `src/Ivy.Hooks.Pty/README.md` (new).
- `src/widgets/Ivy.Widgets.Xterm/README.md` — rewritten against the actual `Terminal.cs` props/events.

**Sample**
- `src/widgets/Ivy.Widgets.Xterm/.samples/Apps/OutputCaptureApp.cs` (new) — live `Terminal` next to a `PtyHandle.Output` transcript pane, refreshed via `UseInterval`; also detects `https?://` URLs in the captured, ANSI-stripped transcript and lists them as clickable link buttons.

## Manual Testing

1. Run the Xterm samples app: `cd src/widgets/Ivy.Widgets.Xterm/.samples && dotnet run` (add `--find-available-port` if the default port is busy).
2. Open the "Output Capture" app (Terminal group) in the browser.
3. In the left terminal pane, type `echo héllo ✅ 🎉 ┌─┐` and press Enter.
4. Observe the right-hand read-only pane: it should show the same accented/emoji/box-drawing characters with no `�` replacement characters, and no visible ANSI escape codes.
5. Keep producing output (e.g. repeat the echo, or run `dir`/`ls`) until more than 4000 characters have been printed; confirm the right pane's earliest lines are dropped while the newest output remains, and its length never grows unbounded.
6. Type a command that prints a URL (e.g. `echo Server running on https://localhost:5011`) and confirm it appears as a clickable link in the "Detected URLs" section below the panes.

This was also independently verified by the `IvyFrameworkVerification` delegated check, which built
a standalone Playwright-driven sample exercising the same scenarios plus `PtyOptions.OnOutput` and
`AnsiEscape.Strip` in isolation.

---
## Commits

- `f49eea237` sample(hooks-pty): detect URLs in captured PTY output
- `e33578c52` sample(hooks-pty): add Output Capture demo app
- `a51f3246d` docs(hooks-pty): add Ivy.Hooks.Pty README, rewrite stale Xterm README
- `faf7d463f` test(hooks-pty): add Ivy.Hooks.Pty.Test project
- `630fc483f` fix(hooks-pty): fix UTF-8 chunk-boundary corruption, add output capture

---

### DIFF
--- a/src/Ivy-Framework.slnx
+++ b/src/Ivy-Framework.slnx
@@ -39,6 +39,8 @@
   <Project Path="Ivy.Docs.Shared/Ivy.Docs.Shared.csproj" />
   <Project Path="Ivy.Docs.Test/Ivy.Docs.Test.csproj" />
   <Project Path="Ivy.Docs/Ivy.Docs.csproj" />
+  <Project Path="Ivy.Hooks.Pty/Ivy.Hooks.Pty.csproj" />
+  <Project Path="Ivy.Hooks.Pty.Test/Ivy.Hooks.Pty.Test.csproj" />
   <Project Path="Ivy.Samples.Shared/Ivy.Samples.Shared.csproj" />
   <Project Path="Ivy.Samples/Ivy.Samples.csproj" />
   <Project Path="Ivy.Test/Ivy.Test.csproj" />
--- a/src/Ivy.Hooks.Pty.Test/AnsiEscapeTests.cs
+++ b/src/Ivy.Hooks.Pty.Test/AnsiEscapeTests.cs
@@ -1,0 +1,93 @@
+namespace Ivy.Hooks.Pty.Test;
+
+public class AnsiEscapeTests
+{
+    [Fact]
+    public void Strip_CsiColorCodes()
+    {
+        var result = AnsiEscape.Strip("\x1b[31mred\x1b[0m");
+
+        Assert.Equal("red", result);
+    }
+
+    [Fact]
+    public void Strip_OscSequences_BelTerminated()
+    {
+        var result = AnsiEscape.Strip("\x1b]0;title\x07text");
+
+        Assert.Equal("text", result);
+    }
+
+    [Fact]
+    public void Strip_OscSequences_StTerminated()
+    {
+        var result = AnsiEscape.Strip("\x1b]0;title\x1b\\text");
+
+        Assert.Equal("text", result);
+    }
+
+    [Fact]
+    public void Strip_OscSequences_UnterminatedAtEndOfString()
+    {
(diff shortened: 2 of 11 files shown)


artem-ivy-ai:
### Staging removed

PR #4762, opened 2026-08-21
Support arbitrary URL query parameters as appArgs in frontend getAppArgs

rorychatt:
### Summary
When users visit URLs containing individual query parameters like `/drafts?planId=00001-...&share=1`, `getAppArgs()` in the frontend previously returned `null` because it only looked for `?appArgs=...`. As a result, the SignalR connection was initiated without the query parameters, preventing `AppContext.GetArgs<T>()` and `UseArgs<T>()` from reading `planId`.

### Changes
- Updated `getAppArgs()` in `utils.ts` to collect all non-reserved query parameters into a JSON string if `appArgs` is not explicitly provided.
- Added comprehensive unit tests in `utils.test.ts` for explicit and implicit query parameter extraction.
- Rebuilt frontend assets.

### DIFF
--- a/src/frontend/src/lib/utils.test.ts
+++ b/src/frontend/src/lib/utils.test.ts
@@ -811,3 +811,76 @@ describe("buildSignalRUrl", () => {
     expect(parsed.searchParams.has("oauthLogin")).toBe(false);
   });
 });
+
+describe("getAppArgs", () => {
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #4764, opened 2026-08-25
fix(rich-text): keep strikethrough visible on links

nielsbosma:
## Problem

`RichTextBuilder.Link` accepts `strikeThrough`, and `TextRun.StrikeThrough` reached the frontend, but the decoration was silently dropped when rendering a link run — a struck-through link rendered underlined only.

Both link branches in `RichTextBlockWidget` composed their classes as `cn(className, "underline")`, where `className` already carried `"line-through"`. `cn` is `twMerge(clsx(...))`, and tailwind-merge puts `underline` and `line-through` in the same `text-decoration-line` group, so the later class won:

```
twMerge('font-semibold italic line-through', 'underline')
  -> 'font-semibold italic underline'
```

## Fix

Decoration is kept out of the shared `className`. Non-link runs still use `line-through`; link runs get a single combined utility, `[text-decoration-line:underline_line-through]`, so tailwind-merge has nothing to collapse. Applies to both the `<a>` and the `OnLinkClick` `<button>` path.

## Verification

- New `RichTextBlockWidget.test.tsx` — 5 tests: plain link stays underline-only, struck link keeps both decorations on both link paths, bold/italic survive alongside, non-link runs still use `line-through`.
- Ran the suite against the **pre-fix** source: the 3 strikethrough-link tests fail (`["underline"]` received), the 2 controls pass. So the tests genuinely pin the behavior rather than just describing it.
- Verified in the running samples app — computed `text-decoration-line`:

| Link | Computed |
|---|---|
| struck link | `underline line-through` |
| plain link | `underline` |
| bold + struck link | `underline line-through` |

- `tsc -b` clean; full frontend suite 1304 passed. The 5 failures in `src/widgets/charts/sharedUtils.test.ts` are pre-existing on `development` (confirmed by stashing this change and re-running).

## Also

Adds a **Strikethrough Links** section to the RichText sample app covering struck link, plain link, struck non-link text, and bold+struck link.

## Not addressed

`Button`/`LinkBuilder` links lose their strikethrough **on hover** — the `link` variant's `hover:underline` overrides the base `line-through`. Base state renders correctly. Left alone as out of scope.

### DIFF
--- a/src/Ivy.Samples.Shared/Apps/Widgets/Primitives/RichTextApp.cs
+++ b/src/Ivy.Samples.Shared/Apps/Widgets/Primitives/RichTextApp.cs
@@ -37,6 +37,18 @@ public class RichTextApp : SampleBase
             .Link("Ivy Framework", "https://github.com/example/ivy")
             .Run("for more info.");
 
+        var struckLinks = Text.Rich()
+            .Run("The")
+            .Link("old guide", "https://github.com/example/ivy", strikeThrough: true)
+            .Run("has moved to the")
+            .Link("new guide", "https://github.com/example/ivy")
+            .Run(". A struck-through link keeps both its underline and its line-through.")
+            .LineBreak()
+            .StrikeThrough("Plain struck text")
+            .Word("sits next to a")
+            .Link("bold struck link", "https://github.com/example/ivy", bold: true, strikeThrough: true)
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

